PR #5, opened 2018-04-06
Chores/table columns #156586034

slowbot:
Add styles for min table cell widths and wrapping table headers.

CSS only so... yeah. 

https://github.com/Exygy/sf-dahlia-pattern-library/pull/72


chrisdolendo (approved):
Looks good!

PR #13, opened 2018-04-26
features/page-header-component-#15478417

fddayan:
[#154784173]


chrisdolendo (approved):
This looks good. I'm curious about the benefit once we do an integration with the page header story, so let's tune into that. I'm looking forward to seeing this in action and insights on that.

chrisdolendo:
I see the presentational aspect of it since you added logic! Now I'm wondering if this would be challenging to deliver as a story... how would we be able to show the different logic that's in here without repeating work? Let me know if you have an idea -- if you think this is repeating the work and we might as well feed two birds with one seed. But the code looks good, I'm thinking about process now.

app/javascript/components/organisms/PageHeaderSimple.js
```diff
@@ -0,0 +1,17 @@
...
+  return (
+    <header className="lead-header short bg-snow">
+    	<div className="row full-width inner--3x">
+    		<div className="large-12 columns ">
+    			<hgroup className="lead-header_group">
+    				<h1 className="lead-header_title small-serif c-oil">{title}</h1>
+    			</hgroup>
+    		</div>
+    	</div>
+    </header>
```
slowbot, line 13: @fddayan We are using spaces everywhere else and this section looks like its using tabs.

fddayan: got it .. I will adjust my editor

PR #19, opened 2018-05-08
Chores/lap pl fixes #157375882

slowbot:
Fixes: 
Dropdown menu item import component naming

Refactoring:
ContentSection: as molecule that is imported into TabCard
ContentSectionBordered: is a new container with border style and spacing rules for a collection of view only FormGroupValues used on Short Form tab. (Could be done as a variant of ContentSection)

New:
TableSimple: Used for non-editable tabular data within Short Form and Supplemental
TableExpandable: Used for editable tabular data within Short Form and Supplemental
EditablePanel: Container for nested FormGridRows used in TableExpandable
FormGridTable: Container with .scrollable-table-container

https://www.pivotaltracker.com/story/show/157375882


fddayan (approved):
Code looks good. There are a few things that can be improved with ES6 and React Magic, but we'll take care of that

PR #21, opened 2018-05-11
Features/page header integration #154064861

callaghanc:
This approach to integrating the page header uses a Rails layout partial and defines page header content in controller methods. We could, alternatively, add a render component line for the page header to each individual view and define the page header content in the view. Or perhaps there is a more React-y way to do this! 

This PR still needs to get tests passing and possibly have some updates that adjust the appearance of the page header.


app/views/layouts/_page_header.html.slim
```diff
@@ -0,0 +1,9 @@
+- if @page_header
```
fddayan: My 2 cents about this approach: I would try to use `PageHeader` inside a React component instead rendering the page header in the layout using `react_component`. For example, An `<ApplicationPage>` component like 

```
 <div class='applications'>
 	<PageHeader ...>
 	<IndexTable ...>
 </div>
```

Then

```
react_component 'ApplicationPage', { results: @applications, fields: @fields, links: ['View Application'] }
```

we could create a <PageLayout> component that wrapps `PageHeader` if we want to remove duplication
```
<PageLayout ...>
 	<IndexTable ...>
</PageLayout>
```
Also, another component could wrap `PageHeaderSimple` or `PageHeader` and decide what to render.

I would do this because: 
* it looks like we are doing React all the way down, so we do not get married to `Rails` to handle our layouts.
* easier integration for `mobx` or any state managment 
* easier migration to SPA or migration to `Routes` using `react-routes`
* ...in a way, better jest test coverage

**However** I 100% feel this approach and I understand how it's removing duplication. It's not the end of world (or maybe it is :| ......... just kidding). I just think there is value in staying in React and not mixing Rails features and just try to use Rails as a backend API as much as possible.

As general rule I would to call `react_component` only once per request.

@chrisdolendo @creativecoder  any thoughts?

creativecoder: +1. I'd prefer to keep React self contained for all the reasons stated.

callaghanc: Sounds good! That's why I was trying so hard to get ahold of a React person to pair on this 😉 @fddayan, let's find time today or tmrw to pair on the React-y solution to this 🙂

fddayan: 👍

callaghanc:
@fddayan @chrisdolendo I've applied the general approach to page structure that we collaborated on across the site. Take a look and see what you think. This PR still needs some CSS/styling work on the headers that is unrelated to the structural work we have been doing. That's what I'll be working on next while you guys take a look at the PR so far. Once we all feel happy with the structural approach, I will go on to update tests.

callaghanc:
Note that the changes made by this PR cause the page header on the listing page to appear below the nav tabs for Listing and Application on that page. Because the tabs are currently rendered in Rails, they appear outside of the React content of the page, which now includes the page header. Previously, the tabs were appearing below the page header. We should not worry about trying to reposition the Rails-rendered tabs in this story, since those tabs will be able to be replaced by a proper tab navigation React component, once both this PR and the tabs nav PR (https://github.com/Exygy/sf-dahlia-lap/pull/20/files) are merged to master. If we don't already have a story for this, I will create one.

app/javascript/components/lease_ups/LeaseUpsPage.js
```diff
@@ -1,15 +1,48 @@
...
+}
+
 const LeaseUpsPage = ({ listing, results }) => {
-  return (
-    <div>
-      <LeaseUpsHeader listing={listing} />
-      <LeaseUpsTableContainer listing={listing} results={results} />
-    </div>
-  )
+  if (listing) {
```
fddayan, line 31: It looks like we are using the same controller for handling LeaseUp for specific listing and all lease ups. Should we have `LeaseUpPage` to handle lease up without listing and and create `ListingLeaseUpPage` to handle lease ups for listings ?

callaghanc: @fddayan I think the lease up-related structure that currently exists in this PR will definitely change based on the work you're doing right now in the controller refactor PR : https://github.com/Exygy/sf-dahlia-lap/pull/24/. Based on our discussion of that PR this morning, it sounds like we are very close to getting that PR finished, accepted, and merged. So I would be happy to wait for that to happen, then merge master with that controller refactor work into this PR, and then do the lease up page structure based off that.

callaghanc: After further discussion with @fddayan, it was decided to leave the lease up-related page structure as-is in this PR and let future lease up-specific stories deal with the specifics of lease up pages.

fddayan (approved):

PR #23, opened 2018-05-15
Chores/add overcommit #157123836

callaghanc:
Recommend trying out some test commits in which you modify Ruby files, JS files, try to commit directly to master, etc. Test whether Overcommit is doing what we want it to do!


chrisdolendo (approved):
looks good!

PR #25, opened 2018-05-21
Features/single lease up status history #154072850

fddayan:
## Review Instructions
1. Go to https://dahlia-lap-full-pr-25.herokuapp.com/applications/a0o0x000000OHykAAG/supplementals
1. Acceptance Criteria 1) 
  - you should see the status history at the bottom. 
  - The status history list should match the entries in this view https://sfhousing--full.cs95.my.salesforce.com/a12? fcf=00B0x000000O65j&rolodexIndex=-1&page=1. Those entries belong to the same application (APP-00186221)
1. AC 2) list is sorted by date. Entries with no date are at the bottom.
1. AC 3) Add comment button should be display. Clicking in the button will not trigger a modal.
1. AC 4) go here http://localhost:3000/applications/a0o0P00000FFc2yQAD/supplementals. Status history should not be displayed
1. AC 5) You can go here to check with use the right SOQL https://github.com/Exygy/sf-dahlia-lap/pull/25/files#diff-f0c19da4297984adec2f0fc1ff62a42bR8


fddayan:
@slowbot can you check again ? I made those changed and pushed

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -0,0 +1,72 @@
...
+}
+
+const mapProperties = ({application, statusHistory}) => {
+  return {
+    application: mapApplicationProperty(application),
+    statusHistory: mapStatusHistoryProperty(statusHistory)
+  }
+}
+
+export default mapProps(mapProperties)(SupplementalApplicationPage)
```
fddayan: @chrisdolendo @callaghanc @creativecoder 
Here I'm using the property mapping "technique". 
Basically, I transform the properties based giving in `react_component` to another properties shape. 
Whatever `mapProperties` returns is what `SupplementalApplicationPage` applications is going to get. 
I'm doing this to avoid propagating SalesForce domain object thru our component tree. 
For example, to avoid having `application.Listing.Id` in a component. Instead use.
propMaping  should be done in all `*Page.js` component that get a Saleforce object tree.

I think this is useful because:
* Our domain model does not exactly match salesforce domain model. 
* Easier to reason about components.
* A state view rules how the component works. We map the domain to the view state, so there is less coupling. This is easier for coding component isolated either by defining the state first and then the DOM or the other way around. (Basically to do not need to know what saleforce returns to do the component)
* Easier for testing too. 
* We do not propagate a domain model we do not have control of. I know it's a bit paranoid 😜, but we do not have full control of those structures and a change there might imply a change in many places. If something changes in the structure, we know where to go.

Notes:
* We could have our domain objects and loading them in the `mapProp`, but too soon to tell.
* In a way this could be accomplish in the backend by returning our domain objects. 

**This is my personal design choice, but I think we all should agree with this.**

creativecoder: My opinion is that ideally, this domain mapping would be a server-side concern and not pollute the front-end code. We have some ideas around that planned in this chore: https://www.pivotaltracker.com/story/show/154747783

However, it may be some time before we're able to get to that chore. For now, I think this front-end field mapping is a great idea, and lays a nice groundwork for the server side changes. When we do get to server-side changes, we can use the mapping logic here to inform how we restructure the Salesforce data, and then updates the front-end code will mostly be removing these mapping functions--making for a pretty clean transition.

So let's keep adding front-end mapping until we get to the server-side chore!

callaghanc: @fddayan @creativecoder Yep as Grant says above this general idea is something we've talked about for a while, so overall the team is in agreement on mapping being a good + necessary thing. Seeing this, though, I'm wondering why we would spend time doing this in the frontend now and then re-doing this in the backend later. Is the work needed to do this server-side significantly greater than the work we see here? If the answer is yes, then what is here is good and let's roll with this. If the answer is no, then we should discuss bringing the server-side mapping story up in priority.

callaghanc: We all discussed this in stand-up and agreed to approve this approach 👍

app/javascript/components/molecules/ContentSection.js
```diff
@@ -1,22 +1,39 @@
 import React from 'react'
```
chrisdolendo, line 1: seeing two ConentSection components

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
chrisdolendo: this will get bigger and bigger, so as it gets bigger, we need to either pull it out to a central file now or later.

app/services/force/application_service.rb
```diff
@@ -22,6 +22,7 @@ def application(id)
         AND Status__c != '#{DRAFT}'
         AND #{user_can_access}
       ))
+      # prybug
```
chrisdolendo: remove :)

app/services/force/base.rb
```diff
@@ -75,6 +75,9 @@ def api_post(endpoint, params)
       api_call('post', endpoint, params)
     end
 
+    # def api_get
```
chrisdolendo: let's remove commented code to keep it clean

app/views/pattern_library/index.slim
```diff
@@ -32,7 +34,7 @@ h2 Molecules
 
 h2 Organisms
 
-== pl_component 'ContentSection', { title: 'Title', subtitle: 'Some description', label: 'Text input', id: 'text-area', name: 'Area', placeholder: 'Enter text', describeId: 'lable', note: 'More text', error: 'Please enter a First Name' }
+/== pl_component 'ContentSection', { title: 'Title', subtitle: 'Some description', label: 'Text input', id: 'text-area', name: 'Area', placeholder: 'Enter text', describeId: 'lable', note: 'More text', error: 'Please enter a First Name' }
```
chrisdolendo: any reason why we're keeping the commented out pl_components?

chrisdolendo (approved):
Looks great @fddayan! Just had some small stuff here and there to remove commented code, and a Q around pulling out the property mapper to another file either now or at a later story when it gets too big.

app/views/pattern_library/index.slim
fddayan: no reason, removing

app/javascript/components/molecules/ContentSection.js
fddayan: where is the other component?

fddayan:
@chrisdolendo fixed all comments except `seeing two ConentSection components`. Where is it?

PR #26, opened 2018-05-24
Features/page layout #157673726

fddayan:
https://www.pivotaltracker.com/story/show/157673726


app/services/force/base.rb
```diff
@@ -41,9 +41,29 @@ def debug(q)
     end
     # run a Salesforce SOQL query
 
+    # This method was added to help developers debug/code faster.
+    # because some queries take way too long (like Applications)
+    # It's not designed to be used in prod.
+    # it can be enable by setting CACHE_SALESFORCE_REQUESTS=yes
+    # I'm not using cache_query since the logic it's different. cache_query should be removed. Also, it's not being used.
+    # Fed
+    def query_with_cache(q)
```
fddayan, line 50: @chrisdolendo @callaghanc @creativecoder I added this crazy method to cache responses to improve coding productivity. Some queries take too long and moving around the app it's kind of slow. We talked about caching, but this is `caching for developers`. Let me know if we can keep it or if we should remove it. There is an extra check in `query` to make this work. https://github.com/Exygy/sf-dahlia-lap/pull/26/files#diff-b6d0b13d1138c63c05d714686a5426b0R62

app/javascript/components/applications/ApplicationDetails.js
```diff
@@ -1,24 +1,17 @@
 import React from 'react'
+import _ from 'lodash'
```
creativecoder: Not sure how we are handling this elsewhere, but if you import only what you need (i.e. `import { map } from 'lodash'`) you can save quite a bit on final bundle size.

app/javascript/components/applications/ApplicationPage.js
```diff
@@ -1,24 +1,29 @@
...
-      <PageHeader
-        title={`Application ${application.Name}`}
-        content={<span>Name of Listing: <a href={`/listings/${application.Listing.Id}`}>{application.Listing.Name}</a></span>} />
-    </div>
-  )
+const buildActionLinkIfNecessary = (application) => {
+  if (!application.Is_Lottery_Complete &&
+      application.Application_Submission_Type === 'Paper') {
+    return { title: 'Edit Application', link: `/applications/${application.Id}/edit` }
+  }
```
creativecoder: Might be more readable to explicitly return `undefined` when the Edit link isn't needed.

app/javascript/components/applications/ApplicationsPage.js
```diff
@@ -20,12 +12,17 @@ const ApplicationsPageTable = ({ applications, fields }) => {
...
+  pageHeader: {
+    title: 'Applications'
+  }
+}
+
 const ApplicationsPage = (props) => {
   return (
-    <div>
-      <ApplicationsPageHeader/>
+    <TableLayout {...layout}>
```
creativecoder: Any specific reason you're de-structuring `layout`, rather than just passing `pageHeader` as a prop? Seems like it would be nice to try and use a consistent style.

app/javascript/components/layouts/CardLayout.js
```diff
@@ -0,0 +1,39 @@
...
+import PageHeaderLayout  from './PageHeaderLayout'
+import TabsSection from '../organisms/TabsSection'
+import TabCard from '../organisms/TabCard'
+import AppCard from '../molecules/AppCard'
+
+class CardLayout extends React.Component {
+
+  componentWillMount = () => {
+    let orig = document.body.className
+    document.body.className = orig + (orig ? ' ' : '') + 'bg-snow'
```
creativecoder: May be able to use `classList` here, checking with @slowbot on browser support.

app/javascript/components/organisms/TabCard.js
```diff
@@ -1,14 +1,33 @@
...
-import ContentSection from '../molecules/ContentSection'
 
-const TabCard = ({ children }) => {
+import classNames from 'classnames'
+
+// import ContentSection from '../molecules/ContentSection'
+
+// <div className="app-card form-card tabs-card max-width expand-on-small">
+//   {children}
+// </div>
```
creativecoder: Remove comment

app/views/layouts/application.html.slim
```diff
@@ -17,11 +17,12 @@ html
     / include toolkit.css and other CSS overrides
     == stylesheet_link_tag "application", :media => 'all', 'data-turbolinks-track' => 'reload'
     == csrf_meta_tags
-  body.bg-vapor
+  body
+    //.bg-vapor
```
creativecoder: Remove comment

creativecoder (commented):
Generally the Card and Table Layouts work how I'd expect.

One possible optimization: I'm not seeing how using both `PageHeader` and `PageHeaderSimple` in those layouts is helping us. It seems like everything can use PageHeader, you just pass only the props you want to see in the header.

If that works, we can eliminate `PageHeaderLayout` as well.

app/javascript/components/applications/ApplicationsPage.js
fddayan: 👍 no reason, I guess I was experimenting. I will make it consisting with pageHeader

app/services/force/base.rb
callaghanc: @fddayan I'm fine with this if it speeds up development. Maybe add an env == dev clause to line 62 to make it even more clear this is just a dev convenience.

fddayan: 👍

app/javascript/components/layouts/CardLayout.js
slowbot: @creativecoder @fddayan As far as our users go this looks safe to me.

fddayan: fixed

chrisdolendo (approved):
Great refactoring

PR #27, opened 2018-05-24
Adding simple spinner and loading wrapper

slowbot:
1. Go to https://dahlia-lap-full-pr-27.herokuapp.com/pattern_library 
1. You should see two new components. 
   - Spinner: https://dahlia-lap-full-pr-27.herokuapp.com/pattern_library#spinner is the css loading animation
   - Loading: https://dahlia-lap-full-pr-27.herokuapp.com/pattern_library#loading is a responsive container centering the animation vertically and horizontally


app/javascript/components/molecules/Loading.jsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
```
fddayan: I would rename this file from Loading.**jsx** to Loading.**js** since that is what we are using for all the jsx files.

app/javascript/components/atoms/Spinner.js
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+
+const Spinner = ({  }) => {
```
fddayan: I would replace `({  })`  for just `()`  You probably saw I was doing that on one of my components, but since we probably are going to implement eslint soon, it's going to complain about it.

fddayan (changes requested):
Looks good. Just 2 minor cleanup changes and we can approve.

fddayan (approved):

fddayan:
@slowbot approved

PR #29, opened 2018-05-30
Features/apps table status change #154064901

callaghanc:


app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -1,12 +1,90 @@
...
+        comment: comment,
+        applicationId: applicationId
+      }
+
+      var response = await apiService.createLeaseUpStatus(data)
+
+      if (response) {
+        // find the application in question in the applications table data
+        // and update its lease up status value and status updated date
+        // in the table
```
callaghanc, line 84: @fddayan After we chatted about whether to update data in the table or to pull from Salesforce after the user creates a new field update comment, I started working on the "pull just the updated application" approach we talked about, but it didn't feel quite right. I felt like it was bypassing the normal flow of data btwn Rails and the lease ups table and adding complexity that could be distracting. So I decided in the end to go with the approach of updating the new status and date in the table itself.

callaghanc:
@chrisdolendo I've resolved merge conflicts on this and tests are passing locally, but tests are not passing on Semaphore. I'm over hours for the week, so can you take a look at Semaphore for this tmrw?

callaghanc:
@slowbot I messed around a bit with the modal dropdown but haven't yet come across a satisfactory solution to its display issues. I'm done for the week, so can you take a look and see if you can get it to be the right width and position? For its overflow, I actually think it may be better to keep it contained inside the modal, not letting it overflow, and instead have the dropdown menu scroll.

chrisdolendo (approved):
:+1: Great! 🎉

app/javascript/components/applications/ApplicationDetailsContentCard.js
```diff
@@ -38,16 +38,16 @@ var generateContent = (dataCollection, field, labelMapper, i) => {
...
+// var generateData = (label, value) => {
+//   return (
+//     <div className="margin-bottom--half" key={i}>
+//       <h4 className="t-sans t-small t-bold no-margin">
+//         {label}
+//       </h4>
+//       <p>{value}</p>
+//     </div>
+//   )
+// }
```
fddayan: should we removed this commented code?

app/javascript/components/applications/application_form/AlternateContactSection.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import _ from 'lodash'
```
fddayan: bases on grant suggestions we are importing from lodash whatever we are actually using to help performance in our bundle

app/javascript/components/applications/application_form/HouseholdMembersSection.js
```diff
@@ -1,4 +1,5 @@
 import React from 'react'
+import _ from 'lodash'
```
fddayan: same thing here about lodash..not sure if you worked a lot in this components and refactoring this is out of the scope of this PR

app/javascript/components/lease_ups/LeaseUpsStatusModalWrapper.js
```diff
@@ -0,0 +1,61 @@
...
+    };
+  };
+
+  render() {
+    return (
+      <FormModal
+        header="Update Status"
+        primary="update"
+        secondary="cancel"
+        isOpen={this.props.isOpen}
```
fddayan: about `props` I would follow this conventions when using many props: Instead of referencing `this.props.whatever` I would extract the props in the first line in the render method like `const { whatever } = this.props` so you can see what properties are being using in the render method and you do not have to scan the code to see what it's being used. Also, we save a few characters in the JSX, but this is minor improvement :P

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -1,10 +1,90 @@
...
+        showAlert: null,
+        loading: false,
+      },
+      results: this.props.results
+    }
+  }
+
+  updateState = (path, value) => {
+    this.setState(prevState => {
+      var newState = cloneDeep(prevState)
```
fddayan: what are we deep cloning the state? whatever we return in it gets merged to the current state. so if you state is  `{a:1, b2}` and you return `{b:3}`. The state becomes `{a1, b:3}`. you do not lose the other keys.
also passing a function to setState is technically used only if you need a value from the `prevState` because react updates the state async. Ping me on slack if this is confusing.

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
fddayan: 👍

app/javascript/components/lease_ups/leaseUpsHelpers.js
```diff
@@ -0,0 +1,24 @@
+import _ from 'lodash'
+
+export const LEASE_UP_STATUS_OPTIONS = [
+ {value: 'Processing', label: 'Processing', style: 'is-processing'},
+ {value: 'Withdrawn', label: 'Withdrawn', style: 'is-withdrawn'},
+ {value: 'Appealed', label: 'Appealed', style: 'is-appealed'},
+ {value: 'Waitlisted', label: 'Waitlisted', style:'is-waitlisted'},
+ {value: 'Disqualified', label: 'Disqualified', style:'is-disqualified'},
+ {value: 'Approved', label: 'Approved', style:'is-approved'},
+ {value: 'Lease Signed', label: 'Lease Signed', style:'is-leased'},
```
fddayan, line 10: 👏

app/javascript/components/molecules/FormModalLoading.js
```diff
@@ -0,0 +1,12 @@
+import React from 'react'
+import Loading from '../molecules/Loading'
+
+const FormModalLoading = () => {
```
fddayan, line 4: Just adding a note that we might need a loading HOC to handle all the loadings for components. Nothing we have to implement now. Just making a note.

fddayan (changes requested):
Overall looks good. I added some cleanup changes and a question about why doing deep clone if states

app/javascript/components/applications/ApplicationDetailsContentCard.js
callaghanc: not even sure why it was commented out in the first place! we don't use this function anywhere so yes, will remove this commented code

app/javascript/components/applications/application_form/HouseholdMembersSection.js
callaghanc: yeah what this import is and the one above it is me adding imports where lodash was used but never imported, did so to pass overcommit, which was checking the whole codebase, not just the files i was editing. so even though this story didn't involve these components i added those imports. that was before grant pointed out that we should just import the specific lodash functions we need. i can definitely go back and update these to do that, since i'm bothering to add the imports at all i'll update them to use our new convention

app/javascript/components/lease_ups/LeaseUpsStatusModalWrapper.js
callaghanc: nice, i'll update to use that convention, thanks!

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
callaghanc: my aim here was to not override nested objects - i'll ping you on slack to discuss further - definitely would be great if we don't have to deep clone!

callaghanc: @fddayan i pushed the new state update functions we worked on

fddayan (approved):

PR #31, opened 2018-06-05
Features/confirmed demographics #155196549

fddayan:
https://www.pivotaltracker.com/story/show/155196549


app/javascript/components/molecules/FormGridRow.js
```diff
@@ -11,4 +11,24 @@ const FormGridRow = ({ label, id, name, placeholder, describeId, note, error })
   )
 }
 
+const FormGrid = {}
```
callaghanc: @fddayan Why is this block of code here as well as in the file above?

app/javascript/components/supplemental_application/mapProperties.js
```diff
@@ -2,6 +2,7 @@ import _ from 'lodash'
 import moment from 'moment'
 
 import { mapApplication } from '../propMappers'
+import { updateApplicationAction } from './actions'
 
 // NOTE: what kind of saleforce domain object is this one? Fed
```
callaghanc, line 7: @fddayan This is a `Field_Update_Comment__c` Salesforce object - if you check out that object in the Workbench SOQL Explorer you can see the fields on it.

app/javascript/components/molecules/FormGridRow.js
fddayan: forgot to remove it :P

app/javascript/components/supplemental_application/mapProperties.js
fddayan: Great thanks.

app/javascript/components/Alerts.js
```diff
@@ -0,0 +1,10 @@
+const ERROR_MESSAGE = "Oops! Looks like something went wrong. Please try again."
+
+const info = (message) => alert(message)
+
+const error = () => alert(ERROR_MESSAGE)
```
callaghanc, line 5: @fddayan I added a standard error alert using the same default error message we use on the web app for consistency.

callaghanc:
@fddayan looks good! let's chat quickly tmrw about all the mapping - i want to make sure i'm up to date on the whole story there - and then should be good to go 🙂

app/javascript/components/Alerts.js
fddayan: 👍

fddayan:
@callaghanc fixed

callaghanc (approved):
👍

PR #32, opened 2018-06-11
Features/add alice griffith preference #157820576

fddayan:
https://www.pivotaltracker.com/story/show/157820576

### DIFF
--- a/app/javascript/components/application_form/preferences/AliceGriffithFields.js
+++ b/app/javascript/components/application_form/preferences/AliceGriffithFields.js
@@ -14,14 +14,19 @@ const buildTypeOfProofOptions = (values) => {
   ))
 }
 
+const FormGroup = ({children}) => (
+  <div className='form-group'>{children}</div>
+)
+
 const AliceGriffithFields = ({i, householdMembers}) => {
   const typeOfProofOptions = buildTypeOfProofOptions(typeOfProofValues)
   const fieldId = (field) => buildFieldId(i, field)
 
   return (
     <React.Fragment>
-    <Row>
-      <Column span={3}>
+
+    <Row form>
+      <Column span={3} form>
         <label>HH Member on Proof</label>
         <Select
           field={fieldId('naturalKey')}
@@ -29,7 +34,7 @@ const AliceGriffithFields = ({i, householdMembers}) => {
           value={fieldId('naturalKey')}
         />
       </Column>
-      <Column span={3} end>
+      <Column span={3} form end>
         <label>Type of Proof</label>
         <Select
           field={fieldId('preferenceProof')}
@@ -38,30 +43,47 @@ const AliceGriffithFields = ({i, householdMembers}) => {
         />
       </Column>
     </Row>
-    <Row>
-      <Column span={3}>
-        <label>Alice Griffith Address</label>
-        <Text field={fieldId('address')}/>
+
+    <Row form>
+      <Column span={3} form>
+        <FormGroup>
+          <label>Alice Griffith Address</label>
+          <Text field={fieldId('address')}/>
+        </FormGroup>
       </Column>
-      <Column span={3} end>
-        <label>Apt or Unit #</label>
-        <Text/>
-      </Column>
-    </Row>
-    <Row>
-      <Column span={3}>
-        <label>City</label>
-        <Text field={fieldId('city')}/>
-      </Column>
-      <Column span={3}>
-        <label>State</label>
-        <Select field={fieldId('state')} options={[]}/>
-      </Column>
-      <Column span={3} end>
-        <label>Zip</label>
-        <Text field={fieldId('zipCode')}/>
+      <Column span={3} form end>
+        <FormGroup>
+          <label>Apt or Unit #</label>
+          <Text/>
+        </FormGroup>
       </Column>
     </Row>
+
+    <Row form>
+      <Column span={3} form>
+        <FormGroup>
+          <label>City</label>
+          <Text field={fieldId('city')}/>
+        </FormGroup>
+      </Column>
+      <Column span={3} end form>
+        <Row>
+          <Column span={6}>
+            <FormGroup>
+              <label>State</label>
+              <Text field={fieldId('state')} />
+            </FormGroup>
+          </Column>
+          <Column span={6} end>
+            <FormGroup>
+              <label>Zip</label>
+              <Text field={fieldId('zipCode')}/>
+            </FormGroup>
+          </Column>
+        </Row>
+      </Column>
+    </Row>
+
     </React.Fragment>
   )
 }
--- a/app/javascript/components/application_form/preferences/AliceGriffithFields.js
+++ b/app/javascript/components/application_form/preferences/AliceGriffithFields.js
@@ -19,8 +19,8 @@ const AddressRow = ({fieldId}) => (
     <Row form>
       <Column span={6} form>
         <FormGroup>
-          <label>Alice Griffith Address</label>
-          <Text field={fieldId('address')}/>
+          <label htmlFor='address'>Alice Griffith Address</label>
+          <Text id='address' field={fieldId('address')}/>
         </FormGroup>
       </Column>
     </Row>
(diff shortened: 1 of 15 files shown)


app/services/force/base.rb
```diff
@@ -28,6 +28,10 @@ def cache_query(q)
 
     # run a Salesforce SOQL query
     def query(q)
+      if Rails.env.development?
+        puts "SOQL>"
+        puts q
+      end
```
creativecoder, line 34: Do we want to leave this debugging code in our merge to master?

app/javascript/components/application_form/preferences/values.js
```diff
@@ -0,0 +1,40 @@
...
+  'Live or Work in San Francisco Preference': 'L_W',
+  'Neighborhood Resident Housing Preference (NRHP)': 'NRHP',
+  "Rent Burdened / Assisted Housing Preference": 'RB_AHP',
+  "Anti-Displacement Housing Preference (ADHP)": 'ADHP',
+  "Alice Griffith Housing Development Resident": 'AG'
+}
+
+export const typeOfProofValues = [
+  "Letter from SFHA verifying address",
+  "CA ID or Driver’s License",
```
creativecoder: There is a character issue in the text for this proof type. Use a regular apostrophe `'` rather than a right quote `’`. Otherwise you'll get an error from Salesforce when you select this proof type.

app/javascript/components/application_form/preferences/values.js
```diff
@@ -0,0 +1,40 @@
...
+  "Telephone bill",
```
creativecoder: This should be `Telephone bill (landline only)`

app/javascript/components/application_form/preferences/values.js
```diff
@@ -0,0 +1,40 @@
...
+  "Cable and internet bill",
+  "Paystub",
```
creativecoder: This should be `Paystub (listing home address)`

app/javascript/components/application_form/preferences/AliceGriffithFields.js
```diff
@@ -0,0 +1,84 @@
...
+    { value: option, label: option }
+  ))
+}
+
+const AddressRow = ({fieldId}) => (
+  <React.Fragment>
+    <Row form>
+      <Column span={6} form>
+        <FormGroup>
+          <label>Alice Griffith Address</label>
```
creativecoder: Best practice would be to add `for` attribute to all form labels so they are connected to their associated input fields... unfortunately a lot of existing fields are missing this.

app/javascript/components/application_form/preferences/PreferenceForm.js
```diff
@@ -1,102 +1,96 @@
...
+          <Row>
+            <Column span={6}>
+              <label>Preference</label>
+              <Select
+                field={buildFieldId(i, 'listingPreferenceID')}
+                options={listingPreferencesOptions}
+                value={buildFieldId(i,'listingPreferenceID')}
+              />
+            </Column>
+            <div className="small-offset-6 columns"/>
```
creativecoder: This looks like it doesn't belong.

app/javascript/components/application_form/preferences/utils.js
```diff
@@ -0,0 +1,4 @@
+
+export const FIELD_NAME = 'shortFormPreferences'
```
creativecoder, line 2: I'm not seeing the value of extracting this as a constant since it's only used in `PreferenceForm` right now.

creativecoder (commented):
See inline code comments.

app/javascript/components/application_form/preferences/PreferencesSection.js
```diff
@@ -23,7 +23,12 @@ let fieldMapper = {
   Individual_preference: "individualPreference",
   Certificate_Number: "certificateNumber",
   Type_of_proof: "preferenceProof",
-  Preference_Name: "preferenceName"
+  Preference_Name: "preferenceName",
+  State: "state",
+  Zip_Code: "zipCode",
+  City: 'city',
+  Street: 'address',
+  'RecordType.DeveloperName': 'recordTypeDevName'
```
creativecoder, line 32: We also need a `shortformPreferenceID` here, so that we're not creating a new application preference when fields/settings are updated.

creativecoder:
Testing and functionality is working for both admin and leasing agent user using Alice Griffith test listing.

app/services/force/base.rb
fddayan: We have it in our our feature branchs and this is already in master. this is going to be merged into lap_v1_master. I would leave it since it's very useful.

app/javascript/components/application_form/preferences/utils.js
fddayan: It's being used here https://github.com/Exygy/sf-dahlia-lap/pull/32/files#diff-09439580e565a0ba8421f21a86528a21R10

fddayan:
@creativecoder fixed

app/javascript/components/application_form/preferences/PreferencesSection.js
```diff
@@ -39,7 +45,10 @@ const PreferencesSection = ({ formApi, listingPreferences, editValues }) => {
         editPreference["naturalKey"] = naturalKey
         autofillPreferences.push(editPreference)
       }
+      console.log(preference)
```
chrisdolendo: hi @fddayan let's get rid of all the `console.log` -- how were you able to commit this? lol. Did you install overcommit?

chrisdolendo (approved):
Looks good just had that request to remove the `console.log`'s

fddayan:
@chrisdolendo console.log remove 🤦‍♂️

app/javascript/components/application_form/preferences/AliceGriffithFields.js
```diff
@@ -0,0 +1,88 @@
...
+    { value: option, label: option }
+  ))
+}
+
+const AddressRow = ({fieldId}) => (
+  <React.Fragment>
+    <Row form>
+      <Column span={6} form>
+        <FormGroup>
+          <label htmlFor='address'>Alice Griffith Address</label>
```
creativecoder: Nice--yay for better form accessibility. For this section, though, there is more than one address field on the page so `alice-griffith-address`, etc would probably be better, so as not to duplicate html ids.

app/javascript/components/application_form/preferences/AliceGriffithFields.js
```diff
@@ -0,0 +1,88 @@
...
+          <label htmlFor='hhMemberOnProof'>HH Member on Proof</label>
+          <Select
+            id='hhMemberOnProof'
+            field={fieldId('naturalKey')}
+            options={householdMembers}
+            value={fieldId('naturalKey')}
+          />
+        </Column>
+        <Column span={3} form end>
+          <label htmlFor='typeOfProof'>Type of Proof</label>
```
creativecoder: This works, though keep in mind that HTML is not case sensitive, so `typeOfProof` == `typeofproof`. Generally I prefer kebab case in HTML attrs for this reason, i.e. `type-of-proof`

creativecoder: Also, same comment as above--we may have more than one type of proof field on the page, so better to prefix with preference identifier.

app/javascript/components/application_form/preferences/PreferencesSection.js
creativecoder: @fddayan I'm not seeing `shortformPreferenceID` being submitted to Salesforce, and the application preference records are still being overwritten.

creativecoder:
Note that there is still an extraneous console.log in when you visit the edit page for an application. Added a couple of other comments inline, as well.

app/javascript/components/application_form/preferences/PreferencesSection.js
creativecoder: Looks like this is more an issue of how LAP handles preferences compared to Salesforce (overwriting with new application preferences, rather than updating the empty ones already present on the application). We can update that in a separate story rather than hold back this PR.

creativecoder (approved):
Updates look good to me! Thanks @fddayan !

PR #35, opened 2018-06-20
Features/Signs user out of Salesforce and LAP application #157983608

creativecoder:


creativecoder:
@chrisdolendo I think this is ready to deliver as soon as you're satisfied with engineering review.

chrisdolendo (approved):

PR #36, opened 2018-06-20
chores/set-compile-false-#158495929

fddayan:
https://www.pivotaltracker.com/story/show/158495929


chrisdolendo (approved):
:+1: I  also ran `bundle update sprockets` to update the Gemfile.lock

PR #40, opened 2018-06-27
Features/expandable table component #157413053

callaghanc:


callaghanc:
@fddayan After reading through the accessibility AC in the Pivotal story, I changed the table slightly to use CSS to show/hide the expanded rows, instead of using conditional rendering. The AC state that the hidden rows should have aria properties marking whether they are hidden or not, and we can't mark a row that isn't there as hidden.

app/javascript/components/molecules/Dropdown.js
```diff
@@ -102,7 +102,7 @@ class Dropdown extends React.Component {
       <button
         aria-expanded={this.state.expanded ? 'true' : 'false' }
         onClick={this.toggleExpand} ref={(node) => this.buttonRef = node }
-        className={`button dropdown-button has-icon--right text-align-left ${buttonClasses.join(' ')}`}
+        className={`button dropdown-button has-icon--right text-align-left ${buttonClasses ? buttonClasses.join(' ') : ''}`}
```
fddayan, line 105: This could be slightly improved using https://github.com/JedWatson/classnames like:
`className={classNames( 'button dropdown-button has-icon--right text-align-left', buttonClasses)}`

app/javascript/components/molecules/DropdownMenu.js
```diff
@@ -18,7 +18,7 @@ class DropdownMenu extends React.Component {
 
     return (
       <ul
-        className={`dropdown-menu ${classes.join(' ')}`}
+        className={`dropdown-menu ${classes ? classes.join(' ') : ''}`}
```
fddayan, line 21: Same thing here it could be improved with https://github.com/JedWatson/classnames like

app/component_fixtures/expandable_table.yml
```diff
@@ -0,0 +1,35 @@
+columns:
```
fddayan, line 1: I wonder if `app` folder is the right folder to put this kind of files. Should `config/component_fixtures` be a better fit ? just wondering, whatever you decide is fine.

app/javascript/components/molecules/ExpandableTable.js
```diff
@@ -0,0 +1,79 @@
...
+    super(props)
+    this.state = {row: this.props.row}
+  }
+
+  toggleExpandedRow = () => {
+    this.setState((prevState) => {
+      return {
+        row: {
+          ...prevState.row,
+          expanded: !prevState.row.expanded
```
fddayan: we work on this together, but I just realized a problem. This is confusing and might create a problem:

`this.state = {row: this.props.row}` 
and then updating the state in the row like 
`row {...., expanded: !prevState.row.expanded }`

The bad practice here is to the set `this.props.row` as a default state since changes in the parent will not be propagated down. I would leave `this.props.row` as property and create a `this.state.expanded` attribute and we track the expanded from there. Not sure if you need `expanded` outside the component, but if we do we need to create an even like on expanded. In way, the row gets cleaner since it does not need to know about the expanded state.

Usually when setting state from props is for setting default state values. The best practice is to name the property as `defaulBlaBla` so the person using that component know that subsequent changes in that property will not get reflected in the view.

fddayan (approved):
Looks good just small changes

app/javascript/components/molecules/Dropdown.js
callaghanc: thanks for the suggestion! i think in this case because the functionality is pretty trivial, i’d like to err on the side of using fewer dependencies. but good to know that package is out there in case we need more complex functionality in this area!

app/component_fixtures/expandable_table.yml
callaghanc: yeah grant and i deliberated on this for quite a while! there is no single place that’s the obvious right choice for this. we also strongly considered putting it in `config`. thinking about it, it’s not really configuration data, it’s more “data” data. we thought about adding a directory just called `data` somewhere in the directory structure, but again there wasn’t a place that such a directory seemed to fit logically and unambiguously. definitely there are multiple different places one could put this. in the end i went with `app` over `config` because the data doesn’t quite fit the category of configuration, but i still don’t feel that it’s the most graceful and natural place for it. if you can think of what its natural home is, let me know!

app/javascript/components/molecules/ExpandableTable.js
callaghanc: ah i see, thanks for catching that! yes, let’s make the change the way you described. i’ll add the change to this pr

callaghanc: ok i added my changes. now, the state of `ExpandableTableRow` only contains `expanded`. the `toggleExpandedRow` function now just updates the state to have the opposite value for `expanded`. `row` is now only referenced in `ExpandableTableRow` via the component props.

app/component_fixtures/expandable_table.yml
fddayan: no big deal at all, we could change it later if we find a better place. I would say let's leave it here for now. 👍

fddayan (approved):
👍

slowbot (commented):
@callaghanc  Just a note about updating to the latest styles. Otherwise looks good!

callaghanc:
@slowbot have updated to latest css from pl lap-leaseup

PR #48, opened 2018-07-11
Features/confirmed preferences #155196338

fddayan:
https://www.pivotaltracker.com/story/show/155196338


app/javascript/components/mappers/soqlToDomain/application_preference.js
```diff
@@ -14,12 +14,14 @@ export const mapApplicationPreference = (value) => {
     preference_name:value.Preference_Name,
     person_who_claimed_name:value.Person_who_claimed_Name,
     type_of_proof:value.Type_of_proof,
+    lw_type_of_proof: value.LW_Type_of_Proof,
     opt_out:value.Opt_Out,
     lottery_status:value.Lottery_Status,
+    post_lottery_validation:value.Post_Lottery_Validation,
```
callaghanc: @fddayan our style so far has been to put spaces after colons, would you have any objection to me reformatting this file to use that convention?

callaghanc:
@fddayan this is almost good to go, we need to update the rules about which preferences are shown in the confirmed preferences table, i'll push a commit for that soon

app/javascript/components/mappers/soqlToDomain/application_preference.js
fddayan: not at all,  go for it

callaghanc:
@fddayan Ok I updated the rule used to select which prefs to display in the Confirmed Preferences section. I'm going to add an approving review to this PR, but I would also like you to review the commits I added and add your approving review to this PR as well before we deliver it. Also, you will need to update the review instructions on the related story. I suggest having two applications, so that you can show the alternate names of the prefs that have two name possibilities.

callaghanc:
@fddayan I guess you can't technically add an actual review to this PR since it's yours, so just look over the changes I added and comment here your approval/request for changes.

callaghanc (approved):

fddayan:
@callaghanc I reviewed the commits, look good

PR #52, opened 2018-07-23
Features/confirmed preferences save #158164230

fddayan:
https://www.pivotaltracker.com/story/show/158164230


app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -1,48 +1,84 @@
...
-      { title: application.listing.name, link: appPaths.toListingLeaseUps(application.listing.id) },
-      { title: application.name, link: '#' }
-    ]
+class SupplementalApplicationPage extends React.Component {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      persistedApplication: cloneDeep(props.application),
+      application: props.application
```
akegan: setting the state equal to the props seems like pretty weird behavior. then we have this.state.application, this.state.persistedApplication, this.props.application, and this.props.formApi.values all holding the same thing, right?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -1,48 +1,84 @@
...
-      { title: 'Lease Ups', link: '/lease-ups' },
-      { title: application.listing.name, link: appPaths.toListingLeaseUps(application.listing.id) },
-      { title: application.name, link: '#' }
-    ]
+class SupplementalApplicationPage extends React.Component {
+
+  constructor(props) {
+    super(props)
+    this.state = {
+      persistedApplication: cloneDeep(props.application),
```
akegan: not immediately clear to me what "persistedApplication" is.

spec/javascript/components/applications/__snapshots__/ApplicationEditPage.test.js.snap
```diff
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`ApplicationNewPage should render succesfully 1`] = `
+exports[`ApplicationNewPage renders should render succesfully 1`] = `
```
akegan: this seems like a typo

app/javascript/components/supplemental_application/sections/preferences/RentBurdenedPanel.js
```diff
@@ -2,20 +2,20 @@ import React from 'react'
...
         Rent Burdened Housing Preference
       </div>
     </FormItem>
     <FormItem label="Total Household Monthly Rent">
-      <Text field='total_household_rent' type='number'/>
+      <Text field='total_monthly_rent' type='number'/>
     </FormItem>
     <FormItem label="Status">
-      <Select field='post_lottery_validation' options={statusOptions}/>
+      <Select field={fieldName(preferenceIndex,'post_lottery_validation')} options={statusOptions}/>
```
akegan: wasn't this buildFieldId before?

app/javascript/components/supplemental_application/sections/PreferencesTable.js
```diff
@@ -1,35 +1,37 @@
...
-const getTypeOfProof = (preference, proofFiles, fileBaseUrl) => {
-  if (overSome(isCOP, isDTHP)(preference.preference_name))
-    return preference.certificate_number
-  else
-    return getAttachments(preference, proofFiles, fileBaseUrl)
+const matchingPreference = (row) => (preference) => {
+  return getPreferenceName(preference) === row[1].content
 }
 
+/** Presenter **/
```
akegan, line 33: what does this mean? looks like it was here before, but still curious why it's here and what it means

akegan (commented):
Still wrapping my head around it, but here are a few questions to start

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: yes, kind of, we need this because of a complicated behavior for saving the preferences. setting state based on props is usually not recommended but we do not have a global state or state management and this is the root component so we are safe. I was kind of trying to make this page the store. We can talk on stand up if there is any other way to solve this problem

app/javascript/components/supplemental_application/sections/PreferencesTable.js
fddayan: I wanted to make a distinction between the presentational code and the utilities

app/javascript/components/supplemental_application/sections/preferences/RentBurdenedPanel.js
fddayan: yes, I will update that.

spec/javascript/components/applications/__snapshots__/ApplicationEditPage.test.js.snap
fddayan: 👍

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: we need to keep and unmodified copy of the persisted application in saleforce, so we can update the application (bottom save) and "individual preference" (panels) independently.

The root of the problem is that both use the same endpoint that update the application + preferences. Each individual save cannot overwrite the previous saved date, so that is why I keep I copy of the persisted data, if someone saves something we update this so the other calls can use it. 

We kind of try to solve this with the less time consuming approach. Another API end point for saving the preferences and application independently might take a bit more time to develop, since it will also involves vertiba. 

confusing? let me know on slack or we can talk on stand up

fddayan:
@akegan fixed

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -14,11 +14,11 @@ class SupplementalApplicationPage extends React.Component {
   constructor(props) {
     super(props)
     this.state = {
-      persistedApplication: cloneDeep(props.application),
-      application: props.application
+      persistedApplication: cloneDeep(props.application) // salefroce current perssited data, frozen copy. latest saved copy.
```
akegan: A couple of typos in salesforce and persisted. If we're going to commit a comment I think we want it to be a complete sentence, maybe:

"A frozen copy of the application state that is currently persisted to salesforce. This is the latest saved copy.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -1,48 +1,83 @@
...
-    breadcrumbs: [
-      { title: 'Lease Ups', link: '/lease-ups' },
-      { title: application.listing.name, link: appPaths.toListingLeaseUps(application.listing.id) },
-      { title: application.name, link: '#' }
-    ]
+  handleSaveApplication = async (application) => {
+    const { persistedApplication } = this.state
+    const synchedApplication = cloneDeep(application)
+
+    synchedApplication.total_monthly_rent = persistedApplication.total_monthly_rent
```
akegan: I still want a comment here about what is going on in these two lines.

Maybe:
"Monthly rent and preferences are only updated in handleSavePreference below."?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -1,48 +1,83 @@
...
+    await updateApplicationAction(synchedApplication)
+    this.setState({ persistedApplication: synchedApplication })
   }
 
-  const tabSection = {
-    items: [
-      { title: 'Short Form Application',    url: appPaths.toApplication(application.id) },
-      { title: 'Supplemental Information',  url: appPaths.toApplicationSupplementals(application.id) }
-    ]
+  handleSavePreference =  async (preferenceIndex, application) => {
```
akegan: any ideas for how to make it a little less confusing that this isn't an up-to-date copy of the full application?

callaghanc (dismissed):
In testing the functionality of this PR, I've found that the status values from the confirmed pref panels aren't getting successfully updated in Salesforce. I am following up with Rachel to understand the context of how those updates happen in Salesforce, so I can formulate exact change requests for what needs to be changed in this PR to make those work. I've also come up with a couple of UX questions about this PR for Jesse, and if his answers end up requiring changes, I will also add specific change requests for those.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: @akegan maybe renaming `application` here could help? `panelApplication` because it is the copy of the application updated in and coming from the panel? not the greatest name but perhaps slightly clearer

callaghanc (approved):

akegan (approved):

app/javascript/components/mappers/domainToApiFields.js
```diff
@@ -1,81 +0,0 @@
-import { shapeMapper, listMapper } from '~/utils/objectUtils'
```
callaghanc, line 1: @fddayan I have deleted this file per your mention that it is obsolete. I confirmed that it is not being used anywhere in the code before I deleted it.

callaghanc:
This PR uses the `/LeasingAgentPortal/shortForm` custom API endpoint to save updates to preferences from the Confirmed Preferences section, as well as to save updates made to the application in the supplemental application form. We have discovered that whenever we POST application data to that endpoint, that triggers a toggling effect on the application's preferences' Post-Lottery Validation fields. Every POST will toggle the value of that field from "Unconfirmed" to blank and from blank to "Unconfirmed". Due to the fact that we are planning to very soon move away from using the short form API to update preferences, we will not resolve this issue in this PR, but rather address it in an upcoming story where we will switch to using the Salesforce standard REST API for Application Preferences. This PR can therefore be merged in its current state after product acceptance of the features that are working in this PR.

PR #54, opened 2018-07-24
Chores/resolve onsubmit error for application form #159285678

fddayan:
https://www.pivotaltracker.com/story/show/159285678


chrisdolendo (approved):
Am now able to submit an application. :+1:

PR #59, opened 2018-07-27
Bugs/lease up release bugs #159153309

fddayan:


app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -56,33 +58,36 @@ class LeaseUpTableContainer extends React.Component {
 
   hideStatusModalAlert = () => this.updateStatusModal('showAlert', false)
 
-  leaseUpStatusChangeHandler = (applicationId, status) => {
+  leaseUpStatusChangeHandler = (applicationPreferenceId, applicationId, status) => {
+    console.log('leaseUpStatusChangeHandler', applicationPreferenceId, applicationId, status)
```
chrisdolendo: remove `console.log` - how did overcommit not catch this? 🤔

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -56,33 +58,36 @@ class LeaseUpTableContainer extends React.Component {
...
     if (status && comment) {
-      var data = {
+      const data = {
         status: status,
         comment: comment,
         applicationId: applicationId
       }
 
-      var response = await apiService.createLeaseUpStatus(data)
+      const response = await apiService.createLeaseUpStatus(data)
```
chrisdolendo, line 81: interesting, why did you change from var to const?

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
akegan: looks like this got left in there. Also looks like you still have pre-commit hooks disabled

akegan:
do all of these changes relate to the release bugs in the ticket? What ended up being the fix for the salesforce Error?

chrisdolendo (commented):
Not approving just yet since this branch is still wip but here is what I see

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
fddayan: as best practice, we should use `const` for all variable that cannot mutate so we can rely on the linter

fddayan: if the variable can mutate, we should use `let` and not `var`

chrisdolendo (approved):
@fddayan I'm going to deploy to lap-preprod so that jesse can review the fix on the release app so he doesn't have to review it twice. I will make a reference on the bugs for Jesse to use the lap preprod app. Nothing needed on your end here

app/javascript/components/lease_ups/LeaseUpsTable.js
```diff
@@ -9,14 +9,17 @@ import appPaths from '~/utils/appPaths'
 
 import { LEASE_UP_STATUS_OPTIONS, PAGE_SIZE, getLeaseUpStatusStyle } from './leaseUpsHelpers'
 
-const LeaseUpStatusCell = ({ cell, onChange, applicationId }) => {
+const LeaseUpStatusCell = ({ cell, onChange }) => {
+  const  applicationPreferenceId = cell.original.id
```
chrisdolendo, line 13: @fddayan reviewing this more for understanding. can you provide more context on `applicationPreferenceId` usage here? why do we use it at all when updating a lease up status?

PR #65, opened 2018-08-03
Features/application table content update #159507940

chrisdolendo:


chrisdolendo:
This feature was paired with @fddayan so counts as an engineering review - delivered story. Still welcome other reviews.

app/services/force/lease_up_service.rb
```diff
@@ -11,7 +11,7 @@ def lease_ups(listing_id)
         (SELECT Id FROM Listing_Lottery_Preference__c WHERE Listing__c = '#{listing_id}')
         AND Preference_Lottery_Rank__c != NULL
         ORDER BY Preference_Order__c, Preference_Lottery_Rank__c
-        ASC NULLS LAST LIMIT 50 OFFSET 1
+        ASC NULLS LAST LIMIT 500 OFFSET 1
```
akegan: where does this limit come from?

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -81,13 +81,15 @@ class LeaseUpTableContainer extends React.Component {
       const response = await apiService.createLeaseUpStatus(data)
 
       if (response) {
-        // find the application in question in the applications table data
-        // and update its lease up status value and status updated date
+        // find the rows with the application id whos status is being updated
```
akegan: typo whos

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
```diff
@@ -81,13 +81,15 @@ class LeaseUpTableContainer extends React.Component {
...
+        // and update their lease up status value and status updated date
         // in the table
-        const applicationIndex = findIndex(this.state.applications, { id: applicationPreferenceId })
-
-        this.updateResults(`[${applicationIndex}]['lease_up_status']`, status)
-        this.updateResults(`[${applicationIndex}]['status_updated']`, moment().format(utils.SALESFORCE_DATE_FORMAT))
+        each(this.state.applications, (app, index) => {
```
akegan, line 87: this is to update for all visible rows that have the same application, right? Does the status get updated in salesforce too? Is it possible that there would be an application/preference combo that isn't part of the state that would need to be updated?

I might just need some more info about what this is.

akegan (commented):
Could you provide a bit more context about what this piece is doing vs what the bug fix PR was updating?

app/services/force/lease_up_service.rb
chrisdolendo: In implementing this feature, I couldn't see all the preference rows for an application to test the feature. I noticed that we weren’t querying all the lease up applications, they were just 50 at a time.

Removing the limit timed out the request. So I put it up to 500 for demoing purposes for the City, and creating https://www.pivotaltracker.com/story/show/159530254, which is to implement the same pagination on lease up applications as the non-lease-up applications page that we did in this story: https://www.pivotaltracker.com/story/show/157663604. A lot of the infrastructure to paginate was done so this will be smaller in scope :slightly_smiling_face:

We’ll do it in the next sprint.

akegan: ah ok. should we add a fixme or a todo to make that a little clearer?

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
chrisdolendo: `this is to update for all visible rows that have the same application, right?` -> Yes
` Does the status get updated in salesforce too?` -> Yes `const response = await apiService.createLeaseUpStatus(data)` does the salesforce request before we do the ui change.
`Is it possible that there would be an application/preference combo that isn't part of the state that would need to be updated?` -> From a UI standpoint, no because we store all the data in state. In terms of the field update comment, we just create a new one and assign it to the application once and that comment will be the same status for all preference rows because it's the same application :)

chrisdolendo: So the business rule here is that an application can be in multiple preference buckets so it can have a preference rank for multiple preferences. It still shares the same status coz it's the same application, but if I'm paginating from COP to the DTHP ranks in this table, and the application has a COP rank and also has DTHP rank, I want to see the application row for each with the correct status. And when I update the status, I want that to reflect on all the rows correctly as well.

app/services/force/lease_up_service.rb
chrisdolendo: Good idea, will add a ToDo!

app/javascript/components/lease_ups/LeaseUpsTableContainer.js
chrisdolendo: oops :p updating.

app/services/force/lease_up_service.rb
chrisdolendo: Done.

akegan (approved):
I know you don't need an approval, but here is one anyways 😄

PR #73, opened 2018-08-09
Features/caritas add address to alternate contact #159338204

fddayan:
https://www.pivotaltracker.com/story/show/159338204


app/javascript/utils/objectUtils.js
```diff
@@ -1,6 +1,22 @@
-import { forEach, isFunction } from 'lodash'
+import {
+  forEach,
```
akegan, line 2: why not just call them all as _.forEach, etc?

akegan (commented):
@fddayan looks like the DOB PR caused merge conflicts in one of the files, otherwise looks good. going to test it now.

akegan (dismissed):
Info shows up in salesforce, but not in the application preview after you've saved the application
image
image

app/javascript/utils/objectUtils.js
fddayan: if we import individually, webpack only include in the bundle the functions that we use and not the entire lodash library. Our bundle size is smaller. More info here https://www.blazemeter.com/blog/the-correct-way-to-import-lodash-libraries-a-benchmark

akegan (approved):

PR #77, opened 2018-08-12
Features/caritas button to create new app #159246205

fddayan:
https://www.pivotaltracker.com/story/show/159246205

### DIFF
--- a/app/javascript/components/applications/ApplicationPage.js
+++ b/app/javascript/components/applications/ApplicationPage.js
@@ -10,11 +10,11 @@ import labelMapperFields from './application_details/applicationDetailsFieldsDes
 const buildActionLinkIfNecessary = (app, fromPage) => {
   const actions = []
 
-  if (!app.isLotterComplete && app.submissionType === 'Paper')
-    actions.push(<a href={appPaths.toApplicationEdit(app.id)} className='button tiny '>Edit Application</a>)
+  if (!app.is_lottery_complete && app.application_submission_type === 'Paper')
+    actions.push(<a key='edit-application' href={appPaths.toApplicationEdit(app.id)} className='primary button tiny '>Edit Application</a>)
 
   if (fromPage === 'new')
-    actions.push(<a href={appPaths.toApplicationNew(app.listing.id)} className='button tiny margin-left--half'>Add new application</a>)
+    actions.push(<a key='add-new-application' href={appPaths.toApplicationNew(app.listing.id)} className='button tiny margin-left--half'>Add new application</a>)
 
   return actions
 }
--- a/app/javascript/components/applications/ApplicationPage.js
+++ b/app/javascript/components/applications/ApplicationPage.js
@@ -20,7 +20,7 @@ const buildActionLinkIfNecessary = (app, fromPage) => {
 }
 
 const ApplicationPage = (props) => {
-  const { application, fromPage} = props
+  const { application, fromPage } = props
   const pageHeader = {
     title: `Application ${application.name}`,
     content: (<span>Name of Listing: <a href={appPaths.toListing(application.listing.id)}>{application.listing.name}</a></span>),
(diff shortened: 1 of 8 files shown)


app/javascript/components/applications/ApplicationPage.js
```diff
@@ -7,33 +7,38 @@ import mapProps from '~/utils/mapProps'
...
-const buildActionLinkIfNecessary = (app) => {
-  if (!app.isLotterComplete && app.submissionType === 'Paper') {
-    return { title: 'Edit Application', link: appPaths.toApplicationEdit(app.id) }
-  } else {
-    return undefined
-  }
+const buildActionLinkIfNecessary = (app, fromPage) => {
+  const actions = []
+
+  if (!app.is_lottery_complete && app.application_submission_type === 'Paper')
```
akegan: note this is using snake_case even though it's branched off of caritas

app/javascript/components/applications/ApplicationPage.js
```diff
@@ -7,33 +7,38 @@ import mapProps from '~/utils/mapProps'
...
+    actions.push(<a key='edit-application' href={appPaths.toApplicationEdit(app.id)} className='primary button tiny '>Edit Application</a>)
+
+  if (fromPage === 'new')
```
akegan: are we adding this button if we edit an application too? is "new" a good word to use then?

app/views/applications/show.html.slim
```diff
@@ -1,3 +1,3 @@
 / include webpacked react-table CSS
 == stylesheet_pack_tag 'react-table'
-== react_component 'ApplicationPage', { application: @application, fields: @fields, file_base_url: @file_base_url }
+== react_component 'ApplicationPage', { application: @application, fields: @fields, file_base_url: @file_base_url, fromPage: params[:from] }
```
akegan: looks like file_base_url is snake_case and fromPage is not. we should be consistent about our field naming conventions

app/javascript/components/applications/ApplicationPage.js
```diff
@@ -7,33 +7,38 @@ import mapProps from '~/utils/mapProps'
...
+    actions.push(<a key='add-new-application' href={appPaths.toApplicationNew(app.listing.id)} className='button tiny margin-left--half'>Add new application</a>)
+
+  return actions
 }
 
 const ApplicationPage = (props) => {
-  const { application } = props
+  const { application, fromPage} = props
```
akegan: missing a space before bracket here. I think the linter isn't checking this. it's on my list to update the linter

akegan (commented):
just a couple of questions @fddayan

app/javascript/components/applications/ApplicationPage.js
fddayan: It's ok here since it's using the application from the properties and not the shortForm fields

app/javascript/components/applications/ApplicationPage.js
fddayan: Not just when we come from adding a new application. Not sure about the word, it was the first thing that came to my mind. any thoughts? it could be just a flag so there is no dependency with the previous page. showAddNewButton=true

akegan (approved):

akegan:
@callaghanc or @chrisdolendo looks like this is the last caritas story up for eng review if you guys want to add another review to this one

callaghanc (changes requested):
@fddayan looks like Jesse replied to your question on the Pivotal story and answered that we should not show the "Add New Application" button after a user saves an edit. Can you update this PR to not show the button after a user saves their edits to an existing application?

fddayan:
@callaghanc fixed

chrisdolendo (approved):
Tested and code looks good to me :)  :+1:

callaghanc (approved):

PR #79, opened 2018-08-14
Features/confirmed household income #155196444

akegan:
https://www.pivotaltracker.com/story/show/155196444


app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,122 @@
...
+class ConfirmedHouseholdIncome extends React.Component {
+
+  state = {
+    hudAMI: null,
+    stateAMI: null
+  }
+
+  componentDidMount() {
+    // Display a calculated annual income if applicant only provided monthly income
+    if (isNil(this.props.application.annual_income)) {
```
akegan: @fddayan this is my proposed implementation of updating annual income based on monthly income

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,122 @@
...
+      let annualIncome = this.props.application.monthly_income * 12
+      this.props.formApi.setValue('annual_income', annualIncome)
+    }
+
+    console.log('ConfirmedHouseholdIncome did mount')
+    // Fetch initial AMI values based on household size.
```
akegan: I really do think that the AMI work should wait until another story, but here's one proposed way to do it using application state (here and in and componentDidUpdate()). Any initial thoughts on this as an idea?

AMI is just based on householdSize, not income

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,122 @@
...
+
+  state = {
+    hudAMI: null,
+    stateAMI: null
+  }
+
+  componentDidMount() {
+    // Display a calculated annual income if applicant only provided monthly income
+    if (isNil(this.props.application.annual_income)) {
+      let annualIncome = this.props.application.monthly_income * 12
```
akegan: honestly, I'm a bit confused about when to reference props of the component vs formApi.values. should I ever be accessing props?

akegan (commented):
@fddayan hopefully this code helps explain what I was saying this morning about updating the annual income based on monthly income.

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
fddayan: let me see if I follow te requirements correctly
1. We want `anual_income` to be calculated based from other fields
1. However, we want to let the user to set any free text in `anual_income`. Calculated value will be set only if `anual_income` is not present

I feel we can do all this in the `render` function and there is no need for `componentDidMount` or `componentDidUpdate`. 
```
render():
if annual_income is empty 
  formApi.setValue(anual_income,dervied_value_from_fancy_fields())
end
```
  
about using set `fromApi.setValue`, yes, I would use that to set the value (...and that will update the input text)

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
fddayan: I would say that, `formApi.values` (and setValue) should be used when you are dealing with the form. 
Based on how you are doing it in the this code,  you should use from.values. Values from the form and props might reference different objects and an update on a Text field will not be reflected in the props.

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
fddayan: I think we do not need `componentDidUpdate` and it can be handled in the `render()` ( like anual_income)

.overcommit.yml
```diff
@@ -39,6 +39,8 @@ PreCommit:
   RailsBestPractices:
     enabled: true
     description: 'Checking Rails best practices'
+    include:
```
akegan, line 42: FYI @callaghanc @fddayan @chrisdolendo this is because one of the best practice rules, "no unused methods" requires that the entire scope of the directory is included to check for usage. If you run rails best practices on a subdir, then it doesn't check outside of that dir for usages of methods.

app/services/force/base.rb
```diff
@@ -66,7 +66,7 @@ def builder
       Force::SoqlQueryBuilder.new(@client)
     end
 
-    def parsed_index_query(q, type = :index)
+    def parsed_index_query(q, _type = :index)
```
akegan, line 69: required by rubocop

app/services/force/base.rb
```diff
@@ -83,7 +83,11 @@ def api_post(endpoint, params)
       @api.post(endpoint, params)
     end
 
-    private
+    def api_get(endpoint, params)
+      @api.get(endpoint, params)
+    end
+
+    private_class_method
```
akegan, line 90: rubocop

app/services/force/base.rb
```diff
@@ -116,7 +120,7 @@ def massage(h)
       Force::Responses.massage(h)
     end
 
-    def self.load_fields(name, file_path = nil)
+    def self.load_fields(name, _file_path = nil)
```
akegan, line 123: rubocop

app/services/force/listing_service.rb
```diff
@@ -27,9 +27,20 @@ def subqueries
       %(
       	(SELECT Date__c, Start_Time__c, End_Time__c, Venue__c, Street_Address__c, City__c FROM Information_Sessions__r),
       	(SELECT Date__c, Start_Time__c, End_Time__c FROM Open_Houses__r),
-      	(SELECT Id, Total_Submitted_Apps__c, Order__c, Description__c, Available_Units__c, PDF_URL__c, Lottery_Preference__r.Id, Lottery_Preference__r.Name FROM Listing_Lottery_Preferences__r),
-      	(SELECT Unit_Type__c, BMR_Rent_Monthly__c, BMR_Rental_Minimum_Monthly_Income_Needed__c, Status__c, Property_Type__c, AMI_chart_type__c, AMI_chart_year__c, of_AMI_for_Pricing_Unit__c, Reserved_Type__c FROM Units__r ORDER BY Unit_Type__c)
+      	(SELECT Id, Total_Submitted_Apps__c, Order__c, Description__c, Available_Units__c, PDF_URL__c, Lottery_Preference__r.Id,
```
akegan, line 30: just line length changes due to rubocop

app/services/force/listing_service.rb
```diff
@@ -27,9 +27,20 @@ def subqueries
...
+                Lottery_Preference__r.Name FROM Listing_Lottery_Preferences__r),
+      	(SELECT Unit_Type__c, BMR_Rent_Monthly__c, BMR_Rental_Minimum_Monthly_Income_Needed__c, Status__c, Property_Type__c,
+                AMI_chart_type__c, AMI_chart_year__c, of_AMI_for_Pricing_Unit__c, Reserved_Type__c FROM Units__r ORDER BY Unit_Type__c)
       )
     end
+
+    def units(listing_id)
+      parsed_index_query(%(
+        SELECT Unit_Type__c, BMR_Rent_Monthly__c, BMR_Rental_Minimum_Monthly_Income_Needed__c, Status__c, Property_Type__c,
```
akegan, line 39: @callaghanc after all that i went with just returning straight up units because we already had a map units function and I'm pretty sure we'll end up needing the rest of those fields further down the line. I also used Unit__c instead of listing_unit__c so it would have all the same fields we're pulling in with Listing above. 

Is there a strong reason for using Listing_Unit__c instead?

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+    return ami.amount
+  }
+}
+
+const getAMIPercent = ({income, ami}) => {
+  if (!income) {
+    return "Enter HH Income"
+  }
+  if (isNil(ami)) {
+    return "Error: AMI unknown"
```
akegan: this wasn't part of the initial mockups @slowbot but I don't want to hold up rendering of the whole page while waiting for the AMI to load. It should only show up as this message for <1s. Any preferences for what it should say?

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+      <React.Fragment>
+      <FormGrid.Row>
+        <FormGrid.Item>
+          <FormGrid.Group label="Recurring Voucher/Subsidy">
+            <YesNoRadioGroup field="housing_voucher_or_subsidy" />
+          </FormGrid.Group>
+        </FormGrid.Item>
+        <FormGrid.Item>
+          <FormGrid.Group label="Household Assets">
+            <Text field="household_assets"
```
akegan: @callaghanc or @fddayan do you know how to display the value with a '$', while not actually sending a string with '$' to salesforce?

akegan (commented):
This is pretty much feature complete (finally).

Remaining TODOs:
1. Format percentage nicely
2. Figure out how to format dollars nicely without affecting the values that are sent to salesforce.

.overcommit.yml
fddayan: 👍

app/controllers/api/v1/ami_controller.rb
```diff
@@ -0,0 +1,17 @@
+module Api::V1
+  # API to query for Area Median Income values
+  class AmiController < ApiController
+    def get
```
fddayan, line 4: @akegan could this be refactor to use the `resource` actions. In this case show (or index not sure what it does) ?
perhaps using singular resource if there is nothing that can be used as an ID https://guides.rubyonrails.org/routing.html#singular-resources

akegan: I had it as a resource action initially, but in this case it's not really a proper rest resource. the queries are

/ami?chartType='blah'&year='blah' it didn't really seem like an index to me, and it doesn't have an id so doesn't make sense as a show either.

app/services/force/ami_service.rb
```diff
@@ -0,0 +1,13 @@
+module Force
+  # encapsulate all Salesforce AMI querying functions
+  class AmiService < Force::Base
+    def ami(params)
+      response = api_get('/ami', params)
+      amis = []
+      response.each do |ami|
+        amis.push(chartType: ami[:chartType], year: ami[:year], householdSize: ami[:numOfHousehold], amount: ami[:amount])
```
fddayan: suggestion: format this to multi line like
```
{ 
 chartType: ami[:chartType]
....
}
```
OR
can you return `numOfHousehold` as the key instead of `householdSize` ? and then you can just use `ami.slice(:chartType, :year, :numOfHousehold,...)`

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -33,17 +33,19 @@ const SupplementalApplicationPage = ({ application, statusHistory, formFields, o
...
 
-const mapProperties = ({application, statusHistory, file_base_url}) => {
+const mapProperties = ({application, statusHistory, file_base_url, units}) => {
   return {
     application: mapApplication(application),
     statusHistory: mapList(mapFieldUpdateComment,statusHistory),
     onSubmit:  (values) => updateApplicationAction(values),
-    fileBaseUrl: file_base_url
+    fileBaseUrl: file_base_url,
+    units: units.map(unit => mapUnit(unit))
```
fddayan: I would use `mapList` or `map` from lodash to avoid null pointer as a best practice.

Look in this same shape how mapList is being used

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+          </FormGrid.Group>
+        </FormGrid.Item>
+      </FormGrid.Row>
+      <FormGrid.Row paddingBottom>
+        {this.state.amiCharts.map((chart) => {
+          let ami = getAMI({numHousehold: formApi.values.total_household_size, chartName: chart.name, chartYear: chart.year, amis: amis})
+          return (
+            <FormGrid.Item key={chart.name + chart.year}>
+            <FormGroupTextValue label={`Calculated % of AMI - ${chart.name}`}
+                                id={`ami-${chart.name}`}
```
fddayan: `ami-${chart.name}` could be extract out as a variable to avoid repetition

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+
+  componentDidMount() {
+    // Display a calculated annual income if applicant only provided monthly income
+    if (isNil(this.props.formApi.values.annual_income) && !isNil(this.props.formApi.values.monthly_income)) {
+      let annualIncome = this.props.formApi.values.monthly_income * 12
+      this.props.formApi.setValue('annual_income', annualIncome)
+    }
+
+    // Determine which unique AMI charts are associated with units on the listing
+    let chartsToLoad = uniqBy(this.props.units, (u) => [u.ami_chart_type, u.ami_chart_year].join())
```
fddayan: I would use  `const chartsToLoad` instead of `let chartsToLoad`
Instead of (u) => you can do u =>  ...  () are optioal for single parameteres, however we do not have a style guide so it's up to you

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
+import React from 'react'
+import { find, isNil, uniqBy } from 'lodash'
+
+import FormGrid from '~/components/molecules/FormGrid'
+import FormGroupTextValue from '~/components/atoms/FormGroupTextValue'
+import { RadioGroup, Radio, Text } from 'react-form';
+import { getAMIAction } from '~/components/supplemental_application/actions'
```
fddayan: As a best practice I would call the actions to just in the `Page`. So we push all the state, mutations and actions to the top and can manages the state for the page.

In this practice all components read the state from the parent and modify data thru callbacks to the parent.

We never really talked about this. The application is growing in complexity and it's a good time to talk about patterns/practices etc

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+class ConfirmedHouseholdIncome extends React.Component {
+
+  state = {
+    amis: {},
+    amiCharts: []
+  }
+
+  componentDidMount() {
+    // Display a calculated annual income if applicant only provided monthly income
+    if (isNil(this.props.formApi.values.annual_income) && !isNil(this.props.formApi.values.monthly_income)) {
```
fddayan: I would 
```
const { monthly_income, annual_income } = this.props.formApi.values
```

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+
+
+class ConfirmedHouseholdIncome extends React.Component {
+
+  state = {
+    amis: {},
+    amiCharts: []
+  }
+
+  componentDidMount() {
```
fddayan: We talked about this and you showed me some code, I think now I understand better.

calling `componentDidMount` does not look right. I feel it should be called in the top level component (the Page component) if you have time we can pair review this and you can walk me thru the review app.

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+    let chartsToLoad = uniqBy(this.props.units, (u) => [u.ami_chart_type, u.ami_chart_year].join())
+
+    this.setState({'amiCharts': chartsToLoad.map((chart) => {
+      return {"name": chart.ami_chart_type, "year": chart.ami_chart_year}}
+      )
+    })
+
+    // Load the state with ami values for each chart associated with the listing.
+    chartsToLoad.forEach( (chart) =>
+      getAMIAction({chartType: chart.ami_chart_type, chartYear: chart.ami_chart_year}).then(response => {
```
fddayan: I feel this action should be call in the parent component (the Page, that holds the state for the entire page)

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+    return "Error: AMI unknown"
+  }
+  return formatPercent(income / Number(ami))
+}
+
+
+
+class ConfirmedHouseholdIncome extends React.Component {
+
+  state = {
```
fddayan: I feel this state belongs to the Page

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -0,0 +1,137 @@
...
+            <Text field="hh_total_income_with_assets_annual"
+                  placeholder="Enter Amount"
+            />
+            <span className="form-note shift-up" id="final-household-annual-income">Includes % of assets if applicable</span>
+          </FormGrid.Group>
+        </FormGrid.Item>
+      </FormGrid.Row>
+      <FormGrid.Row paddingBottom>
+        {this.state.amiCharts.map((chart) => {
+          let ami = getAMI({numHousehold: formApi.values.total_household_size, chartName: chart.name, chartYear: chart.year, amis: amis})
```
fddayan: I would extract all `formApi.values` on top to make this code a bit more readable.

app/javascript/utils/utils.js
```diff
@@ -8,3 +8,8 @@ export default {
   cleanField,
   SALESFORCE_DATE_FORMAT,
 }
+
+
+export const formatPercent = (value) => {
```
fddayan, line 15: about the name of the file 'utils/utils` we should a TODO to fix this name not sure if it's says much

akegan: yeah we have a lot of util files, I'm not sure what goes where.

akegan:
Fixed all of the comments that I responded to with thumbs up, that don't involve moving things around. I'll move state from confirmed household income to supplementalApplicationPage on monday morning

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
fddayan: I do think it can be done without building  custom component. I would strip the $ in the action before calling the the aapi

Is this an option ? we have to ask jesse
image

fddayan (changes requested):
Waiting to talk about the state in componets

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
akegan: I think we ended up deciding to move it into another ticket.

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
akegan: do you think it belongs in componentDidMount in the supplemental application page too?

app/services/force/listing_service.rb
callaghanc: According to Rachel we should be using Listing Unit when our goal is to get all units that belong to a listing. The Listing Unit object is currently used to represent the relationship between a Listing and a Unit. However, it is then unclear why Units have a field for the listing ID - that makes it seem like the relationship object isn't actually needed. This raises the question of whether the true relationships between listings and units are represented by a) the Listing Unit objects, or b) the listing id field on the Unit object. We haven't dealt directly with this that much yet because the web app uses a custom API call to fetch listing units. I'm following up further with Rachel on this to understand the exact distinctions and to see if there is duplicate functionality present that could be removed to clarify. Until then, my inclination would be to use Rachel's recommendation to use Listing Unit until we have more info suggesting otherwise, and just query and pass the fields needed to fulfill the AC in this story.

callaghanc: Have not heard any more info from Rachel yet, but I looked further into where Unit and Listing Unit are used, and I found that Leases appear to only have a relationship with Units, not with Listing Units. This doesn't say anything about whether the listing-unit relationship is managed through Listing Units or through a `Unit__c` field. But it does provide another example of where it seems to be more convenient to use Units directly. Based on this, I now feel fine with sticking with using Units directly in this PR. I will still follow up with Rachel so we can get clarity on this issue.

akegan: looks like mapUnits expects all of those fields to be there :(

app/controllers/api/v1/ami_controller.rb
akegan: @fddayan any more feedback here? does it still seem like it should be an index to you?

app/services/force/listing_service.rb
callaghanc: ok then let's go ahead with including those fields

akegan:
@fddayan ready for review again!

fddayan (approved):
Looks good 👍

PR #96, opened 2018-09-10
Bugs/preferences type of proof #160361540

akegan:
https://www.pivotaltracker.com/n/projects/1405352/stories/160361540

another snake case vs camel case bug


fddayan (approved):
👍

PR #101, opened 2018-09-14
Features/confirmed preferences panel save ux behavior #159877262

fddayan:


fddayan:
Closing so that we can re-create after cleaning the repo of VCR cassette data.

fddayan:
Closing so that we can re-create after cleaning the repo of VCR cassette data.

PR #104, opened 2018-09-17
Features/confirmed preferences panel save ux behavior #159877262

fddayan:


akegan:
FYI @fddayan semaphore is failing on this PR. Is it ready for review?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -152,33 +153,34 @@ class SupplementalApplicationPage extends React.Component {
...
-    const { statusHistory, fileBaseUrl, application, availableUnits } = this.props
+    const { statusHistory, fileBaseUrl, availableUnits } = this.props
     const {
       confirmedPreferencesFailed,
       amis,
       amiCharts,
       statusModal,
-      loading
+      loading,
+      persistedApplication
```
callaghanc, line 173: @akegan Is this perhaps the same move of application from props to state that you and I discussed at end of yesterday on your lease preference used PR?

akegan: looks like it might be! I'll sync with fed on this now

app/javascript/components/supplemental_application/sections/preferences/LiveOrWorkInSanFranciscoPanel.js
```diff
@@ -25,12 +25,14 @@ const getLWPrefProofTypeOptions = (individualPrefName) => {
 class LiveOrWorkInSanFranciscoPanel extends React.Component {
   constructor (props) {
     super(props)
+    // NOTE: I do not think we need a state for this, Fed
```
callaghanc: @fddayan Let's save refactor of this area for the upcoming story we have that addresses all uses of type of proof in the Confirmed Preferences section.

akegan:
@fddayan is this ready to go now?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -74,19 +74,35 @@ class SupplementalApplicationPage extends React.Component {
...
+    const { persistedApplication } = this.state
     const responses = await Promise.all([
       updateTotalHouseholdRent(application.id, application.total_monthly_rent),
       updatePreference(application.preferences[preferenceIndex])
     ])
     const failed = some(responses, response => response === false)
 
-    this.setState({ confirmedPreferencesFailed: failed })
+    if (!failed) {
+      const updateApplication = cloneDeep(persistedApplication)
```
akegan: typo. should be updatedApplication

callaghanc:
@akegan Yes, this is ready. I'm in the midst of reviewing it, and would like to be able to complete my PR review in addition to yours, and to complete my testing of the behavior on the related review app before we deliver this story.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -74,18 +74,35 @@ class SupplementalApplicationPage extends React.Component {
   }
 
   handleSavePreference = async (preferenceIndex, application) => {
```
callaghanc: @fddayan @akegan How do you guys feel about changing the `application` argument name here to something like `formApiApplicationValues`, `formValues`, `formApplicationVals`, etc.? Because ultimately `formApi.values` is what is getting passed to this function as the `application`. Renaming would help indicate this.

akegan: sounds great to me! I am definitely easily confused with the different applications we're passing around right now

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -74,18 +74,35 @@ class SupplementalApplicationPage extends React.Component {
...
 
-    this.setState({
-      confirmedPreferencesFailed: some(responses, response => response === false)
-    })
+    return !failed
   }
 
-  handleOnDismissError = () => {
-    this.setState({ confirmedPreferencesFailed: false })
+  handleCancelPreferencePanel = (preferenceIndex) => {
```
callaghanc: @fddayan Let's take a look at this function tomorrow in check-in - there may be some optimization we can do here.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: sure we can do that. Just one comment. The shape is the same, so at the end of the day is an application. Naming the var formApi in the Page makes the page kind of "award" of the underline implementation. I feel the contract/spec of that method is that it will get an application shape regardless of the underline implementation.
however...not a big deal, sharing a thought.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: actually, this has some unnecessary code. I'm pushing a new version.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: @fddayan has a very good point and that's the major downside to renaming the argument. However, due to the confusion we've been having with all the different applications being passed around, this change could help readability enough that it's worth the tradeoff. We'll try it and see how it goes.

callaghanc (approved):

PR #106, opened 2018-09-17
Features/paginate leaseup applications #159530254

akegan:
https://www.pivotaltracker.com/story/show/159530254

Still in progress, but ready for an initial review

Remaining tasks:
- [x] How to prevent SOQL injection - *[Ticket made](https://www.pivotaltracker.com/story/show/160629818)*
- [x] Resolve whether or not we need to be able to control the number of rows per page - *This appears to not be important*
- [x] Determine if we need sorting in the lease up applications table, disable if not, implement if yes - *This is the reason for postponing this PR*
- [ ] Fix LeaseUpsPage snapshot test, need to figure out how to mock the handleOnFetchData piece

### DIFF
--- a/app/controllers/api/v1/lease_up_applications_controller.rb
+++ b/app/controllers/api/v1/lease_up_applications_controller.rb
@@ -1,0 +1,18 @@
+module Api::V1
+  # Lease Up Applications controller for access via the API
+  class LeaseUpApplicationsController < ApiController
+    def index
+      applications = lease_up_service.lease_up_listing_applications(lease_up_applications_params)
+      render json: applications
+    end
+
+    def lease_up_service
+      Force::LeaseUpService.new(current_user)
+    end
+
+    def lease_up_applications_params
+      # TODO: allow page as param
+      params.require(:listing_id)
+    end
+  end
+end
--- a/app/controllers/api/v1/lease_up_applications_controller.rb
+++ b/app/controllers/api/v1/lease_up_applications_controller.rb
@@ -12,9 +12,7 @@ module Api::V1
 
     def lease_up_applications_params
       # TODO: allow page as param
-      puts 'PARAMS', params
       params.require(:listing_id)
-            .permit(:page)
     end
   end
 end
--- a/app/controllers/api/v1/lease_up_applications_controller.rb
+++ b/app/controllers/api/v1/lease_up_applications_controller.rb
@@ -12,7 +12,9 @@ module Api::V1
 
     def lease_up_applications_params
       # TODO: allow page as param
+      puts 'PARAMS', params
       params.require(:listing_id)
+            .permit(:page)
     end
   end
 end
--- a/app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
+++ b/app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
@@ -9,6 +9,7 @@ import appPaths from '~/utils/appPaths'
 import apiService from '~/apiService'
 import EagerPagination from '~/utils/EagerPagination'
 import Context from './context'
+import mapProps from '~/utils/mapProps'
 
 class LeaseUpApplicationsPage extends React.Component {
   state = { loading: false, applications: [], pages: 0 }
@@ -19,8 +20,7 @@ class LeaseUpApplicationsPage extends React.Component {
   }
 
   fetchApplications = async (page) => {
-    // FIXME, need to figure out a way to map listing to get Id
-    const response = await apiService.fetchLeaseUpApplications(this.props.listing.Id, page)
+    const response = await apiService.fetchLeaseUpApplications(this.props.listing.id, page)
     return {
       records: !isEmpty(response.records) ? response.records.map(flow(mapApplicationPreference, buildLeaseUpModel)) : [],
       pages: response.pages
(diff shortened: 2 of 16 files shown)


app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -0,0 +1,18 @@
...
+      render json: applications
+    end
+
+    def lease_up_service
+      Force::LeaseUpService.new(current_user)
+    end
+
+    def lease_up_applications_params
+      # TODO: allow page as param
+      params.require(:listing_id)
```
akegan, line 15: @callaghanc you might be able to help me with this. if I put

.permit(:page) in there, then I get the error ```NoMethodError - undefined method `permit' for "<<listing_id>>":String:```

app/services/force/lease_up_service.rb
```diff
@@ -12,25 +12,28 @@ def lease_up_listings
...
-      application_data = massage(query(%(
-        SELECT #{query_fields(:lease_up_listing_application)}
-        FROM Application_Preference__c
-        WHERE Listing_Preference_ID__c IN
-        (SELECT Id FROM Listing_Lottery_Preference__c WHERE Listing__c = '#{listing_id}')
-        AND Preference_Lottery_Rank__c != NULL
-        ORDER BY Preference_Order__c, Preference_Lottery_Rank__c
-        ASC NULLS LAST LIMIT 500
-      )))
+    def lease_up_listing_applications(listing_id, opts = { page: 0 })
```
akegan, line 15: i'm also curious if we need to add any sort of sanitization here to avoid soql injection

akegan:
update: still working on mocking for snapshots to get the tests to make sense

app/controllers/api/v1/lease_up_applications_controller.rb
fddayan: how are you adding permit ?  can you paste the code ?

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -1,39 +1,81 @@
...
+import EagerPagination from '~/utils/EagerPagination'
+import Context from './context'
+import mapProps from '~/utils/mapProps'
+
+class LeaseUpApplicationsPage extends React.Component {
+  state = { loading: false, applications: [], pages: 0 }
+
+  constructor (props) {
+    super(props)
+    this.eagerPagination = new EagerPagination(20, 100)
```
fddayan: Adding a quick note here that we might want to extract 20 and 100 to constants (probably in a reusable file)

app/services/force/lease_up_service.rb
fddayan: We probably should. We can capture this in different story

fddayan:
@akegan so far looks great. 
Remember to remove fetching the applications from the controller, since now you are doing it from the API using ajax. https://github.com/Exygy/sf-dahlia-lap/blob/188ed94981a52f4ac04a90479f6070343cc2f393/app/controllers/listings/lease_ups/applications_controller.rb#L12

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -1,39 +1,81 @@
...
+  }
+
+  fetchApplications = async (page) => {
+    const response = await apiService.fetchLeaseUpApplications(this.props.listing.id, page)
+    return {
+      records: !isEmpty(response.records) ? response.records.map(flow(mapApplicationPreference, buildLeaseUpModel)) : [],
```
fddayan: if you use lodash map, you do not have worry about checking if the value is empty or null, it returns an array always: `map(null, someFunction) => []`

so you can do:
`map(response.records, flow(mapApplicationPreference, buildLeaseUpModel))`

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -1,39 +1,81 @@
...
-      link: `/listings/${listing.id}/lease-ups/export`
-    },
-    breadcrumbs: [
-      {title: 'Lease Ups', link: appPaths.toLeaseUps()},
-      {title: listing.name, link: appPaths.toLeaseUpApplications(listing.id)}
-    ]
+  loadPage = (page) => {
+    const fetcher = p => this.fetchApplications(p)
+    this.setState({ loading: true, page: page })
+    this.eagerPagination.getPage(page, fetcher).then(({ records, pages }) => {
```
fddayan: I know I did the same thing in ApplicationsPage and you got the code from there, but I'm wondering if this might be more readable:
make loadPage `async` and 
```
const { records, pages } = await this.eagerPagination.getPage(page, fetcher) 
this.setState({ applications: records, loading: false, pages: pages })
```

app/controllers/api/v1/lease_up_applications_controller.rb
akegan: ```params.require(:listing_id).permit(:page)```

app/services/force/lease_up_service.rb
akegan: https://www.pivotaltracker.com/story/show/160629818

akegan:
Closing this because it has been moved out of this sprint. The only remaining task is to verifying that the mock data is loading correctly in the test and then update the snapshots.

PR #109, opened 2018-09-19
Features/lease salesforce saving #158157958

akegan:
https://www.pivotaltracker.com/n/projects/1405352/stories/158157958

Outstanding items

- [x] Move saving Tenant to separate story, or resolve salesforce blocker -*Moved*
- [x] Re-pull application data on supplemental application save to avoid creating multiple leases. As it is currently, we don't get the lease Id back on save, so every save without refreshing the page will lead to a new lease object being created. *Added a window href to reload the page, TBC whether that's a good solution*

### DIFF
--- a/app/controllers/api/v1/short_form_controller.rb
+++ b/app/controllers/api/v1/short_form_controller.rb
@@ -49,6 +49,7 @@ class Api::V1::ShortFormController < ApiController
             lease: %i[
               id
               unit
+              leaseStatus
               leaseStartDate
               monthlyParkingRent
               totalMonthlyRentWithoutParking
--- a/app/controllers/api/v1/short_form_controller.rb
+++ b/app/controllers/api/v1/short_form_controller.rb
@@ -6,12 +6,10 @@ class Api::V1::ShortFormController < ApiController
     logger.debug "application_api_params: #{application_api_params}"
     short_form_validator = ShortFormValidator.new(application_api_params)
     if short_form_validator.valid?
-      # If lease information is available, submit it separately.
       if application_api_params.key?(:lease)
         response = lease_service.submit_lease(application_api_params[:lease],
                                               application_api_params[:id],
                                               application_api_params[:primaryApplicantContact])
-        # Don't submit lease application_api_params to the customAPI.
         logger.debug "lease submit response: #{response}"
       end
       application = application_service.submit(application_api_params)
--- a/app/controllers/api/v1/short_form_controller.rb
+++ b/app/controllers/api/v1/short_form_controller.rb
@@ -45,6 +45,14 @@ class Api::V1::ShortFormController < ApiController
             :numberOfDependents,
             :formMetadata,
             :hasSenior,
+            lease: %i[
+              id
+              unit
+              leaseStartDate
+              monthlyParkingRent
+              totalMonthlyRentWithoutParking
+              monthlyTenantContribution
+            ],
             primaryApplicant: %i[
               contactId
               appMemberId
--- a/app/controllers/api/v1/short_form_controller.rb
+++ b/app/controllers/api/v1/short_form_controller.rb
@@ -6,6 +6,14 @@ class Api::V1::ShortFormController < ApiController
     logger.debug "application_api_params: #{application_api_params}"
     short_form_validator = ShortFormValidator.new(application_api_params)
     if short_form_validator.valid?
+      # If lease information is available, submit it separately.
+      if application_api_params.key?(:lease)
+        response = lease_service.submit_lease(application_api_params[:lease],
+                                              application_api_params[:id],
+                                              application_api_params[:primaryApplicantContact])
+        # Don't submit lease application_api_params to the customAPI.
+        logger.debug "lease submit response: #{response}"
+      end
       application = application_service.submit(application_api_params)
       logger.debug "application submit response: #{application}"
       render json: { application: application }
@@ -166,5 +174,9 @@ class Api::V1::ShortFormController < ApiController
 
   def application_service
     Force::ApplicationService.new(current_user)
+  end
+
+  def lease_service
+    Force::LeaseService.new(current_user)
   end
 end
--- a/app/controllers/api/v1/short_form_controller.rb
+++ b/app/controllers/api/v1/short_form_controller.rb
@@ -45,6 +45,7 @@ class Api::V1::ShortFormController < ApiController
             :numberOfDependents,
             :formMetadata,
             :hasSenior,
+            :primaryApplicantContact,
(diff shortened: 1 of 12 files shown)


akegan:
@fddayan I'm not sure how to re-populate the application on application save, do you have any ideas? I'm thinking that we may need to have an explicit action to fetch the application from the API rather than passing it in from the slim template. 

**Update**: See comment below for possible solution

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -60,6 +60,8 @@ class SupplementalApplicationPage extends React.Component {
 
     await updateApplicationAction(synchedApplication)
     this.setState({ persistedApplication: synchedApplication })
+    // Redirect to same page to be able to re-pull data from SalesForce
+    window.location.href = appPaths.toApplicationSupplementals(application.id)
```
akegan: @fddayan @chrisdolendo this is a possible solution for repopulating the application data on save.

a couple of initial concerns I have around this: 
1. What if the fields haven't actually been updated by the time we query the API? Is that an issue we've had in the past?
2. This messes with the synchedApplications state piece, and might actually make it obsolete. Does it also mess with the confirmed preferences piece?

app/services/force/application_service.rb
```diff
@@ -77,9 +78,44 @@ def update(data)
...
     def submit(data)
+      # If lease information is available, submit it separately.
+      if data.key?(:lease)
+        submit_lease(data[:lease], data[:id], data[:primaryApplicantContact])
+        data = data.except(:lease) # Don't submit lease data to the customAPI.
+      end
       api_post('/LeasingAgentPortal/shortForm', application_defaults.merge(data))
     end
 
+    def submit_lease(lease, application_id, primary_contact_id)
```
akegan: comment from fed: move this to a lease service

app/services/force/application_service.rb
```diff
@@ -45,6 +45,7 @@ def lease(application_id)
              .select(:Id,
                      :Unit__c,
                      :Lease_Start_Date__c,
+                     :Lease_Status__c,
```
akegan: move this to a separate service too. Lease Service.

Combine services in the controller

app/services/force/application_service.rb
```diff
@@ -77,9 +78,44 @@ def update(data)
     end
 
     def submit(data)
+      # If lease information is available, submit it separately.
+      if data.key?(:lease)
```
akegan: move this logic to the controller

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -60,6 +60,8 @@ class SupplementalApplicationPage extends React.Component {
 
     await updateApplicationAction(synchedApplication)
```
akegan: check the response here, if it's false, don't do the redirect.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: this seems to work OK, and was also approved as a valid workaround by jesse

akegan:
Lease stuff has been moved to its own service. Ready for review again

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: what about using `window.location.reload()` instead ?

app/services/force/lease_service.rb
```diff
@@ -0,0 +1,48 @@
...
+                     :Total_Monthly_Rent_without_Parking__c,
+                     :Monthly_Tenant_Contribution__c)
+             .where_eq(:Application__c, application_id, :string)
+             .transform_results { |results| massage(results) }
+             .query
+             .records
+             .first
+    end
+
+    def submit_lease(lease, application_id, _primary_contact_id)
```
fddayan: I feel this method might be a bit too long,  what about having a 

`create_lease` and `update_lease`

and then in `submit_lease`

```
if lease[:id]
   update_lease(lease)
else
  create_lease(lease)
end
```

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -6,6 +6,14 @@ def submit
     logger.debug "application_api_params: #{application_api_params}"
     short_form_validator = ShortFormValidator.new(application_api_params)
     if short_form_validator.valid?
+      # If lease information is available, submit it separately.
+      if application_api_params.key?(:lease)
+        response = lease_service.submit_lease(application_api_params[:lease],
```
fddayan: what do we get in the response here? should be return this in the json response the same way we do it for the application?

akegan: If it's a create, we get the lease id or False, if it's an update we get True or False

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -6,6 +6,14 @@ def submit
     logger.debug "application_api_params: #{application_api_params}"
     short_form_validator = ShortFormValidator.new(application_api_params)
     if short_form_validator.valid?
+      # If lease information is available, submit it separately.
```
fddayan: I think this comment is not necessary, it's pretty clear what is happening, at least to me

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: @fddayan I think that if there's an error an alert is raised, I don't think that it actually responds "false" if the update is unsuccessful.

app/controllers/api/v1/short_form_controller.rb
fddayan: I see ok. then no changes here.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: it looks like it returns  the response from the service and raises a `window.alert`. After you dismiss the alert it will redirect. I feel it would be good to NOT redirect if there is an error. 

form the code is doing 
```
  if (response === false) {
    Alerts.error()
  }
```

so if this code is correct, it's returning `false`

you can do 
```
if (response) {
 window.location.reload()
}
```
makes sense?

akegan: thanks, I missed that check for false 🤦‍♀️ . Fixed

fddayan (approved):
looks good 👍

PR #110, opened 2018-09-20
Chores/saleforce rspec tests #157163127

fddayan:


app/services/force/access_token.rb
```diff
@@ -0,0 +1,95 @@
...
+
+    def self.restforce_client
+      Restforce.new(request_new_with_restforce_attributes)
+    end
+
+    ####################################################
+    # Request new token unsing HTTP post with oauth2
+    ####################################################
+
+    def self.request_new_with_http
```
chrisdolendo: Where is this method being used?

chrisdolendo (commented):
I have a question on what looks like an unused method `Force::AccessToken#request_new_with_http`.

Looked at the cassettes and these look secure to me:
1. `spec/vcr/applications_controller/show.yml` - shows a Lappy Laptest application which is a test application.
2. `spec/vcr/listings/applications_controller/index.yml` - retrieves listings data that's meant to be public, and only requests first name, last name, lottery date, app submitted date, submission type, total household size, and application ID, which aren't high risk data exposure
3. `spec/vcr/listings/applications_controller/index_errors.yml` - refers to unsuccessful response
4.  `spec/vcr/listings/applications_controller/new.yml` - retrieves listing data

app/services/force/access_token.rb
fddayan: This is the old way we were getting the token. I thought about leaving the code as a reference in case we might need it in the future. We can remove it, it's not being used.

chrisdolendo: Gotcha. Let's remove unused code to keep maintainability.

fddayan:
@chrisdolendo ready

callaghanc (approved):

PR #111, opened 2018-09-21
Features/add status change supp app #155196224

fddayan:


app/javascript/components/supplemental_application/actions.js
```diff
@@ -13,9 +13,18 @@ export const updateApplicationAction = async (application) => {
...
+  const response = await apiService.getAMI({ chartType, chartYear })
   if (response === false) {
     Alerts.error()
   }
   return response['ami']
 }
+
+export const addCommentsWithStatus = async (applicationId, comment, status) => {
+  const data = { status, comment, applicationId }
+  const response = await apiService.createLeaseUpStatus(data)
```
fddayan: @callaghanc I'm using `createLeaseUpStatus` here. The same method you used in the lease up table to change the status with the dropdown. Here is the code: https://github.com/Exygy/sf-dahlia-lap/blob/master/app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js#L81 

It seems to be doing what we expect. Should be renamed `createLeaseUpStatus` to something more generic like `createFieldUpdateComment` ?

callaghanc: Yes, let's rename it to `createFieldUpdateComment`, that more accurately describes what it does.

app/javascript/components/molecules/Dropdown.js
```diff
@@ -4,32 +4,16 @@ import DropdownMenuMultiSelect from '../molecules/DropdownMenuMultiSelect'
...
-
-const computeLeftWith = (ref) => {
-  return 0
-}
-
 class Dropdown extends React.Component {
   constructor (props) {
     super(props)
-    this.state = { expanded: false, style: { top: 0, left: 0 } }
+    this.state = { expanded: false, style: { top: 40, left: 0 } }
```
fddayan, line 11: Setting the values here since the previous code was not being used and we use css clases for manipulating the top.

fddayan:
@callaghanc ready to review

app/assets/stylesheets/application.css.scss
```diff
@@ -51,6 +51,17 @@
   top: 54px !important;
 }
 
+.dropdown-menu-bottom {
```
akegan, line 54: are we sure we want to be updating the css here? are there classes from the pattern library that we could use instead?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -85,9 +87,55 @@ class SupplementalApplicationPage extends React.Component {
     this.setState({ confirmedPreferencesFailed: false })
   }
 
+  openAddStatusCommentModal = () => {
+    this.setState({
+      showAddCommentModal: true,
+      statusModalHeader: 'Add New Comment',
+      statusModalButton: 'save',
+      addCommentStatus: this.props.application.processing_status
```
akegan: what's addCommentStatus

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -85,9 +87,55 @@ class SupplementalApplicationPage extends React.Component {
...
+  openUpdateStatusModal = (value) => {
+    this.setState({
+      showAddCommentModal: true,
+      statusModalHeader: 'Update Status',
+      statusModalButton: 'update',
+      addCommentStatus: value
+    })
+  }
+
+  closeAddCommentModal = () => {
```
akegan: do we need closeUpdateStatusModal? If this is used for both, can we call the state and the function something more generic like

FieldUpdateCommentModal?

app/javascript/components/supplemental_application/actions.js
```diff
@@ -13,9 +13,18 @@ export const updateApplicationAction = async (application) => {
...
 export const getAMIAction = async ({chartType, chartYear}) => {
-  const response = await apiService.getAMI({chartType, chartYear})
+  const response = await apiService.getAMI({ chartType, chartYear })
   if (response === false) {
     Alerts.error()
   }
   return response['ami']
 }
+
+export const addCommentsWithStatus = async (applicationId, comment, status) => {
```
akegan: addComment**s** or addComment?

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
```diff
@@ -1128,70 +1128,33 @@ Array [
...
               >
-                <li
-                  aria-selected="false"
-                  className="dropdown-menu_item"
-                  role="option"
-                >
-                  <a
-                    href="/some/valid/uri"
-                  >
-                    This is a link
```
akegan, line 1169: are we sure we wanted to lose these links? I have a hard time reading the snapshots but I don't want us to lose anything important

akegan (commented):
button color for status dropdown at the bottom should be the color of the status

app/assets/stylesheets/application.css.scss
callaghanc: @akegan For now, we can update the CSS here. We have an [upcoming story](https://www.pivotaltracker.com/story/show/157670856) for migrating the CSS that's been written in LAP into the PL.

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -79,26 +80,29 @@ const ConfirmedPreferencesSection = ({ application, fileBaseUrl, onSave, confirm
...
-      <ul className='dropdown-menu' role='listbox' aria-hidden='true' aria-activedescendant='' tabIndex={0} style={{display: 'none'}}>
-        <li className='dropdown-menu_item' role='option' aria-selected='false'><a href='/some/valid/uri'>This is a link</a></li>
-        <li className='dropdown-menu_item' role='option' aria-selected='false'><a href='/some/valid/uri'>This is another</a></li>
-        <li className='dropdown-menu_item is-selected' role='option' aria-selected='true'><a href='/some/valid/uri'>Yet another</a></li>
-      </ul>
-      <button className='button primary small save-btn' type='submit' disabled={disabled}>Save</button>
-    </div>
-  </div>
-)
+const ButtonPager = withContext(({ loading, store }) => {
```
callaghanc: @fddayan Does the name "ButtonPager" hold meaning for you? It doesn't mean anything to me, so I'm inclined to change it to something like "StatusDropdownWrapper", or something like that.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: yeah that is not super clear let me think about a better name. That is the status in the modal

callaghanc: @akegan @fddayan How about renaming it to `statusModalStatus`?

callaghanc:
@akegan I've been working on this with @fddayan today, we've found some ways to update the approach, and we'll be pushing a couple more commits before this is ready for review again - will ping you when it's ready! I've re-added the "wip" label and put the related Pivotal story in a Started state.

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
fddayan: I see what you are saying. The reason why the links are not showing is because we are using the `Dropdown` component. The previous code was just a "placeholder".  The way the Dropdown component works is that it renders the menu in the DOM only if it's expanded. Since we initially load the page, the dropdown is not expanded. Options are 1) changing the dropdown to render the menu using  css prop `display`, so we always render the menu in the DOM 2) leave it as it is and simulate a click in the dropdown in the test to render the menu. @callaghanc any thoughts?

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
fddayan: I used ButtonPage since that is the name Jesse used for the class: `button-pager`.
I do not think `StatusDropdownWrapper ` makes sense since we have the status dropdown button AND the save button in there. The button page kind of represent the panel footer + panel actions.
I agree it's not super meaningful. Any other suggestion?

fddayan:
@callaghanc ready for review

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
callaghanc: @fddayan I prefer 2), leave it as it is and simulate a click in the dropdown in the test to render the menu.

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
callaghanc: Yes, good point, it's more than the status dropdown. I'll think on a meaningful name and propose something in a bit.

callaghanc: After chatting with @fddayan we've decided to go with `ActionButtons` for now. @fddayan had a good idea that in the future we could work with design to make a reusable buttons component that goes within the `ContentSection` component.

callaghanc (approved):

callaghanc:
@akegan This is ready for review again.

callaghanc:
@fddayan @akegan I added e2e tests for the work done in this PR around supplemental application status updates - please take a look.

spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
```diff
@@ -485,7 +485,7 @@ Array [
                   }
                 >
                   <div
-                    className="dropdown"
+                    className="dropdown "
```
akegan: do we know why the classname has a space in it now?

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
akegan: has this been resolved yet?

spec/javascript/support/puppeteer/steps/sharedSteps.js
```diff
@@ -8,10 +8,10 @@ const loginAsAgent = async (page) => {
   await page.click('.sign-in-btn')
   await page.waitForNavigation()
 
-  // Sale force login
+  // Salesforce login
   await page.waitForSelector('#username_container')
-  await page.type('#username', process.env.E2E_SALESFORCE_USERNAME)
-  await page.type('#password', process.env.E2E_SALESFORCE_PASSWORD)
+  await page.type('#username', 'dahlia-leasing-agent@exygy.com.full')
+  await page.type('#password', 'oBC12FPxGf3I')
```
akegan: whoops. I've reset the password, but let's never ever put a real password into a file that's in source control even if we don't intend to commit it. that's why we have the env vars.

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Enter a comment into the status modal comment field
+  await page.type('#status-comment', COMMENT)
+
+  // Submit the status modal form
+  await page.click('.form-modal_form_wrapper button.primary')
+
+  // The page should enter a loading state while the status modal
+  // submit action is in progress
+  const pageHasLoadingState = await utils.isPresent(page, '.tabs-card-row > .tabs-card > .loading-panel.loading')
```
akegan: is it possible that this could end up being really flakey because the timing might not actually work out that we'll catch the loading state?

akegan (changes requested):
Nice comments in the e2e tests. Username and password need to be switched back to the env vars

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
callaghanc: yeah i thought about that while putting this in, i'm not decided on it. on the one hand, it's nice to be able to test that the loading state appears when we want it to. on the other hand, timing isn't something we can always rely on. @fddayan what do you think?

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
callaghanc: @akegan The e2e test addresses this by using the dropdown in its testing. Do you think we should add additional testing around this?

spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
fddayan: We are doing this `dropdown ${wrapperClasses ? wrapperClasses.join(' ') : ''}`` in here
https://github.com/Exygy/sf-dahlia-lap/pull/111/files#diff-97e2ac7b9ff31ba87ab51ccba57ee7fbR89

We need the extra space in there to render `wrapperClasses` if present. If it not present it will still render the space. It can be fixed using `classNames` lib.  However, it will feel a bit inconsistent because we are using the `join` approach in multiple places. 
@callaghanc any thoughts?

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
fddayan: good point... another option is to use `await page.waitForSelector('[...] .loading')`. This is going to wait until the selector is present and continue with the test. if the selector is not present, the test will time out and fail.

spec/javascript/support/puppeteer/steps/sharedSteps.js
```diff
@@ -8,7 +8,7 @@ const loginAsAgent = async (page) => {
   await page.click('.sign-in-btn')
   await page.waitForNavigation()
 
-  // Sale force login
+  // Salesforce login
```
fddayan, line 11: 😅

spec/javascript/support/puppeteer/steps/sharedSteps.js
fddayan: weird.. how the password ended up here? did we force a push ? I cannot find the commit

fddayan:
@callaghanc E2E looks good.

spec/javascript/support/puppeteer/steps/sharedSteps.js
callaghanc: I removed it soon after it went up - the reason it was there in the first place is that I was having some issues getting the env vars to work locally, so to speed up development i used them directly, intending not to commit them at all! i managed not to commit the other env vars but these slipped by me 🤦‍♀️ 🙄 ...i need a combo face palm/eye roll emoji....

spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
callaghanc: @fddayan We can use `classNames` in this case - would you mind updating to use that?

spec/javascript/support/puppeteer/steps/sharedSteps.js
callaghanc: 🤓

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
callaghanc: we've removed the loading state testing from this test and will test it in a future unit test

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
callaghanc: @akegan check out the recently added unit test that tests the rendering of the status dropdown button and status dropdown menu on the supp app page

akegan (approved):

PR #115, opened 2018-09-26
Chore/ production lap tests #158582269

akegan:
https://www.pivotaltracker.com/story/show/158582269

Solution:

Make listing ID configurable via env var:
- Set to Brannan on Semaphore Full
- Set to yellow acres on Semaphore Prod

It defaults to yellow acres on full, but currently that breaks things, so you can just set the env var E2E_LEASE_UP_LISTING_ID if you want it to be something else locally. 

Here's a passing build on the production semaphore job, running against the prod salesforce instance: https://semaphoreci.com/exygy/sf-dahlia-lap-prod/branches/chore-production-lap-tests-158582269/builds/3

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,7 @@ E2E_SALESFORCE_USERNAME=dahlia-leasing-agent@exygy.com.full
 E2E_SALESFORCE_PASSWORD=<ask the team for this password>
 ```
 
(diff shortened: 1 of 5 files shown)


callaghanc:
@akegan could you recreate this branch using the `chores` naming convention, rather than `chore`?

akegan:
closing to remake with different branch name

PR #118, opened 2018-10-01
Features/update confirm preference status #159935677

fddayan:


app/javascript/components/mappers/domainToSoql/application_preference.js
```diff
@@ -0,0 +1,25 @@
+import { compactShape } from '../utils'
```
akegan: don't forget to remove this

app/controllers/api/v1/applications_controller.rb
```diff
@@ -1,14 +1,40 @@
...
+      end
+
+      private
+
+      def application_params
+        params.require(:application).permit(
+          :id,
+          :total_monthly_rent,
+        )
+      end
```
callaghanc, line 29: Here we only allow the set of params that we actually need in the Salesforce standard REST API application update requests we have so far. So far, we only make one kind of application REST API update, and that's to update just the total monthly rent field.

app/controllers/api/v1/preferences_controller.rb
```diff
@@ -0,0 +1,32 @@
...
+      private
+
+      def preference_params
+        params.require(:preference).permit(
+          :id,
+          :individual_preference,
+          :post_lottery_validation,
+          :type_of_proof,
+        )
+      end
```
callaghanc, line 25: Similar to my comment above, here we only allow the set of params that we need in the Salesforce standard REST API preference update requests that we have so far. So far, we only let users update three fields on preferences from the Confirmed Preference panels.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -81,19 +81,19 @@ class SupplementalApplicationPage extends React.Component {
...
   handleSavePreference = async (preferenceIndex, application) => {
     const { persistedApplication } = this.state
-
-    // We clone the latest saved copy, so we can use the latest saved fields.
     const synchedApplication = cloneDeep(persistedApplication)
 
-    // We use the persisted copy and set only the fields updated in the panel
+    // We need to set the total_monthly_rent in the global application, so we do not overwrite it.
     synchedApplication.total_monthly_rent = application.total_monthly_rent
-    synchedApplication.preferences[preferenceIndex] = application.preferences[preferenceIndex]
```
callaghanc: I know @fddayan is out, so he can't answer this right now, but I'm wondering why we removed this line and what the effect is? @akegan @chrisdolendo do you have further insight?

app/javascript/components/supplemental_application/actions.js
```diff
@@ -18,3 +18,21 @@ export const getAMIAction = async ({chartType, chartYear}) => {
   }
   return response['ami']
 }
+
```
callaghanc, line 21: @fddayan has mentioned that we likely do not really need this actions file going forward. if people would like to remove these functions and place their content directly in the place where they are called in the supplemental application page, i would be fine with that. i'm also fine with leaving this this way until we have a more solid idea of how we want to organize our React requests to Rails.

app/models/force/object_base.rb
```diff
@@ -0,0 +1,42 @@
+module Force
+  # Base representation of a Salesforce object. Provide mapping between
+  # SOQL API and LAP domain field names for preferences.
+  class ObjectBase
+    def initialize(fields, format)
+      raise ArgumentError, 'Field values and format are required.' unless fields && format
+
+      @fields = {
+        soql: {},
```
callaghanc: As I've mentioned to everyone in person, "soql" here is kind of a misnomer. Although we use this set of field names in SOQL queries, the field names themselves have nothing to do with SOQL at all. They are the field names on the Salesforce objects. A task for us as a team going forward is to come up with better terminology to refer to "the set of field names on Salesforce objects"!

callaghanc:
@chrisdolendo @akegan This PR is now ready for review!

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: my guess is that it may have something to do with the fact that we're refreshing the page on application save now? but I don't know why that wouldn't get rid of the whole synched application thing

app/models/force/object_base.rb
akegan: are there separate names for the restforce API and for SOQL?

If not, what if we just call it salesforce? or salesforce_field

so then we could have

domain
salesforce
custom_api?

akegan:
It's hard to tell if this is just because the animation isn't there yet, but if I save the preference status it saves in salesforce, but if I save the value the new status isn't showing up in the table

image

akegan (approved):
Code looks good to me, I'm just not 100% on what the interactivity is meant to be (see comment above)

callaghanc:
@akegan Yeah, that's the current state of the UX, we have a story that Fed has already started for adding in UX to show that update in the table: https://www.pivotaltracker.com/story/show/159877262.

app/models/force/object_base.rb
callaghanc: the salesforce standard REST API that we use through the restforce gem uses the same field names as SOQL does, so yeah there's no difference there in terms of field naming. you know, just "salesforce" as a name isn't bad. it's still a little bit confusing because the custom API is also a way of interacting with salesforce. but it's closer to being accurate than "soql" is. i'd be down to change "soql" to "salesforce" here as a step in the right direction, and i like "custom_api" as a slight improvement over just "custom", and then as we continue to move forward with this we can continue to improve our naming system.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: ok so i tested this and it looks like removing this line introduces a bug if you follow these steps:
1) update a preference in a preference panel (there's no page refresh after updating just a preference)
2) scroll down to bottom of page and click the supp app form Save button
3) the page will refresh, and you will see your changes to the preference undone, because the application was saved with the old preference values

i tried putting this line back in and doing the same test, and with the line in, the above bug behavior doesn't happen, because the supp app form application gets updated with the changes you made to the preference. @fddayan, when you're back online on Monday morning, could you take a look at this and let us know if it's fine for me to put this line back into this PR, or if there was a particular intention behind removing that line?

callaghanc (changes requested):
Approval pending resolution of question about the line in `handleSavePreference` that updates the supp app form application with the preference panel changes.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: I've chatted with @fddayan and he confirmed that he did not intend to remove this line, so I'm going to put it back in.

callaghanc (approved):

fddayan (commented):
looks good 👍 (approved on my end, but I cannot approve it since I'm the author)

PR #120, opened 2018-10-01
Chores/update use of env vars #160877304

callaghanc:
After this PR is approved and merged, the team will:

1. Remove any DAHLIA-related env vars from their local environments.
2. Download the updated .env for LAP from LastPass.

### DIFF
--- a/.git-hooks/pre_commit/scan_for_env_vars.rb
+++ b/.git-hooks/pre_commit/scan_for_env_vars.rb
@@ -1,0 +1,33 @@
+require 'dotenv/parser'
+
+module Overcommit
+  module Hook
+    module PreCommit
+      # Custom Overcommit pre-commit filter to scan for any uses of environment
+      # variable values in committed code and warn about them so they can be
+      # removed. Uses the dotenv gem's parser to read in the vars from .env.
+      class ScanForEnvVars < Base
+        def run
+          env_vars = Dotenv::Parser.call(File.read('.env'))
+          errors = []
+
+          applicable_files.each do |file|
+            env_vars.each do |name, val|
+              File.foreach(file) do |line|
+                if line.include?(val)
+                  errors << "#{file}: File contains the value of the environment variable #{name}. " \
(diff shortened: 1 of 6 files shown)


.git-hooks/pre_commit/scan_for_env_vars.rb
```diff
@@ -0,0 +1,33 @@
+require 'dotenv/parser'
+
+module Overcommit
```
akegan, line 3: do you want help QAing this?

akegan (changes requested):
Can you update the instructions for the e2e tests that say you need to have the env vars actually in your environment?

.git-hooks/pre_commit/scan_for_env_vars.rb
callaghanc: would be great if any/all who are doing PR review can do QA as part of review - so, please actually try out everything in this PR locally and see if it works for you!

callaghanc:
@akegan Updated e2e test section in README.

akegan:
OK
- E2E tests worked for me with unset env vars
- Overcommit caught when I was using a value of an env var. The only headache it might cause is that I have yellow acres set as the default listing ID right now in https://github.com/Exygy/sf-dahlia-lap/blame/master/spec/javascript/support/puppeteer/consts.js#L2. We might run into trouble with those.

image

akegan (approved):

callaghanc:
@akegan For the default listing/application ID in the consts file, it seems that if we have that as a default, a person would not need that same value in their local .env? Is that what you were referring to?

akegan:
yep. I was thinking the same thing too but I just wanted to flag it.

callaghanc:
yep makes sense. in general it will be something we'll have to consider if we in the future need to add an env variable with a common value - for example, if we need to add a flag env var whose value is just true/false. in that case, we'll need to add a way to exclude certain variables from this check. but we can handle that when that case arises.

PR #121, opened 2018-10-02
Features/pl status update component #158885250

callaghanc:
In the course of putting together the StatusUpdate component, I found myself assembling what would have been our fourth "status update dropdown", so I created a component for that to reduce our repetition of this concept. This PR therefore contains a new StatusDropdown component, in addition the StatusUpdate component that is the main focus of this story.

Also, when adding the StatusUpdate component to the pattern library, I did some organizing of the pattern library while I was in there, since it's a been a little while since we tidied that up, so you will also see a bit of that in this PR.


app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -2,26 +2,25 @@ import React from 'react'
...
 
-import Dropdown from '../molecules/Dropdown'
+import StatusDropdown from '../molecules/StatusDropdown'
+import { getLeaseUpStatusClass } from '~/utils/statusUtils'
 import appPaths from '~/utils/appPaths'
 import { cellFormat } from '~/utils/reactTableUtils'
 import classNames from 'classnames'
 
-import { LEASE_UP_STATUS_OPTIONS, PAGE_SIZE, getLeaseUpStatusStyle } from './leaseUpsHelpers'
+const PAGE_SIZE = 5
```
callaghanc, line 11: I would eventually like to move this to a constants file of some kind as part of a bit of refactor of how we are organizing these components. We can address that in another chore.

akegan: I think this is actually already changed in the PR I had to page the applications that has not been implemented yet

callaghanc: Ok cool, sounds like we will be handling it in that story then 👍

app/javascript/packs/pattern_library.js
```diff
@@ -1,81 +1,90 @@
 import WebpackerReact from 'webpacker-react'
```
akegan, line 1: what does this file do? (This is the first time I think I've encountered a PR that touches the pattern library)

akegan (approved):
Looks good to me. Just one question to help me understand better

app/javascript/packs/pattern_library.js
callaghanc: that's a good question, i'd actually like to know more details about this too! i believe Fed set it up initially, we could ask him for more details when he's back. broadly, i know it does setup for the pattern library, but i don't know the details about it.

fddayan: `WebpackerReact` is the library that comes with `rails` to connect `react_component` in the views with our React components under `components`. Basically using `WebpackerReact.setup` we are registering the component, so it can be used by `react_component` in the views. Tecnically we should register only our `*Page.js` files. I created this file `pattern_library.js` that is being imported in [`react_application.js`](https://github.com/Exygy/sf-dahlia-lap/blob/master/app/javascript/packs/react_application.js#L22)  because we need the component in the patter library, but not the the web, since never use `react_component` with a PL component.

callaghanc:
The story for this PR is accepted in Pivotal but Josh has a question about a possible AC clarification, so please do not merge this PR. I will do the merge when I've finished discussing that question with him.

PR #130, opened 2018-10-18
Chores/update rubocop target ruby version #161303475

callaghanc:


akegan (approved):
approved assuming semaphore passes!

PR #132, opened 2018-10-18
Hotfix/fix apps table pagination #161296162

chrisdolendo:


app/javascript/utils/EagerPagination.js
```diff
@@ -42,7 +42,9 @@ class EagerPagination {
       this.server.currentPage = newServerPage
       const result = await fetchPage(this.server.currentPage)
       this.records = result.records
-      this.pages = (result.pages * result.records.length) / this.eager.size
+      let numerator = result.pages * result.records.length
```
akegan, line 45: I'm having trouble understanding what the bug was. Can you explain what was happening and what fixed it? If it's weird enough it might be worth adding a comment in here too

chrisdolendo: Sure, what was happening was that Salesforce was returning 80 applications for the listing in question, but React was only displaying 20 in the table. The issue was in splitting up the returned results into multiple pages, in which the formula (seen here) didn't work for smaller numbers of applications. So we updated it to take that expand that use case. I created a chore to refactor this file to make it more understandable, but this is the hotfix to fix the bug.

akegan: was pages===0 for paper apps on 1101 connecticut?

app/javascript/utils/EagerPagination.js
```diff
@@ -42,7 +42,9 @@ class EagerPagination {
...
+      if (result.pages === 0) { numerator = result.records.length }
```
akegan, line 48:
```suggestion
      // If there is < 1 page of results from salesforce, we get pages == 0. We need to round up to 1 page. 
      // TODO: Refactor this logic to avoid this step. 
      if (result.pages === 0) { numerator = result.records.length }
```

akegan (approved):
Thanks!

chrisdolendo:
This is approved and accepted! Are we setting up a deploy to production for this solo branch or waiting for other hotfixes?

akegan:
@chrisdolendo we are waiting on the heroku issue to be resolved, and this PR is going to go out with the application language bug

PR #136, opened 2018-10-22
Features/paginate leaseup apps #159530254

chrisdolendo:
https://www.pivotaltracker.com/story/show/159530254


akegan:
FYI looks like rubocop is failing

app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -0,0 +1,18 @@
+module Api::V1
```
akegan, line 3: This will fix rubocop error
```suggestion
# frozen_string_literal: true
module Api::V1
```

akegan:
One question that I have on this PR is why the ordering of the preferences seems to be different between this query and what we had before. 

It's just the custom pref 1211 for All preferences, but it's still worth figuring it out.

akegan:
something is wrong with the paging right now. after 20 pages, when we go for the next page on the server, we're getting the same results as for the first server page (back to COP 1 again)

spec/javascript/components/lease_ups/LeaseUpsPage.test.js
```diff
@@ -39,61 +73,25 @@ describe('LeaseUpApplicationsPage', () => {
...
-
     const wrapper = renderer.create(
-      <LeaseUpApplicationsPage listing={listings} applications={applications} />
+      <LeaseUpApplicationsPage listing={listing} />
     )
 
     expect(wrapper.toJSON()).toMatchSnapshot()
   })
 
   test('Should render order by Prefrence_Order and Preference_Lottery_Rank', () => {
```
akegan: typo in Preference order here
```suggestion
   test('Should render order by Preference_Order and Preference_Lottery_Rank', () => {
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -1,38 +1,146 @@
...
+      records: map(response.records, flow(mapApplicationPreference, buildLeaseUpModel)),
+      pages: response.pages
+    }
+  }
+
+  loadPage = async (page) => {
+    const fetcher = p => this.fetchApplications(p)
+    this.setState({ loading: true, page: page })
+    const { records, pages } = await this.eagerPagination.getPage(page, fetcher)
+    this.setState({ applications: records, loading: false, pages: pages })
```
akegan, line 49: looks like the issue is that the page isn't getting updated here

akegan (dismissed):

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
chrisdolendo: See commit 05c4f84 which fixes loading the next set of pages after 20 (this is for Brannan)

chrisdolendo:
@akegan 

Thanks for catching re after 20 pages. Fixed.

Yes was wondering the same about Custom Pref 1211. So the SOQL query for this page requests that it returns the applications by order `'Preference_Order__c' and 'Preference_Lottery_Rank__c'` (aka, order by `Preference_Order__c` and then within that `Preference_Lottery_Rank__c`) -- which is correct, that's the order we want. 

Looking at the data, the  `Custom pref 1211` has`"Preference_Order" => 0.0`, and the next row has `Preference_Order" => 1.0`. So the behavior is correct here given the data that was returned.

Seeing a conflict on EagerPagination.js just now -- fixing now.

chrisdolendo:
Ready for re-review

fddayan (approved):
Looks good 👍

akegan (approved):
This looks good to me now. We won't be able to check all of the different cases of different number of pages until we implement the filtering and sorting, but hopefully it should work!

akegan:
I can't seem to get e2e tests to run. can you test locally @chrisdolendo ?

chrisdolendo:
e2e tests passing is blocked by: https://www.pivotaltracker.com/story/show/161555292

chrisdolendo:
Tests pass, no code updates needed, merging.

PR #139, opened 2018-10-23
Bugs/error when snapshot not found #161333414

callaghanc:
@akegan This implements the alert message that we discussed. This is fine for a fix that we can quickly fit into the Pilot release. Going forward, would be great to integrate this error case into the more structured error handling behavior that we'll be implementing down the line.


akegan:
image
looks good to me!

akegan (approved):

PR #140, opened 2018-10-23
Features/rental assistance part 1 #155196535

fddayan:


callaghanc:
Update on this PR's status: it is now ready for initial code review. This PR still requires tests, which I am in the process of writing, so the WIP label is going to remain for now. However, someone can get a head start on reviewing it in the meantime.

callaghanc:
@akegan @chrisdolendo This PR is now officially ready for review!

app/javascript/components/molecules/ExpandableTable.js
```diff
@@ -13,7 +13,7 @@ class ExpandableTableRow extends React.Component {
   }
 
   render () {
-    const { row, numColumns, expanderRenderer, expandedRowRenderer } = this.props
+    const { row, numColumns, expanderRenderer, expandedRowRenderer, original } = this.props
```
akegan, line 16: what is original?

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -68,30 +71,33 @@ const ConfirmedHousehold = ({ amis, amiCharts, formApi }) => {
...
 
+const StatusHistorySection = withContext(({ store }) => {
+  const { statusHistory, openAddStatusCommentModal } = store
+  return !isEmpty(statusHistory) && (
+    <ContentSection.Sub title='Status History' borderBottom={false}>
+      <StatusList items={statusHistory} onAddComment={openAddStatusCommentModal} />
+    </ContentSection.Sub>
+  )
+})
+
```
akegan, line 100: looks like things got moved around a bunch in this file which makes the diff hard to follow. Is it pretty much just the rental assistance was added or were there other changes?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -43,7 +47,10 @@ class SupplementalApplicationPage extends React.Component {
       statusModal: {
         loading: false,
         status: props.application.processing_status
-      }
+      },
+      addNewRentalAssistance: false,
```
akegan: is this whether or not to show the newRentalAssistance panel? It might be worth making it a bit more explicit and booley so that it's clear that it's true/false and not a function.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -167,6 +178,68 @@ class SupplementalApplicationPage extends React.Component {
...
+      }
+    })
+  }
+
+  handleUpdateRentalAssistance = (rentalAssistance) => {
+    // TODO: Add a real call to Salesforce to update the
+    // rental assistance record. Here, we mock the real
+    // Salesforce update by just returning the rental
+    // assistance values.
+    const updatedRentalAssistance = ((rentalAssistance) => rentalAssistance)(rentalAssistance)
```
akegan, line 213: what's this?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -195,23 +271,33 @@ class SupplementalApplicationPage extends React.Component {
...
+      handleStatusModalSubmit: this.handleStatusModalSubmit,
+      rentalAssistances: rentalAssistances,
+      addNewRentalAssistance: addNewRentalAssistance,
+      showAddRentalAssistanceBtn: showAddRentalAssistanceBtn,
+      hideAddRentalAssistanceBtn: this.hideAddRentalAssistanceBtn,
+      handleOpenRentalAssistancePanel: this.handleOpenRentalAssistancePanel,
+      handleCloseRentalAssistancePanel: this.handleCloseRentalAssistancePanel,
+      handleSaveNewRentalAssistance: this.handleSaveNewRentalAssistance,
+      handleUpdateRentalAssistance: this.handleUpdateRentalAssistance,
+      handleDeleteRentalAssistance: this.handleDeleteRentalAssistance
```
akegan, line 304: this context is getting super long! it looks like it's been organized a bit. what's the plan for ordering them? Do we have to start to worry about it getting out of hand at some point?

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -0,0 +1,262 @@
...
+import FormGrid from '~/components/molecules/FormGrid'
+import ExpandablePanel from '~/components/molecules/ExpandablePanel'
+import YesNoRadioGroup from '../YesNoRadioGroup'
+import formUtils from '~/utils/formUtils'
+import { withField, Field } from '~/utils/form/Field'
+
+const { ExpanderButton } = ExpandableTable
+
+const typeOfAssistance = [
+  'Catholic Charities',
```
akegan, line 18: We should think about storing these in salesforce long term if we find that they change on a regular basis

spec/javascript/components/supplemental_application/sections/RentalAssistance.test.js
```diff
@@ -0,0 +1,78 @@
...
+  test('should not render a table if rental assistances is empty', () => {
+    const context = cloneDeep(baseContext)
+    const component = renderer.create(
+      <Context.Provider value={context}>
+        <RentalAssistance />
+      </Context.Provider>
+    )
+
+    let tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
```
akegan, line 37: how do we check that this works as expected? the snapshot has a lot of lines in it 😕

akegan (changes requested):
Just a few variable name comments and questions!

There's also a bug in switching from "Other" with a name to any other type of assistance
rental_assistance_bug

app/javascript/components/molecules/ExpandableTable.js
callaghanc: the original entire object that that row represents

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
callaghanc: it's just that the rental assistance section was added in the lease information section, between the lease form and the demographics section

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: i agree, how about `showNewRentalAssistancePanel`?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: it's a stub of the update call. this story, part 1 of rental assistance, doesn't actually include doing any real saving or updating of rental assistances in salesforce, but the story still requires us to build out all the structure around that saving and updating, so this story contains stub calls to represent where the real ones will go in part 2

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: good question! perhaps it's time to revisit the larger question of state management. is it maybe time to graduate to a next level of state management? or maybe we can use Context in a more organized way? this would be a good post-stand-up discussion to have soon

app/javascript/components/supplemental_application/sections/RentalAssistance.js
callaghanc: Yes, for now I think it makes sense to have them here, and it would be good to follow up with SMEs about the frequency with which this changes, so we can plan for if/when to store/draw these from Salesforce.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: sounds good!

spec/javascript/components/supplemental_application/sections/RentalAssistance.test.js
callaghanc: i agree the snapshots do contain a fair bit of info, and i think that's sort of the point - if we want to test the exact rendered state of a component, we'll have to check its exact rendered state. when i review snapshots that are newly generated or updated, i open the page containing the component in question, and review that side by side with the snapshot lines, to check that the snapshot lines match what I see on the page, and to check that what I see on the page also matches what is intended in the pivotal story. it forces me to understand very precisely what the correct, intended render is for that component. i'm definitely open to exploring new ways to test the exact render state of components too! or to exploring other kinds of tests that could somehow fulfill or obviate the need to test the rendering of our components. do you have any thoughts around that?

akegan: this seems like a good solution for now, i'll work on being more careful with my review of new snapshots

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -64,9 +66,10 @@ class RentalAssistanceTable extends React.Component {
   buildRows = () => this.props.rentalAssistances.map(ra => {
     const appMember = this.props.applicationMembers.find(m => m.id === ra.recipient)
     const appMemberName = appMember ? `${appMember.first_name} ${appMember.last_name}` : ''
+    const typeOfAssistance = isOther(ra) ? ra.other_assistance_name : ra.type_of_assistance
```
akegan: could we still end up with an issue where we're sending a value to salesforce for "other_assistance_name" when we shouldn't be? This might get resolved in rental assistance pt 2, I just wanted to check

spec/javascript/components/supplemental_application/sections/__snapshots__/RentalAssistance.test.js.snap
```diff
@@ -718,9 +716,7 @@ Array [
             >
               <td
                 className=""
-              >
-                Other
```
akegan: I wonder why this updated 🤔

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -167,6 +178,73 @@ class SupplementalApplicationPage extends React.Component {
...
+      return {
+        rentalAssistances: [...prev.rentalAssistances, savedRentalAssistance],
+        showNewRentalAssistancePanel: false,
+        showAddRentalAssistanceBtn: true
+      }
+    })
+  }
+
+  handleUpdateRentalAssistance = (rentalAssistance) => {
+    // Set other assistance name to null of Other is not selected (Reset to blank).
```
callaghanc: thanks for adding this update @fddayan! the following two lines are pretty clear, so i think we can do away with this comment

spec/javascript/components/supplemental_application/sections/__snapshots__/RentalAssistance.test.js.snap
callaghanc: The snapshot tests weren't very complete at this time - I've updated them a lot so check out the new snapshots and see what you think.

callaghanc:
@akegan This PR is now ready for re-review.

app/javascript/components/supplemental_application/sections/RentalAssistance.js
callaghanc: @fddayan has just added a change to prevent that - now `other_assistance_name` is set to null in the update action if `type_of_assistance` isn't "Other". This does make more sense to test in part 2, however, so although it's added here, the data correctness is something to test in part 2.

akegan (approved):
looks good to me now! E2E tests failed on semaphore, but looks like it was just a timeout so I'm rebuilding now

jrubenoff:
Just accepted this! I just made a slight change to the "Type of Assistance" dropdown to match the capitalization in the InVision mockups, so please wait for the build to pass before merging.

PR #149, opened 2018-10-26
Features/rental assistance part 2 #158503202

fddayan:


config/routes.rb
```diff
@@ -27,24 +29,28 @@
...
       end
-      get 'ami' => 'ami#get'
+
+      scope '/flagged-applications' do
+        put 'update' => 'flagged_applications#update'
+      end
+
       resources :preferences, only: %w[update]
+
+      resources :rental_assistances, path: '/rental-assistances', only: %w[create update destroy]
```
callaghanc, line 51: This is the only important changed line relevant to rental assistances in this file. The rest of the changes are just tidying and alphabetization.

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -137,23 +138,24 @@ const validateError = (values) => {
...
       value: member.id
     }
-  ))
+  })
 
   return (
     <Form onSubmit={onSave} defaultValues={values} validateError={validateError}>
       {formApi => (
-        <div className='app-editable expand-wide scrollable-table-nested'>
+        <div className='app-editable expand-wide scrollable-table-nested rental-assistance-form'>
```
callaghanc: This class and the others added below this in this file are added to enable us to test. The e2e tests need identifiers to be able to locate and interact with the rental assistance form components.

spec/javascript/e2e/ApplicationNewPage.e2e.js
```diff
@@ -34,7 +34,7 @@ describe('ApplicationNewPage', () => {
     expect(values).toContain(LAST_NAME)
     expect(values).toContain(DATE_OF_BIRTH)
 
-    browser.close()
+    await browser.close()
```
callaghanc, line 37: All of these `await`s that you see added in this PR before the `browser.close()`s in our e2e tests are to ensure that we actually wait for the browser to close! Not having these awaits is what was causing the browsers in the e2e tests to never actually close and causing the whole test process to hang for a long time after finishing the tests.

callaghanc:
This PR is code-complete, but we are going to hold off on reviewing this until part 1 is fully QA'ed and merged to master, so this PR will remain WIP for now.

callaghanc:
This PR is ready for review.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -182,62 +183,85 @@ class SupplementalApplicationPage extends React.Component {
...
+        return {
+          rentalAssistances: [...prev.rentalAssistances, rentalAssistance],
+          showNewRentalAssistancePanel: false,
+          showAddRentalAssistanceBtn: true,
+          rentalAssistanceLoading: false
+        }
+      })
+      return true
+    } else {
+      Alerts.error()
```
akegan: does this already display a message or do you have to pass one in?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -182,62 +183,85 @@ class SupplementalApplicationPage extends React.Component {
...
+      this.setState({ rentalAssistanceLoading: false })
+      return false
+    }
```
akegan, line 263: do you think that it'd be possible to condense the create/update/destroy or are there enough slight differences that it's not worth it?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -332,14 +358,15 @@ const setApplicationsDefaults = (application) => {
...
+const mapProperties = ({ application, statusHistory, fileBaseUrl, units, availableUnits, rentalAssistances }) => {
   return {
     application: setApplicationsDefaults(mapApplication(application)),
     statusHistory: mapList(mapFieldUpdateComment, statusHistory),
     onSubmit: (values) => updateApplicationAction(values),
     fileBaseUrl: fileBaseUrl,
     units: mapList(mapUnit, units),
-    availableUnits: mapList(mapUnit, availableUnits)
+    availableUnits: mapList(mapUnit, availableUnits),
+    rentalAssistances: rentalAssistances
```
akegan, line 370: why doesn't rentalAssistances need to use mapList?

app/services/force/soql/rental_assistance_service.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
akegan, line 1: so many new clean and organized services! 🎉

spec/fixtures/model/force/lease/lease_domain.json
```diff
@@ -0,0 +1,12 @@
+{
+  "application_id": "a0o0P00000FFc2yQAD",
+  "id": "a130U0000001n4wQAA",
+  "leaseStartDate": "2028-01-01",
```
akegan: I thought domain was snake_case? I believe other domain fixtures are.

spec/fixtures/model/force/lease/lease_salesforce_from_domain.json
```diff
@@ -0,0 +1,12 @@
+{
```
akegan, line 1: does this file name need the "from_domain" or is it just lease_salesforce?

spec/fixtures/model/force/rental_assistance/rental_assistance_salesforce_from_domain.json
```diff
@@ -0,0 +1,8 @@
+{
```
akegan, line 1: same here re: filename

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
```diff
@@ -39,7 +39,140 @@ describe('SupplementalApplicationPage', () => {
...
+
+    // Wait for the new rental assistance form to appear
+    await page.waitForSelector('.rental-assistance-new-form')
+
+    // Set up the values we'll use to fill out the rental assistance form
+    const [typeVal, typeLabel] = await page.$eval(
+      '.rental-assistance-new-form .rental-assistance-type option:nth-child(2)',
+      e => [e.value, e.textContent]
+    )
+    const recurring = 'Yes'
```
akegan: nit: this section might be bit clearer if the page.$evals are grouped before or after the recurring/amount.

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
```diff
@@ -39,7 +39,140 @@ describe('SupplementalApplicationPage', () => {
...
+
+    // Wait for the API rental assistance create call to complete
+    await page.waitForResponse(request => request.url().includes('http://localhost:3000/api/v1/rental-assistances'))
+
+    // Check that the rental assistances table has increased in size by one
+    const newTableSize = await page.$$eval('.rental-assistances > tbody > .tr-expand', elems => elems.length)
+    expect(newTableSize).toEqual(prevTableSize + 1)
+
+    // Check that the last rental assistance in table has the values we saved
+    const lastRentalAssistanceSelector = '.rental-assistances > tbody > .tr-expand:nth-last-child(2)'
```
akegan: why is it nth-last-child(2) if it's the last row?

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
```diff
@@ -21,7 +21,7 @@ describe('SupplementalApplicationPage', () => {
```
akegan: at some point soon we may want to think about breaking this up into multiple files so it doesn't get too long. For now, do you think it might help to add another layer of "describe" to help organize the e2e tests a bit?

akegan (commented):
I think these are all only comments/questions. Looks great! I really appreciate all the cleanup on the backend and the e2e tests!

also hooray for adding the arrow down! 🎉

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: this triggers our standard window alert, which has the message "Oops! Looks like something went wrong. Please try again."

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: there are enough differences in the post-request steps that it's probably not worth squishing them into a single method; however, perhaps the loading state management and error alert could be DRYed a bit - @fddayan what are your thoughts?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: because the rental assistances already come in in domain format! 🎉 thanks to our refactor work, the SOQL rental assistance service uses SOQL to fetch the rental assistances, and then translates them to domain before passing them up to the controller. that way the controller can just hand the frontend domain-formatted rental assistances. the ultimate goal here is to do away with all use of `mapList` in the frontend. and then, even further then that, once that's done and the frontend is entirely in domain, we can consider whether we'd like to convert domain format from snake case to camel case - but that's a little further off yet.

app/services/force/soql/rental_assistance_service.rb
callaghanc: 😍

spec/fixtures/model/force/lease/lease_domain.json
callaghanc: yeah, for some reason all the already-existing lease domain fields were in camel case when i came to this story. i think you and @chrisdolendo did the previous lease work - does anyone recall why this is? i'm fine with changing them to snake case in this PR if there's no problem with doing that. even though we may ultimately decide that we want to convert domain to camel case down the line, until then it would be better to maintain consistency

spec/fixtures/model/force/lease/lease_salesforce_from_domain.json
callaghanc: we specify both the origin format and the result format, because the content of a translated object will depend on where that object originated from. objects can have fields that are only present when they are fetched from salesforce, or fields that are only present when they are fetched from the custom api. if i fetch, say, an application from the custom api, and then translate it to domain, it will only have the domain correlates of the custom api fields. if i fetch that same application via SOQL and translate it to domain, it will have the domain field correlates for those salesforce fields, which are a slightly different set of fields from what is returned by the custom api. this is a lot more relevant for application and preference objects, but i wanted to maintain consistency when testing all objects.

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
callaghanc: yeah i've also been thinking that - i wonder if there's any problem with doing that breaking up now? @fddayan @chrisdolendo how would you guys feel about breaking up the supp app e2e file now?

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
callaghanc: i wanted to keep them in the order that they're filled in on the page - especially for debugging, it makes it a lot easier to see "ok, we wanted to use this value for this field, then this value for the next....and the test got to this field, then the next..."

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
callaghanc: because it's not actually the last table row, it's only the last visible row - there's a hidden row right after it that is its corresponding expandable panel

spec/fixtures/model/force/lease/lease_domain.json
akegan: yeah probably my fault . If you're up for it I think we should make it consistent. If you'd prefer, I can followup in a new PR after this one

akegan (approved):
looks good. 

Only outstanding things are possibly
1. Updating lease to have snake case domain model
1. Splitting up or organizing supplemental app tests

But it all looks good to me!

spec/fixtures/model/force/lease/lease_domain.json
callaghanc: i'll do it in this PR, i got most of it done already during retro and sprint planning 😀

callaghanc: done!

callaghanc:
I've updated this PR with standardized lease domain field names. This PR is awaiting some responses from @fddayan and/or @chrisdolendo on some questions.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
fddayan: I would keep a method for each action. I agree we would DRY up those methods.

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
fddayan: 👍I'm ok with splitting the file if it's growing. We could re think the naming convention we use for e2e tests. Instead of using the Page as the main test we can can `SupplementalApplicationPage/feature_name.e2e.test` 
like 
`SupplementalApplicationPage/manage_rental_asistances.e2e.test` .... or any name

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: @akegan Fed has added a commit showing a way to draw out the common parts of the rental assistance action functions - please review and let us know if you'd like to use that new way or keep it as it was before.

spec/javascript/e2e/SupplementalApplicationPage.e2e.js
callaghanc: @akegan Fed and I have broken apart the supp app e2e tests in a couple of recent commits - please review.

callaghanc:
@akegan Fed and I have added a few commits to this PR. In addition to addressing the two items you listed as outstanding in your last comment, these commits also do a bit of lease mapping work and correct a field mapping issue with preferences. So please review the commits since you last reviewed and let us know your thoughts.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: I'm ambivalent! Either way works. Up to you guys!

spec/javascript/e2e/SupplementalApplicationPage/preferences.e2e.js
```diff
@@ -0,0 +1,61 @@
+import puppeteer from 'puppeteer'
```
akegan, line 1: small nit: filenames aren't super consistent for these new broken-out things.

action_buttons
preferences
manage_rental_assistance
status_history

I think it might be a bit clearer if we pick a convention and stick with it. 

I'd prefer if we keep them to domain-related topics, e.g. 
-preferences
-rental_assistance
-status_update (combine status history and action_buttons)

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -26,9 +27,9 @@ const typeOfAssistance = [
   'Q Foundation: Person with Disability Rent Subsidy',
   'Q Foundation: Senior Rent Subsidy',
   'San Francisco AIDS Foundation',
-  'Section 8 HCV (Tenant) Voucher',
+  'Section 8 HCV (tenant) voucher',
   'Self-Help for the Elderly',
-  'VASH Voucher',
+  'VASH voucher',
```
callaghanc, line 32: @jrubenoff I've changed the capitalization on these two options back to their original Salesforce picklist capitalizations, so that they will exactly match the picklist values.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
callaghanc: Ok let's leave it as it is now

spec/javascript/e2e/SupplementalApplicationPage/preferences.e2e.js
callaghanc: sounds good to me, done!

app/javascript/components/supplemental_application/sections/RentalAssistance.js
jrubenoff: That's unfortunate, but OK!

akegan:
@callaghanc I've merged latest master into this 🤞the tests will pass and we can merge this in

PR #150, opened 2018-10-30
Bugs/paper address save issue #161599608

akegan:
https://www.pivotaltracker.com/story/show/161599608

It turns out that both the Primary applicant mailing address, and the alternate contact mailing address had the same issue


app/services/force/custom_api/application_service.rb
```diff
@@ -48,6 +48,7 @@ def application_snapshot(id)
       end
 
       def submit(data)
+        puts 'WHAT is being posted', application_defaults.merge(data)
```
chrisdolendo: Remove puts

chrisdolendo (approved):
Just remove the puts :)

PR #152, opened 2018-11-02
Features/Rename the app to DAHLIA Partners #161687493

jrubenoff:
We decided on the new name in SME yesterday.

This PR:

- Uses the new name on the homepage, page title, and README
- Changes "Leasing Agent Sign In" to "Partner Sign In"
- Has a new SVG logo for resolution independence

It looks like we had the wrong app name in the navbar ("DAHLIA" instead of "Leasing Agent Portal"), so I updated that as well.


jrubenoff:
@slowbot Mostly looking for a confirmation that I edited the right CSS file...

jrubenoff:
The SVG logo looks fine in my local environment, but doesn't display in the Heroku deploy.

Anyone have any idea why?

callaghanc:
@jrubenoff I tried deploying this branch to a one-off Heroku app and the logo showed up. That makes me believe the issue on this PR's review app may have been a Heroku deployment blip. As we've seen, Heroku has been having some issues with their review app pipeline lately, so this could be related to all that. I also added a small change to this branch to shorten the syntax used to get the logo path in SCSS. I deployed that to the one-off app and the logo continued to appear correctly. I then deployed this change to this PR's review app and the logo appeared on the review app after deploy. While it might look like my small syntax change fixed the issue, I believe it's probably more likely that the issue was fixed by simply re-deploying. I am going to move onto other stories, but you are welcome to test re-deploying without my recent change to check this assumption.

jrubenoff:
Cool! @slowbot, please review when you can.

slowbot (approved):
Looks good to me!

PR #153, opened 2018-11-02
bugs/paginate-application-table-#161465925

fddayan:
https://www.pivotaltracker.com/story/show/161465925


app/javascript/utils/EagerPagination.js
```diff
@@ -3,6 +3,8 @@ import { slice } from 'lodash'
 // Note: This needs to match the page size defined on the server in soql_query_builder.rb.
 const SERVER_PAGE_SIZE = 100
 
+const MAX_SERVER_LIMIT = 2000
```
akegan: It's technically 2100 (max offset is 2000, but we can get the last page so we can show 2100 results)

app/javascript/utils/EagerPagination.js
```diff
@@ -26,6 +28,10 @@ class EagerPagination {
     this.server.currentPage = -1
   }
 
+  isOverLimit (page) {
+    return (page * this.eager.size) >= (MAX_SERVER_LIMIT + SERVER_PAGE_SIZE)
```
akegan: oh, I see here. I think it's easier to just have it be >= MAX_SERVER_LIMIT and set MAX_SERVER_LIMIT to 2100, then it'll be correct when you display the message too

app/javascript/components/applications/ApplicationsTable.js
```diff
@@ -17,7 +18,9 @@ const SubComponent = (row) => (
 /*      Presenter                         */
 /******************************************/
 
-const ApplicationsTable = ({ applications, onFetchData, loading, pages, rowsPerPage }) => {
+const ApplicationsTable = ({ applications, onFetchData, loading, pages, rowsPerPage, maxPages }) => {
+  const maxPagesMsg = `You have reached the maximum number of records we can display at this time. There are ${pages * rowsPerPage} records that match your selected filters, but the maximum we can display is ${MAX_SERVER_LIMIT}. Please use filters to narrow the number of matching records.`
```
akegan: Would it be easy to format the numbers with commas?

akegan: also, is pages here the rounded up number, or the number of pages with the decimal? If it's the rounded number, would it be possible to have the real number here? There could be a partial page at the end and we don't want to display incorrect information.

akegan (dismissed):
I forgot to put the other comments together into a review, but I wanted to follow the we will of requesting changes if we have any sort of request.

fddayan:
@akegan ready for review again

akegan:
looking at this now before I sign off. Sorry for delay!

akegan:
OK. I just did some more QA and we still have the bug for lease up table paging. It's not technically in this bug, but if it's an easy fix it'd be great to add it in! let me know if that's possible @fddayan. I might be able to fix it tomorrow morning before you're online too if that's easier

akegan:
I've updated the code to use the maxPage logic for the lease up applications table too.

Remaining issues/thoughts:

- [ ] The total number of applications displayed in the message is currently incorrect (it rounds up to the nearest whole page. Suggested options:
  - Change wording to "...approximately {rounded up number of applications} that match your filter..."
  - Change wording to "...There are {number of pages} pages of records that match your filter, but the maximum we can display is 420/105..."
  - alternatively, with more code, we could pass the true value into the table, but if we could get away with not doing that then that would be idea. 

@jrubenoff any thoughts on the above options? 

- [ ] I think that the maxPages variable name could be updated to be more boolean-like. I thought that it was a number when I was first reading through the code. Maybe something like `atMaxPages`? I am leaving that stylistic choice up to you @fddayan since this is your PR.

jrubenoff:
For the error message, let's just keep it super simple:

screen shot 2018-11-07 at 10 33 36 am

I think the the exact number of search results is important enough that you should see it on _every_ page, not just the 106th. So let's move that to a separate story. How difficult would it be to implement?

akegan:
@jrubenoff I made a ticket for displaying the total number of records: https://www.pivotaltracker.com/story/show/161811785 please add some more details about where you think it should go when you get a chance.

fddayan:
@akegan message updated, ready for review again

akegan (approved):
@fddayan I updated the message for lease ups table too, and added in max_server_pages to dynamically calculate the message.

I've approved your pieces, so when you think it's ready to go feel free to deliver to product (don't forget to add review instructions

fddayan:
Approved on my end. Delivering story.

akegan:
Fixed some merge conflicts on this one because of the filtering. It should be good to go but we should retest before merge

akegan:
OK tested and it still works! merging now!

PR #154, opened 2018-11-05
Chores/Upgrade loofah gem #161711831

akegan:
https://www.pivotaltracker.com/story/show/161711831

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -113,7 +113,7 @@ GEM
       rb-fsevent (~> 0.9, >= 0.9.4)
       rb-inotify (~> 0.9, >= 0.9.7)
       ruby_dep (~> 1.2)
-    loofah (2.2.2)
+    loofah (2.2.3)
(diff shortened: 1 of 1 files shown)


Gemfile.lock
```diff
@@ -373,4 +373,4 @@ RUBY VERSION
    ruby 2.3.7p456
 
 BUNDLED WITH
-   1.16.6
+   1.17.1
```
fddayan, line 376: This PR is for upgrading `loofah` and it's upgrading the bundle version in the lock file. It's fine with me I just want to check if you notice this.

akegan: yeah, my bundler version was < bundler version in the lockfile, but when I upgraded it became > bundler version in the lockfile so I updated it.

I don't think it should have any consequences, but I can remove if you think that it might

fddayan (approved):
bundler version is not a problem. Looks good 👍

PR #156, opened 2018-11-07
Features/lease up export #154064904

akegan:
https://www.pivotaltracker.com/story/show/154064904


config/routes.rb
```diff
@@ -29,6 +29,8 @@
 
   resources :pattern_library, only: %w[index]
 
+  get 'reports/:id' => redirect { |params| "#{ENV['SALESFORCE_INSTANCE_URL']}/#{params[:id]}?csv=1" }
```
akegan: @chrisdolendo @callaghanc I am trying to do the redirect here but I don't know if we have access to whether or not a user is an admin here and this needs to be ENV['COMMUNITY_LOGIN_URL'] if they're not an admin

callaghanc:
@akegan I changed the setting of the `SALESFORCE_BASE_URL` global to use a more standard format for interpolating Ruby into an embedded Javascript Slim section. That now works in both dev and prod. Now I'm interested in why the syntax you had before worked in development at all! I'm going to move onto other things I have to do today, but you may want to look into that a bit for your own edification, and I'll return to this tmrw to follow up as well.

akegan:
Thanks! that's what I was going to try next 😄

spec/vcr/listings/applications_controller/index.yml
```diff
@@ -21,7 +21,7 @@ http_interactions:
       message: OK
     headers:
       Date:
-      - Mon, 22 Oct 2018 19:10:14 GMT
```
akegan: VCR cassettes had to be updated to include the Report__id

callaghanc:
@akegan I tried this out on the review app and I wasn't able to download the report while logged into the review app as an admin user. When I clicked on the Export link, a new tab opened with a Salesforce login page, asking me to log in. Could you look into this? It's likely something on the Salesforce side.

Somewhat relatedly, perhaps we should think of adding testing around whether users of different permission levels can access the reports, since that's really been the sticky question all along with this feature.

callaghanc (changes requested):
I'm putting a changes requested review on this to indicate that a review was done and the result was there are some outstanding questions around this functionality. Don't actually have any specific code change requests.

akegan:
I was able to reproduce your issue @callaghanc. It looks like I had been separately logged into test.salesforce.com in my browser when I was testing. It looks like this is an issue with salesforce sessions transferring over. I'll check with vertiba to see if this can be easily resolved.

If not, we might have to keep the login step for MOHCD admins if they need to download a report this way

akegan:
@callaghanc see chat on the ticket. We are not going to implement a fix for the bug you found (although it's a good one!) pls review/accept ASAP so we can get it over to product testing

callaghanc (approved):

akegan:
did another quick spot check on this and it's still good to go! merging now

PR #157, opened 2018-11-07
Features/add filters to lease up apps table #160574827

chrisdolendo:


app/controllers/listings/lease_ups/applications_controller.rb
```diff
@@ -22,7 +21,7 @@ def listing_service
     end
 
     def compacted_listing(full_listing)
-      full_listing.slice('Id', 'Name', 'Building_Street_Address')
+      full_listing.slice('Id', 'Name', 'Building_Street_Address', 'Listing_Lottery_Preferences')
```
akegan, line 24: I just added a field to this too. By the end of next month it'll be all the listing fields 😆

app/javascript/components/lease_ups/LeaseUpApplicationsFilter.js
```diff
@@ -0,0 +1,46 @@
+import React from 'react'
+import { Form, Text, Select } from 'react-form'
+import formUtils from '~/utils/formUtils'
+import Loading from '~/components/molecules/Loading'
+
+const StatusOptions = formUtils.toOptions(['No Status', 'Processing', 'Disqualified', 'Approved', 'Lease Signed', 'Waitlisted', 'Withdrawn', 'Appealed', [null, 'All Status']])
```
akegan: These are defined in app/javascript/utils/statusUtils.js (aside from 'No Status' and 'All Status')

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -102,7 +120,8 @@ class LeaseUpApplicationsPage extends React.Component {
 
   render () {
     const listing = this.props.listing
-
+    console.log(listing)
```
akegan: probably just here because it's still WIP, but wanted to give you a heads up just in case

app/services/force/soql/preference_service.rb
```diff
@@ -7,15 +7,44 @@ class PreferenceService < Force::Base
...
-          SELECT #{query_fields(:show_preference)}
+          SELECT #{query_fields(:app_preferences_for_application)}
           FROM Application_Preference__c
           WHERE Application__c = '#{application_id}'
         ), :show_preference)
 
         result.map { |r| Force::Preference.from_salesforce(r).to_domain }
       end
+
+      def application_preferences_for_listing(opts)
```
akegan: this file has some variable names with application abbreviated, some with preference abbreviated, and some with both.

I think we should settle on one or the other and stay as consistent as possible

app/services/force/soql/preference_service.rb
```diff
@@ -7,15 +7,44 @@ class PreferenceService < Force::Base
       FIELD_NAME = :preferences
       FIELDS = load_fields(FIELD_NAME).freeze
 
-      def application_preferences(application_id)
+      def application_preferences_for_application(application_id)
```
akegan: what is an `application_preference_for_application`?
I'm also not super clear on what an `application_preference_for_listing` is

config/salesforce/fields/listings.yml
```diff
@@ -86,3 +86,13 @@ listings:
     Credit_Rating__c:
     Lottery_Status__c:
     Office_Hours__c:
+  lease_up_listing_fields:
```
akegan, line 89: do you still need this if you're using the compacted listing?

akegan (commented):
I know this is still WIP so some of the things you maybe were already planning on updating, I just wanted to get a head start on checking it out and had some questions.

app/controllers/listings/lease_ups/applications_controller.rb
chrisdolendo: hehe

config/salesforce/fields/listings.yml
chrisdolendo: This was not related to the feature that utilizes compacted listing, it was a refactor from getting rid of the `lease_up_service.rb` and moving the `lease_up_listings` (which is the data needed for the lease up listings page) method to the `listing_service.rb`. Looks like it requests a few different fields so it was given its namespace.

app/services/force/soql/preference_service.rb
chrisdolendo: Preferences are a bit complex in the domain model but from what I gathered, the `Application_Preference` object allows us to see preferences and most importantly their order and rank -- this is the object needed for the lease ups apps table since we need the order and rank. `application_preference_for_listing` was essentially `lease_up_service#lease_up_listing_applications` but the method name was not accurate because we weren't getting applications, we were getting `application_preferences` that were matched to the listing and then constructing the data so that we also get application fields from its related application. 

So to make it more RESTful we got rid of `lease_up_service` since it was not an actual resource, and moved `#lease_up_listing_applications` to `preference_service` and renamed it as `application_preference_for_listing`. And then renamed `application_preferences` to `application_preferences_for_application` (because of the `WHERE` clause`).

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
chrisdolendo: removed!

app/javascript/components/lease_ups/LeaseUpApplicationsFilter.js
chrisdolendo: I was looking for that! Thanks :)

app/services/force/soql/preference_service.rb
chrisdolendo: made it consistent :+1:

app/javascript/utils/statusUtils.js
```diff
@@ -1,20 +1,30 @@
...
   {value: 'Appealed', label: 'Appealed', style: 'is-appealed'},
-  {value: null, label: 'All Status'}
+  {value: 'Waitlisted', label: 'Waitlisted', style: 'is-waitlisted'},
+  {value: 'Disqualified', label: 'Disqualified', style: 'is-disqualified'},
+  {value: 'Approved', label: 'Approved', style: 'is-approved'},
+  {value: 'Lease Signed', label: 'Lease Signed', style: 'is-leased'}
+]
+
+export const LEASE_UP_STATUS_FILTER_OPTIONS = [
+  'No Status',
```
akegan: why the update to duplicate here? Seems like it'd be hard to track updates even if they're in the same file

akegan:
@chrisdolendo looks like e2e tests are failing on this PR too

akegan:
Looks like E2E tests are just failing because of the same bug that I fixed in https://github.com/Exygy/sf-dahlia-lap/pull/160. 

Other than that, I just have one last question on this PR @chrisdolendo

app/javascript/utils/statusUtils.js
chrisdolendo: @akegan it was overwriting the status update dropdown to consolidate the two, but you're right it risks having to track in more than one place. I took a look again and just updated the code so it uses on constant and just appends and prepends the other two values when generating the filter -- see latest commit at f63e7a37d22bbd83d0777c3846c7b6338f1f7830.

chrisdolendo:
@akegan ready for your re-review!

akegan (approved):
Looks good to me! we have the outstanding question about filtering null lottery numbers/removed applications but that can most likely go into another story.

PR #158, opened 2018-11-08
Bugs/commas in number fields #161196317

fddayan:
This PR is very WIP, these are the items to look at

1. Application new save is not working anymore. It was working with this changes.
1. If you save annual income like $1,000.0 and you refresh the page. You will see 1000.00 in annual income.
1. Improve validation message.


app/javascript/components/applications/application_form/PaperApplicationForm.js
```diff
@@ -59,9 +59,22 @@ const buildPrefValidations = (prefs) => {
   return prefValidations
 }
 
+const validateAnnualIncome = (value) => {
+  if (isNil(value)) {
+    return null
+  } else {
+    if (/[^0-9$,.]/.test(value)) {
+      return 'only accepts numbers, $, commas, and periods.'
```
callaghanc: this error message wording is provisional. i have a question out to design for input on the wording here

callaghanc:
Since @fddayan is ramping down from DAHLIA, I'll take over from him on this PR. He is still available for questions until the 26th though!

Remaining items:
- [x] Look into application save issue mentioned by Fed above (I believe he may having been seeing the expected error message when viewing a newly added application on a Lease Up listing - that throws an error because the new app doesn't have a snapshot. No error occurs when adding and viewing a new app on a non-Lease Up listing).
- [x] Get input from design on Annual Income field error message wording.
- [x] Get decision from design and product on whether to add complete currency parsing in this story.
- [x] Add unit test for the error state of the Annual Income field in the paper application form.

app/javascript/components/applications/application_form/PaperApplicationForm.js
callaghanc: resolved with new wording provided by josh

spec/javascript/components/applications/application_form/__snapshots__/PaperApplicationForm.test.js.snap
```diff
@@ -0,0 +1,7290 @@
...
+                        "id": "a0o0x000000OcOzAAK",
+                        "listing": Object {
+                          "id": "a0W0x000000GhJUEA0",
+                          "name": "Test 5/30",
+                        },
+                        "lottery_number": "00191270",
+                        "lottery_number_manual": null,
+                        "monthly_income": null,
+                        "name": "APP-00191270",
+                        "preferences": Arra
```
callaghanc: this error class on the form group is the beginning of the error state we expect to see when the annual income field is in an invalid state. please examine the following elements to note several more instances where error classes are applied. also notice the presence of the small error message below the field. there will be another comment below this marking the end of the section that is relevant to the error state of the annual income field.

callaghanc: this ends the section of the snapshot that deals with the error state of the annual income field

akegan:
@callaghanc looks like this PR has a couple of small merge conflicts. after they're resolved let me know and I can review/approve!

akegan:
@callaghanc I went ahead and resolved the merge conflicts since you have such limited hours this week.

spec/javascript/components/applications/application_form/PaperApplicationForm.test.js
```diff
@@ -0,0 +1,41 @@
...
+          listing={listing}
+          application={applicationWithInvalidAnnualIncome}
+        />
+      )
+
+      wrapper.find('form').first().simulate('submit')
+
+      await wait(100)
+
+      expect(wrapper).toMatchSnapshot()
```
akegan: @callaghanc and @chrisdolendo Unrelated to this PR, but a question I have in general: Would it be possible to do the same sort of checks that we do in e2e tests for the presence of e.g. a validation error in these snapshot tests?

For example, in the ApplicationNewPage "should fail if required fields are missing" test, we have this expectation. 
```
const errors = await page.$$eval('.form-group.error span.error', divs => divs.map(d => d.textContent))
expect(errors).toContain('Please enter a First Name')
```

we could add something like that in addition to checking that it matches the snapshot, to be a little more explicit about what we're expecting.

callaghanc: already chatted about this in slack, but check out the new test updates i added that do something very similar to this

callaghanc:
@akegan Tests passed on Semaphore! I believe this should be ready for a final PR review from you.

akegan (approved):
New tests strategy is great. I'll try to incorporate it into my future PRs

PR #160, opened 2018-11-14
Chores/supplemental app e2e failing  #161933825 

akegan:
https://www.pivotaltracker.com/story/show/161933825


akegan:
It appears that the issue was just flakiness on the page refresh after preferences were updated. I don't have a way to tell that this absolutely fixes the problem, but I've run the e2e tests 5 times and they've succeeded each time so I think this should fix our issues.

chrisdolendo (approved):
@akegan BRILLIANT! Merging.

PR #163, opened 2018-11-26
 Bugs/error reading first name applications #161459497

chrisdolendo:
https://www.pivotaltracker.com/story/show/161459497


spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
```diff
@@ -7131,9 +7241,15 @@ exports[`SupplementalApplicationPage should render the status update dropdown bu
                                                     "is_lottery_complete": false,
                                                     "lease": undefined,
                                                     "listing": Object {
+                                                      "account": Object {},
```
akegan, line 7485: hm. this looks like these fields were not being included at all previously, but now they're being included as empty objects. Could that potentially have other ramifications?

chrisdolendo: @akegan really great question -- I looked and yes it has some side effects e.g. it renders blank fields on forms if there is no value for adding preferences on the add application page. The universal implementation allows us to utilize a "Null Object Pattern" so that there are no error on page renders if the nested object value doesn't exist, but there can be some side effects. I am meeting with Cat tomorrow morning to discuss further. Will give you an update from what comes out of that.

akegan:
@callaghanc @chrisdolendo were you going to add some more work to this? Should I label it as WIP?

chrisdolendo:
still in progress, just added the label!

chrisdolendo:
@akegan @callaghanc this is ready for review. I implemented Null Object Pattern throughout the whole site so that rendering nested values that are undefined will not break the app, it also required me to go to templates and explicitly check if collections are `!isEmpty` before rendering its section, so that markup doesn't get rendered if they're not meant to.

I'll be Q/A-ing this and would appreciate a second eye as well before delivering to PO.

P.S. For some reason removing the "WIP" label on the PR isn't working for me. Will try again later.

akegan:
looks OK to me at first glance. what was your strategy for making sure you put "isEmpty" everywhere that needed it?

chrisdolendo:
@akegan Cool I looked at every template in the whole codebase where the check is made and made sure snapshots didn't correctly change as a result.

akegan:
@chrisdolendo it looks like the the branch isn't automatically deploying with new commits. do you know why that might be?

app/javascript/components/applications/application_details/ApplicationDetails.js
```diff
@@ -68,7 +68,7 @@ const ApplicationDetails = ({ application, fields, fileBaseUrl }) => {
       <ApplicationCard
         title='Declared Household Income'
         fields={declareHousholdIncome} />
-      { application.flagged_applications && (
+      { !isEmpty(application.flagged_applications) && (
```
akegan, line 71: This application doesn't have any flagged applications, but the section still shows up. Is that what you're trying to prevent here?
https://dahlia-lap-full-pr-163.herokuapp.com/applications/a0o0P00000ItC5QQAV

image

akegan:
Can we add a test to make us feel a bit more confident about the changes?

akegan (changes requested):
throwing on a request changes since I have an open question on this PR

app/javascript/components/applications/application_details/ApplicationDetails.js
chrisdolendo: I tried to check that application on full but looks like it was deleted? But to test I did pull this application up:`/applications/a0o0U000000XDKzQAO`

It doesn't have flagged applications and the table doesn't show up as expected.

akegan: ok interesting. good to go then!

akegan (dismissed):

akegan:
@chrisdolendo heads up there are merge conflicts on this. Were you also going to add tests?

chrisdolendo:
Merge conflicts fixed, test added and updated, ready for your re-review!

akegan:
looks like snapshot tests are failing on semaphore

spec/javascript/components/applications/ApplicationPage.test.js
```diff
@@ -1,22 +1,45 @@
...
+import { clone } from 'lodash'
 import renderer from 'react-test-renderer'
-
 import ApplicationPage from 'components/applications/ApplicationPage'
 import sharedHooks from '../../support/sharedHooks'
 import domainApplication from '../../fixtures/domain_application'
 
-describe('ApplicationPage', () => {
-  sharedHooks.useFakeTimers()
+describe('ApplicationPage for app without flagged apps', () => {
```
akegan: looks like the 2nd test here ends up being:
"ApplicationPage for app without flagged apps should render successfully with flagged apps table". I'm not sure that makes sense.

Should the "without flagged apps" be removed from this top-level describe?

akegan (commented):
I hate to keep this PR open for way too long, but I wonder if we can add a test for the initial bug. 

Can we have a fixture with several applications for a listing, and check that the table loads successfully even if some of the required info is missing for one of them?

It'd also be great to be able to confirm that the way we render the applications that are missing required info in tables is what we want (e.g. do they show up or not show up, if they do show up, what is displayed for first/last name)

chrisdolendo:
@akegan good idea, I'll write a test for that with some missing info on the fixture.

akegan (dismissed):
Popping a request changes on here to to make it clear that there are changes that you're working on

akegan:
i'm going to add the test

akegan:
I'm seeing an error where if I try to search for a malformed app:
e.g.
APP-00270393 on
http://localhost:3000/listings/a0W0P00000FCwdmUAD/applications

I get 
```
invariant.js:42 Uncaught Error: IndexTableCell(...): Nothing was returned from render. This usually means a return statement is missing. Or, to render nothing, return null.
    at invariant (invariant.js:42)
    at reconcileChildFibers (react-dom.development.js:12693)
    at reconcileChildrenAtExpirationTime (react-dom.development.js:13016)
    at reconcileChildren (react-dom.development.js:13007)
    at mountIndeterminateComponent (react-dom.development.js:13442)
    at beginWork (react-dom.development.js:13821)
    at performUnitOfWork (react-dom.development.js:15864)
    at workLoop (react-dom.development.js:15903)
    at HTMLUnknownElement.callCallback (react-dom.development.js:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:138)
invariant @ invariant.js:42
reconcileChildFibers @ react-dom.development.js:12693
reconcileChildrenAtExpirationTime @ react-dom.development.js:13016
reconcileChildren @ react-dom.development.js:13007
mountIndeterminateComponent @ react-dom.development.js:13442
beginWork @ react-dom.development.js:13821
performUnitOfWork @ react-dom.development.js:15864
workLoop @ react-dom.development.js:15903
callCallback @ react-dom.development.js:100
invokeGuardedCallbackDev @ react-dom.development.js:138
invokeGuardedCallback @ react-dom.development.js:187
replayUnitOfWork @ react-dom.development.js:15311
renderRoot @ react-dom.development.js:15963
performWorkOnRoot @ react-dom.development.js:16561
performWork @ react-dom.development.js:16483
flushInteractiveUpdates$1 @ react-dom.development.js:16728
batchedUpdates @ react-dom.development.js:2143
dispatchEvent @ react-dom.development.js:4556
interactiveUpdates$1 @ react-dom.development.js:16715
interactiveUpdates @ react-dom.development.js:2150
dispatchInteractiveEvent @ react-dom.development.js:4533
2react-dom.development.js:14227 The above error occurred in the <IndexTableCell> component:
    in IndexTableCell (created by ReactTable)
    in div (created by TdComponent)
    in TdComponent (created by ReactTable)
    in div (created by TrComponent)
    in TrComponent (created by ReactTable)
    in div (created by TrGroupComponent)
    in TrGroupComponent (created by ReactTable)
    in div (created by Tbody)
    in Tbody (created by ReactTable)
    in div (created by TableComponent)
    in TableComponent (created by ReactTable)
    in div (created by ReactTable)
    in ReactTable (created by IndexTable)
    in IndexTable (created by ListingApplicationsTable)
    in ListingApplicationsTable (created by ListingApplicationsPage)
    in section (created by TabCard)
    in TabCard (created by TabsSection)
    in TabsSection (created by TableLayout)
    in TableLayout (created by ListingApplicationsPage)
    in ListingApplicationsPage (created by MapProps)
    in MapProps

Consider adding an error boundary to your tree to customize error handling behavior.
Visit https://fb.me/react-error-boundaries to learn more about error boundaries.
```

akegan:
@callaghanc update on this. instead of passing empty strings for specific values, I tackled the error at it's source: IndexTableCell. It turns out returning undefined instead of null was causing issues with react. This way we cover the case of any of the values being undefined instead of just first and last name

spec/javascript/components/IndexTable.test.js
```diff
@@ -22,6 +28,15 @@ describe('IndexIndexTable', () => {
     expect(tree).toMatchSnapshot()
   })
 
+  test('Should render Empty string if values are undefined', () => {
```
akegan: I added this test to check if errors are raised if undefined is passed into the table cell. I was able to confirm that if i remove my check if val is undefined then the tests fail

spec/javascript/components/__snapshots__/IndexTable.test.js.snap
```diff
@@ -1,5 +1,1669 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`IndexIndexTable Should render Empty string if values are undefined 1`] = `
```
akegan: @callaghanc we've been talking a bit about how to do snapshots better. Here's a good candidate: this snapshot is ~1600 lines long and very hard to read. Maybe we can pair on finding a better solution?

akegan:
@callaghanc I think this is in a state for an initial review. 

Still TBD: is there a better way to test this that doesn't involve a crazy long snapshot.

akegan:
I just pushed updated tests and snapshots to explicitly check that flagged applications table is rendered/not rendered based on whether or not there are any flagged applications present. I tried using enzyme selectors for the first time so check it out

app/javascript/components/applications/application_details/ApplicationDetails.js
callaghanc: @akegan On the application page, did we want to prevent entire sections from showing up when there is no content available for them, or are we choosing to still show the section header when there is no content for the section?

spec/javascript/components/applications/ApplicationPage.test.js
```diff
@@ -1,22 +1,56 @@
...
+            'id': 'a0r0P00002X4r08QAB',
+            'rule_name': 'Name + DOB',
+            'total_number_of_pending_review': 51
+          }
+      }]
 
-  test('should render succesfully', () => {
-    const fileBaseUrl = 'http:/www.someurl.com'
+      // expect(applicationWithFlagged.flagged_applications).toHaveLength(1)
+      expect(applicationWithFlagged.flagged_applications[0].flagged_record).toBeTruthy()
```
callaghanc: @akegan seems like we could get rid of these two lines, what do you think?

spec/javascript/components/applications/ApplicationPage.test.js
```diff
@@ -1,22 +1,56 @@
...
+          file_base_url={fileBaseUrl} />
+      )
+      // Flagged application content card should render
+      expect(wrapper.exists(flaggedAppCardSelector)).toEqual(true)
+      // Check that the row is there and contains the right rule name.
+      expect(wrapper.find(`${flaggedAppCardSelector} > table > tbody > tr`)).toHaveLength(1)
+      expect(wrapper.find(`${flaggedAppCardSelector} > table > tbody > tr > td`).first().text()).toEqual('Name + DOB')
 
-    expect(wrapper.toJSON()).toMatchSnapshot()
+      expect(wrapper).toMatchSnapshot()
```
callaghanc: @akegan what are your thoughts on this approach of using both enzyme selectors and a full snapshot check? could we fully replace the snapshot check with enzyme selectors that check the items that are relevant to the test?

spec/javascript/components/__snapshots__/IndexTable.test.js.snap
callaghanc: @akegan let's see if we can use the `enzyme-to-json` package that we already have installed, and also if we can think through what needs to be in this snapshot - can we find time today to pair?

callaghanc (changes requested):
@akegan broadly looks good! let's pair today to look at tests, and also to see if we can sort out why this branch isn't being auto-deployed to the review app

app/javascript/components/applications/application_details/ApplicationDetails.js
callaghanc: @jrubenoff do you have an opinion here on whether we should hide the section entirely when no data is available for it, or if we should still show the section header? it seems like as an eng team we're leaning towards hiding the section entirely, but we want to validate this with design

jrubenoff: Ideally we should show a blank state informing the user that there's no content for that section, but hiding it entirely is fine for now

akegan: We definitely agree that that would be better. I thought we had a ticket for improving our display of blank fields in the application view page somewhere but I'm having trouble finding it. @jrubenoff do you remember if you already made one? If not, I can create a new one

jrubenoff: I don't think there's a story for it

app/javascript/components/listings/ListingApplicationsPage.js
```diff
@@ -31,9 +31,7 @@ const tableFields = {
     'label': 'Total Household Size'
   },
   'application_submission_type': {
-    'label': 'Application Submission Type',
-    'editable': true,
```
akegan, line 35: IndexTable on it's own doesn't have an editing option for fields. We also don't want users to be able to change the application submission type in this table. 

See [SpreadsheetIndexTable](https://github.com/Exygy/sf-dahlia-lap/blob/master/app/javascript/components/SpreadsheetIndexTable.js#L37-L42) for an example of IndexTableCell edit-ability in action

spec/javascript/components/IndexTableCell.test.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react'
+import renderer from 'react-test-renderer'
+import IndexTableCell from 'components/IndexTableCell'
+
+// TODO: Add tests for editable cell functionality.
```
akegan, line 5: I ended up limiting my tests here pretty severely. Let me know if you think it's worth it to add a few more

callaghanc:
@akegan is this ready for re-review?

akegan:
@callaghanc yep!

app/javascript/components/applications/application_form/preferences/PreferencesSection.js
```diff
@@ -1,17 +1,17 @@
...
-  if (formApi.values && formApi.values.preferences && listingPreferences) {
+  if (formApi.values && !isEmpty(formApi.values.preferences) && listingPreferences) {
     return formApi.values.preferences.length === listingPreferences.length
   }
 }
 
 const hasHouseholdMembers = (formApi) => {
-  let hasPrimaryApplicant = formApi.values.applicant && formApi.values.applicant.first_name
-  let hasHouseholdMembers = formApi.values.household_members && formApi.values.household_members.length
+  let hasPrimaryApplicant = !isEmpty(formApi.values) && formApi.values.applicant.first_name
```
callaghanc: @akegan is there a reason we don't check for the presence of `formApi.values.applicant` here?

app/javascript/components/applications/application_form/preferences/utils.js
```diff
@@ -10,7 +10,7 @@ export const naturalKeyFromMember = (member) => {
 }
 
 export const memberNameFromPref = (p) => {
-  if (p && p.application_member) {
+  if (!isEmpty(p) && p.application_member) {
```
callaghanc: @akegan is there a reason we don't do a `!isEmpty` check on `p.application_member` here?

app/javascript/components/applications/application_form/preferences/PreferencesSection.js
akegan: no I'm not sure. This was a chris update. I'm not 100% sure where this is being used.

callaghanc: ok, i'm going to add a `!isEmpty(formApi.values.applicant)` condition here to be safe

akegan:
@callaghanc latest commit looks good to me! good to see link show up properly in snapshot and have so much snapshot disappear!

callaghanc (approved):

akegan:
Just waiting on QA from @callaghanc and @emilyshore on this one.

callaghanc:
@akegan I recall that last we spoke we agreed that you and @emilyshore would do QA on this one, but let me know if I'm recalling that incorrectly

PR #165, opened 2018-11-29
Features/remove lease ups from gen listings #157445311

chrisdolendo:


akegan (approved):
Looks good! nice quick fix 😄

akegan:
@chrisdolendo this has been product accepted but has merge conflicts 😢

app/services/force/listing_service.rb
```diff
@@ -10,7 +10,7 @@ class ListingService < Force::Base
     def listings
       parsed_index_query(%(
         SELECT #{query_fields(:index)} FROM Listing__c
-        WHERE #{user_can_access}
+        WHERE Status__c != 'Lease Up'
```
chrisdolendo, line 13: @akegan just checking to make sure - master doesn't have the WHERE #{user_can_access} line anymore to allow more viewing permsissions correct? This PR included which is why there was a merge conflict. I updated this logic to just say WHERE Status__c != 'Lease Up' and then I'll merge to master, but wanted to make sure first.

chrisdolendo: @akegan just checking on this before I merge it!

akegan: yep we're good to remove it! the function should be gone too 🤔

akegan: looks the the function is also gone so good to go!

PR #166, opened 2018-12-05
Bugs/rental assistance delete error #162430991

callaghanc:
The rental assistance delete functionality is only possible via a custom Salesforce API endpoint that allows rental assistance deletion. This Salesforce work is not yet deployed to Prod, so we are disabling the rental assistance delete function by hiding the button. When the Salesforce work is on Prod, we will re-enable rental assistance deletion. This PR also contains the disabling or removal of any tests or parts of tests that involved rental assistance delete functionality.


akegan (approved):
Confirmed that there is no rental assistance delete button showing up 
image

PR #168, opened 2018-12-06
Bugs/lease up apps listing filter fix #162431638

chrisdolendo:


chrisdolendo:
Context: the lease ups applications filter used `listing_id` for it's listing filter param, while the applications table filter used `listing` for it's listing filter param. When `General` is selected on the filter of the lease ups apps table, it calls the same service function that the applications table uses -  `ApplicationService#applications`, which adds a soql filter for `listing`. So I standardized the filter param name so it is both using the same param which is now `listing_id` since that is most accurate. That fixed the bug.

akegan (approved):
I was able to confirm on full that I can load General apps for 855 brannan on this branch, but not on master.

PR #169, opened 2018-12-06
Features/update priority types #161762862

akegan:
https://www.pivotaltracker.com/story/show/161762862

### DIFF
--- a/spec/javascript/components/applications/__snapshots__/ApplicationPage.test.js.snap
+++ b/spec/javascript/components/applications/__snapshots__/ApplicationPage.test.js.snap
@@ -557,7 +557,7 @@ Array [
                     Has ADA Priorities Selected
                   </h4>
                   <p>
-                    Vision impaired;Mobility impaired;Hearing impaired
+                    Vision impairments;Mobility impairments;Hearing impairments
                   </p>
                 </div>
                 <div
--- a/app/javascript/components/applications/application_form/formOptions.js
+++ b/app/javascript/components/applications/application_form/formOptions.js
@@ -140,9 +140,9 @@ const preferenceProofOptionsWorkInSf = labelize([
 ])
 
 const priorityOptions = [
-  'Mobility impaired',
-  'Vision impaired',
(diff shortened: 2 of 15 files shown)


spec/javascript/components/applications/__snapshots__/ApplicationPage.test.js.snap
```diff
@@ -557,7 +557,7 @@ Array [
                     Has ADA Priorities Selected
                   </h4>
                   <p>
-                    Vision impaired;Mobility impaired;Hearing impaired
+                    Vision impairments;Mobility impairments;Hearing impairments
```
callaghanc, line 560: @akegan This is fine in that it is the direct value from Salesforce and it matches the formatting that already existed here, but perhaps we can add spaces after the colons to make it look a little nicer and be easier to read?

akegan: I think there's a ticket in the backlog already to make all the formatted stuff look nicer in the application edit page

callaghanc: is this one you're thinking of? https://www.pivotaltracker.com/story/show/162027506

akegan: this one https://www.pivotaltracker.com/n/projects/1405352/stories/162028377.

callaghanc: ok, well looks like both of those cover different aspects of app page formatting, so seems like we've already planned to do this at a later time 👍

callaghanc (approved):

spec/javascript/components/applications/__snapshots__/ApplicationPage.test.js.snap
akegan: oh whoops. didn't see ADA priorities in the other one.

PR #172, opened 2018-12-11
Features/remove browser alert on confirmed pref error #160637476

chrisdolendo:


chrisdolendo:
Task before merge: Remove commit 3fe18b7. It was added only for PO testing to replicate an API error. Test fail on purpose as a result.

akegan (approved):
Glad it ended up being so easy! I'll go ahead and deliver it

chrisdolendo:
Removed commit 3fe18b7. It was added only for PO testing to replicate an API error. Tests should pass now, and will merge.

PR #174, opened 2018-12-12
Bugs/paper application field save #162591755

akegan:
This PR fixes two minor bugs found in testing paper application saving

Filipino/Tagalog not saving: https://www.pivotaltracker.com/story/show/162591755

Agency name not saving: https://www.pivotaltracker.com/story/show/162591170

### DIFF
--- a/app/javascript/components/applications/application_form/AlternateContactSection.js
+++ b/app/javascript/components/applications/application_form/AlternateContactSection.js
@@ -49,7 +49,7 @@ const AlternateContactSection = ({editValues}) => {
               </div>
               <div className='small-4 columns'>
                 <label>Agency (if applicable)</label>
-                <Text field='agency' />
+                <Text field='agency_name' />
               </div>
             </div>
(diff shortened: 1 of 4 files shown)


akegan:
@callaghanc @chrisdolendo this is a fix for 2 small bugs that Josh discovered during regression testing. We should prioritize having an E2E test for the paper application that involves filling out all of the fields, but in this case we're in a bit of a time crunch to get this deploy out in time to do the timed release on monday night.

callaghanc (approved):
approved. noting that although our policy is to include testing in all PRs, exceptions can be made for cases like this which are very time-sensitive. in these cases, i'd like to acknowledge the testing we're skipping by creating a specific chore for completing that testing at a later date. @akegan would you be able to create a chore specifically for testing the fields of the paper app, if we don't already have one?

akegan:
here's the chore: https://www.pivotaltracker.com/story/show/162620968

PR #177, opened 2018-12-13
Features/update annual income field #160776676

chrisdolendo:
https://www.pivotaltracker.com/story/show/160776676


spec/controllers/applications_controller_spec.rb
```diff
@@ -24,7 +23,7 @@
     context 'with a non lease up application' do
       let(:expected_non_lease_up_app) { fixture('controllers/applications/non_lease_up_application_domain.json') }
 
-      it 'should return a domain application' do
+      it 'should return a domain application snapshot' do
```
akegan: Hm, I think with a non-lease up application is should *not* return a snapshot, right?

akegan (dismissed):
just one question. otherwise looks good

akegan:
@chrisdolendo the ticket also needs review instructions

spec/controllers/applications_controller_spec.rb
chrisdolendo: Good catch, updated.

chrisdolendo:
@akegan hola! tests pass and ready for your re-review

akegan:
@chrisdolendo can you add more details to the review instructions around how this value should compare to the value in the snapshot view, and maybe a check on salesforce to make sure that it's saving to the correct field?

akegan (changes requested):
Moving this comment to request changes just so it's clear that this PR is waiting on something:

@chrisdolendo can you add more details to the review instructions around how this value should compare to the value in the snapshot view, and maybe a check on salesforce to make sure that it's saving to the correct field?

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -44,13 +44,12 @@ const ConfirmedHouseholdIncome = ({ amis, amiCharts, formApi }) => {
           </FormGrid.Group>
         </FormGrid.Item>
         <FormGrid.Item>
-          <FormGrid.Group label='Household Annual Income'>
-            <Text field='annual_income'
+          <FormGrid.Group label='Confirmed Household Annual Income'>
```
akegan: From the AC it looks like this should be updated to Confirmed Annual income so the field title doesn't wrap

app/javascript/components/mappers/domainToApi/application.js
```diff
@@ -24,6 +24,7 @@ export const applicationFieldMapper = {
   has_developmental_disability: 'hasDevelopmentalDisability',
   answered_community_screening: 'answeredCommunityScreening',
   annual_income: 'annualIncome',
+  confirmed_annual_income: 'confirmedHouseholdAnnualIncome',
```
akegan: I'm also going to update this to be confirmed_household_annual_income. It's long but income fields are confusing enough that I think we need to be very consistent.

app/javascript/components/mappers/soqlToDomain/application.js
```diff
@@ -29,6 +29,7 @@ export const mapApplication = (a) => {
     application_submission_type: a.Application_Submission_Type,
     application_submitted_date: a.Application_Submitted_Date,
     annual_income: a.Annual_Income,
+    confirmed_annual_income: a.Confirmed_Household_Annual_Income,
```
akegan: update here too

akegan (dismissed):

akegan:
I confirmed that the field is tied to the correct field on salesforce, and that the field label and helper text matches the AC:
image

@callaghanc I added some changes here so if you'd be up for a quick review on this that would be great

callaghanc (approved):

akegan:
thank you for approving! did you also run through all of the review steps? would you be up for clicking "accept" on the ticket?

callaghanc:
i did also do the Pivotal story review steps, so i'll accept the story in Pivotal

spec/fixtures/controllers/supplementals/index.json
```diff
@@ -0,0 +1,394 @@
+{
```
akegan, line 1: @chrisdolendo how did you generate this file? We need to update it but I'm not sure how you made it in the first place

PR #181, opened 2019-01-18
Features/Show 20 leaseup applications #163337965

akegan:
https://www.pivotaltracker.com/story/show/163337965

### DIFF
--- a/spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
+++ b/spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
@@ -1279,6 +1279,2301 @@ Array [
                 </div>
               </div>
             </div>
+            <div
+              className="rt-tr-group"
+              role="rowgroup"
+            >
+              <div
+                className="rt-tr -padRow -even rt-tr-status"
+                role="row"
+                style={Object {}}
+              >
+                <div
+                  className="rt-td td-min-narrow"
+                  onClick={[Function]}
+                  role="gridcell"
+                  style={
+                    Object {
+                      "flex": "100 0 auto",
+                      "maxWidth": null,
+                      "width": "100px",
+                    }
+                  }
+                >
+                  <span>
+                     
+                  </span>
+                </div>
+                <div
+                  className="rt-td text-left"
+                  role="gridcell"
+                  style={
+                    Object {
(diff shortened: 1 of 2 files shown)


akegan:
@callaghanc this was a really minor request that came out of testing with Mel over at MOHCD. Finishing up listing service refactor is definitely a higher priority, but if you have a couple of minutes between tasks it'd be great to get this in.

spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
```diff
@@ -1279,610 +1279,7634 @@ Array [
                 </div>
               </div>
             </div>
-          </div>
```
callaghanc, line 1282: @akegan are you able to speak to all the changes in this snapshot?

akegan: Sure! I haven't gone over them line-by-line but I think they are all mostly empty rows being added since now we're loading a table with 20 rows instead of 5 rows.

callaghanc: yes looks like that sort of thing mostly. i'm wondering why the "previous" button is removed here? we can also look at this when i stop by the DS office later

spec/javascript/components/lease_ups/__snapshots__/LeaseUpsPage.test.js.snap
```diff
@@ -1279,610 +1279,7634 @@ Array [
...
+                  }
+                >
+                  <span>
+                     
+                  </span>
+                </div>
+                <div
+                  className="rt-td text-left"
+                  onClick={[Function]}
+                  role="
```
akegan, line 3595: @callaghanc it moved down here. Just bad github diffing

callaghanc: ok cool

callaghanc (approved):
did my best to review as much of the large snapshot as possible. as discussed with @akegan we are working as we go forward on reducing snapshot sizes. next time we do any substantial work on the lease up page, that will be an opportunity to reduce the size of the snapshot in this PR.

PR #182, opened 2019-02-13
Features/supp app currency validation #162897074

akegan:
https://www.pivotaltracker.com/story/show/162897074


app/javascript/components/applications/ApplicationNewPage.js
```diff
@@ -5,12 +5,10 @@ import CardLayout from '../layouts/CardLayout'
 import mapProps from '~/utils/mapProps'
 import { mapListing } from '~/components/mappers/soqlToDomain/listing'
 import { saveApplication } from './actions'
-import { parseHouseholdIncome } from './application_form/utils'
 
 const ApplicationNewForm = ({ listing }) => {
   const saveNewApplication = async (submitType, submittedValues, application, listing, editPage) => {
     submittedValues.listing = { id: listing.id }
-    submittedValues.annual_income = parseHouseholdIncome(submittedValues.annual_income)
```
akegan, line 13: I moved this translation into the domainToApi mapping to keep it consolidated

app/javascript/components/supplemental_application/sections/LeaseInformationInputs.js
```diff
@@ -37,46 +38,72 @@ const toggleNoPreferenceUsed = (formApi, value) => {
...
+                </FormGrid.Item>
+                <FormGrid.Item>
+                  <Field.Text
+                    label='Monthly Parking Rent'
+                    field='monthly_parking_rent'
+                    placeholder='Enter Amount'
+                    errorMessage={(label, error) => error} />
+                </FormGrid.Item>
+                <CustomFormGrid.Plain
+                  label='Total Monthly Rent'
```
akegan, line 98: this has issues with the currency fields, but will be removed in https://www.pivotaltracker.com/story/show/163567761 so I'm not going to fix it here

akegan (commented):
WIP: Still need to add/update tests

akegan:
@callaghanc I still need to add e2e tests to check the conversion of string to float before we send to salesforce but I was wondering if you could take a preliminary look at the changes I've made here

app/javascript/components/mappers/domainToApi/application.js
```diff
@@ -23,20 +23,20 @@ export const applicationFieldMapper = {
...
   housing_voucher_or_subsidy: 'householdVouchersSubsidies',
   terms_acknowledged: 'agreeToTerms',
   number_of_dependents: 'numberOfDependents',
   number_of_seniors: 'numberOfSeniors',
   number_of_minors: 'numberOfMinors',
   reserved_senior: 'hasSenior',
   application_submission_type: 'applicationSubmissionType',
-  total_monthly_rent: 'totalMonthlyRent',
   processing_status: 'processingStatus',
+  householdAssets: (source) => currencyToFloat(source.household_assets),
```
akegan, line 34: @callaghanc mostly interested in your thoughts on this change.

app/javascript/components/mappers/domainToApi/lease.js
```diff
@@ -0,0 +1,14 @@
+import { createFieldMapper } from '~/utils/objectUtils'
+import { currencyToFloat } from '~/utils/utils'
+
+// Lease is sent to backend with domain keys.
+export const leaseFieldMapper = {
```
akegan, line 5: and this one

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
```diff
@@ -594,17 +594,24 @@ Array [
                 <div
                   className="form-group"
                 >
-                  <label
-                    htmlFor=""
+                  <div
```
akegan, line 597: this is a larger number of changes than I expected, but skimming through none of the changes look out of place, and the actual visual interaction looks correct

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
```diff
@@ -29,6 +29,44 @@ const testStatusModalUpdate = async (page) => {
...
+  const input = await page.$(selector)
+  const valueHandle = await input.getProperty('value')
+  return valueHandle.jsonValue()
+}
+
+const savePage = async (page) => {
+  const selector = '#save-supplemental-application'
+  // To allow waitForNavigation to work post-save, we need to scroll
+  //   down to the save button, then wait for the url to update to the scrollable anchor
+  // FIXME this will cause issues if we use savePage when we're already at the
```
akegan, line 59: The issue here:
If the page scrolls down to the save button, the url updates to add #status-history-section. Without this step the timing was such that my page.waitForNavigation() after save was just catching this url update rather than the actual page reload. 

I tried importing goToAnchor from 'react-scrollable-anchor' and it didn't appear to do anything.
I also tried doing page.goto('#status-history-section') and that didn't work either. 

Let me know if you have any other ideas of how to avoid the waitForNavigation here.

spec/javascript/e2e/SupplementalApplicationPage/confirmed_household.e2e.js
```diff
@@ -0,0 +1,46 @@
+import puppeteer from 'puppeteer'
```
akegan, line 1: I broke up lease and confirmed HH income tests to help with clearness of testing, but it does lead to more overhead with login and navigation to the page. thoughts on whether we should try to combine them or keep them separated?

akegan:
I have done all of the testing that I was planning to do so this is officially ready for review!

app/javascript/components/mappers/domainToApi/lease.js
callaghanc: @akegan with this and the above, it is nice to see how we can use what we already have to accomplish the goal of this story. i also wonder if it may be worth adding a bit more new work to achieve greater clarity, by doing value conversion/formatting separately from field mapping. what are your thoughts on this?

akegan: do you have an idea of where it might make more sense to do the conversion/formatting?

spec/javascript/e2e/SupplementalApplicationPage/confirmed_household.e2e.js
callaghanc: would be nice to not have that additional overhead. perhaps we can situate both within the same test, but break the steps for checking lease fields and the code for checking HH income into separate files, to keep some separation and clarity?

akegan: I think the issue with that would be that we still wouldn't get a helpful error message if one of the tests failed

callaghanc: could we use unit testing here? would that get at what we're trying to test in this scenario?

akegan: unfortunately I don't think so because we don't have any other E2E integration testing for the supplemental application yet and the stripping of the currency values happens in the mapping step

callaghanc (dismissed):
@akegan after a quick look i didn't immediately see a way around the scroll-to-anchor issue, so i'm approving this PR and will move onto QA'ing the Pivotal story

akegan:
ok thanks for looking into it

callaghanc (changes requested):
Just need to add validation on the currency fields as described in AC on https://www.pivotaltracker.com/story/show/162897074.

app/javascript/components/mappers/domainToApi/lease.js
callaghanc: update on this conversation: @akegan and I discussed on a call and decided to wait for the upcoming upgrade to a new form library to break out these kinds of formatting.

akegan:
I've added the number limits for currency and lease fields, but realized that I had forgotten about rental assistance.

akegan:
Last piece left is to add tests for rental assistance and isUnderMaxVal

akegan:
@callaghanc this is ready for review again. I'll work on updating the review instructions now.

spec/javascript/e2e/SupplementalApplicationPage/confirmed_household.e2e.js
callaghanc: update on this conversation: @akegan and I discussed on a call and decided to let these tests be but also to cap the number of separate e2e test areas for the supp app to the current 5

callaghanc:
@akegan posting my comment from stand-up here: i'm ready to approve this PR and i've also QA'd the pivotal story and it looks good to go, if you can get the tests passing on this PR then we can merge today

akegan:
@callaghanc found the issue, it was a rounding issue that only came up intermittently

We were sending 47.50 and getting back "47.5" and then checking against "47.50".

akegan:
Should pass now and be ready for re-review

callaghanc (approved):

PR #184, opened 2019-02-19
Features/update app member from confirmed prefs #160307353

chrisdolendo:
https://www.pivotaltracker.com/story/show/160307353


akegan:
@chrisdolendo just started with some quick QA and it doesn't look like it's working quite right
If you look at [Portia toole's application on the review app](https://dahlia-lap-full-pr-184.herokuapp.com/applications/a0o0P00000GZazsQAD/supplementals)

The app member is showing up as "undefined undefined" for most of the preferences.

If you go into salesforce, those preferences do have application members defined https://sfhousing--full.cs26.my.salesforce.com/a0o0P00000GZazsQAD

akegan (dismissed):
@chrisdolendo just started with some quick QA and it doesn't look like it's working quite right
If you look at [Portia toole's application on the review app](https://dahlia-lap-full-pr-184.herokuapp.com/applications/a0o0P00000GZazsQAD/supplementals)

The app member is showing up as "undefined undefined" for most of the preferences.

If you go into salesforce, those preferences do have application members defined https://sfhousing--full.cs26.my.salesforce.com/a0o0P00000GZazsQAD

chrisdolendo:
@akegan sorry didn't push my latest work -- see latest build. also writing review instructions now

akegan:
@chrisdolendo is there a reason why app member can't be changed on rent burdened?
Nevermind, makes sense now. Sounds like assisted housing can be updated, but not rent burdened? 🤷‍♀️ 
From ticket:
> And in those panels that do offer a field for selecting the app member, the field is labeled “Name on Lease” in the Assisted Housing preference panel, and is labeled “HH Member on Proof” in all the other panels that have app member dropdown fields. 

Does look like assisted housing should be updated to say  "name on lease" instead of "HH Member on Proof" though

image

akegan (dismissed):
Looks much better! One small issue is that app member update doesn't save on entire page save if confirmed pref panel isn't saved first.

We have the functionality to save the pref proof value on page save, looks like we just need to add app member to that

chrisdolendo:
@akegan good catch! looking into it now

akegan (approved):
Looks good with changes!

I was able to update and save HH member change on all prefs with the edit dropdown both by individually hitting the save button and by changing them all and hitting save on the whole application

PR #188, opened 2019-03-07
Features/validate alt contact #162623064

chrisdolendo:


akegan:
@chrisdolendo did you find a solution on this?

app/javascript/components/applications/application_form/AlternateContactSection.js
```diff
@@ -1,23 +1,39 @@
...
+import validate from '~/utils/form/validations'
+import { Field } from '~/utils/form/Field'
 
 const {
   alternateContactOptions,
   phoneTypeOptions
 } = formOptions
 
+const validateError = (formValues) => {
+  return {
```
akegan, line 23: the way we do it for householdMember validations, is to check if it's empty before doing the validations. 
https://github.com/Exygy/sf-dahlia-lap/blob/master/app/javascript/components/applications/application_form/PaperApplicationForm.js#L63-L74. I think if you did a similar thing here that would get rid of the need for `validateFirstLastName`

akegan (dismissed):
just one comment/request. Otherwise code looks good. I'll QA now

app/javascript/components/applications/application_form/AlternateContactSection.js
chrisdolendo: Yes, I initially did this but the formApi wouldn’t get updated when I tried to clear the empty string. I can try again and get a 2nd pair of eyes on it

chrisdolendo: @akegan I implemented this way and looks like it works -- modifying formValues directly than accessing the formApi :+1:

chrisdolendo:
@akegan ready for re-review!

app/javascript/components/applications/application_form/AlternateContactSection.js
akegan: hm, for this I meant that you could replace validateFirstLastName with just validate.isPresent. Let me know if you want to hop on a quick call about it

akegan (dismissed):
I put this in slack too but want to reiterate here, that right now if I add an alternate contact to a paper app, save, then go to edit it, I can't successfully remove the alternate contact

I get these validation errors
image

akegan (approved):
Looks good! did a bunch of editing and adding and taking away alt contacts and it looks good

PR #193, opened 2019-03-11
Bugs/New lease creation error #164555093

akegan:
https://www.pivotaltracker.com/story/show/164555093

[#164555093] Update short form controller to check for lease presence
SOQL lease service returns  {} if lease cannot be found.


app/controllers/api/v1/short_form_controller.rb
```diff
@@ -199,7 +199,7 @@ def submit_lease
       # from the supp app page, that will have created a blank lease before the
       # entire supp app form was saved.
       existing_lease = soql_lease_service.application_lease(application_api_params[:id])
-      if existing_lease
+      if existing_lease.present?
```
akegan, line 202: soql_lease_service returns {} if the lease doesn't exist.

akegan:
Seeking thoughts on testing for this. My thought is that options are:

#### 1. Full E2E test for adding lease information if lease doesn't exist on supp app:
- This would probably require a separate testing application (so as not to interfere with rental assistance tests)
- We'd have to start the test by deleting the lease associated with the supp app before then creating a new one. This isn't a functionality we currently have in the backend but we could add it

#### 2. Add rake spec tests
- We could mock the response from soql_lease_service in different scenarios with different payloads.
- If the issue with this bug is that the response from soql_lease_service changed, then that test wouldn't necessarily fix it. 

Any other ideas?

akegan:
This is ready for review.

The tests are fairly minimal, but will be added to when we implement this follow-up ticket to give the lease object its own controller https://www.pivotaltracker.com/story/show/164589943

app/controllers/api/v1/short_form_controller.rb
chrisdolendo: easy enough!

chrisdolendo (approved):
Looks good to me, nice quick fix.

PR #194, opened 2019-03-11
Bugs/new lease creation error #164555093

akegan:
just looking for that test coverage reporting


akegan:
don't need this PR anymore because we're switching code climate back

PR #196, opened 2019-03-13
Move lease logic to separate controller #164589943

akegan:
https://www.pivotaltracker.com/story/show/164589943

This PR started off as a refactor to make lease updates easier to test, but ended up including some improvements and a minor bug fix. Moving lease updates out of the short form API led to some refactoring of how we handle errors, and a (hopefully) a speed improvement where we now send lease and application update requests at the same time. 

*Front end*
- [x] No preference used field now being passed to API
- [x] Lease updates made to be parallel with app updates

*Backend*
- [x] Lease updates separated out from short form API

*Still TODO*
- [x] Tests for leases controller
- [x] Tests for API service and updateApplication action
~- [ ] (Potentially) tests for lease service~


app/javascript/apiService.js
```diff
@@ -109,6 +110,23 @@ const deleteRentalAssistance = async (rentalAssistanceId) => {
...
 
+const createLease = async (lease, applicationId) => {
+  const postData = {
+    lease: lease,
+    application_id: applicationId
+  }
+  return apiCall('post', `/leases`, postData)
+}
+
+const updateLease = async (lease, applicationId) => {
```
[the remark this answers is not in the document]
akegan: I debated back and forth whether it made sense to have create and update lease resource on backend, seemed like a good idea to be explicit and match what we had for rental assistance, but open to feedback from non-bots on this too!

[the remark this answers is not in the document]
akegan: ended up going for code climate's advice here and consolidating on the frontend, leaving two functions separated on the backend

app/javascript/components/supplemental_application/actions.js
```diff
@@ -1,11 +1,31 @@
...
+  console.log('returning with a response', appResponse, leaseResponse)
+  return appResponse !== false && leaseResponse !== false
+}
+
+const updateLease = async (lease, primaryApplicantContact, applicationId) => {
+  if (!isEmpty(lease)) {
+    let leaseApi = domainToApi.mapLease(lease)
+    // TODO: We should consider setting the Tenant on a Lease more explicitly
+    // either via a non-interactable form element or using Salesforce
+    leaseApi['primary_applicant_contact'] = primaryApplicantContact
```
akegan, line 23: this is super duper awkward. I'm working on figuring out what the long term plan is for this Tenant field so we can come up with a better solution.

config/routes.rb
```diff
@@ -34,7 +34,9 @@
     namespace :v1 do
       get 'ami' => 'ami#get'
 
-      resources :applications, only: %w[index update]
+      resources :applications, only: %w[index update] do
+        resources :leases, only: %w[create update]
```
akegan, line 38: I ended up making leases a sub-resource of applications because that's how the architecture works on salesforce, in a follow-up PR I'd love to do more similar work to help make it clearer what belongs to what exactly

akegan:
I've also been considering whether or not it makes sense to add a "show" function to the leases controller, but as it's setup right now I don't think it makes a ton of sense to pull it out. 

It might make sense to do that work if we implement async loading of different sections of the supp app page.

app/javascript/api/request.js
```diff
@@ -0,0 +1,21 @@
+import axios from 'axios'
```
akegan, line 1: definitely open to other suggestions of file structure/ file naming / function naming around this new file

akegan:
ready for review!

app/controllers/api/v1/leases_controller.rb
```diff
@@ -0,0 +1,65 @@
...
+          logger.debug "lease update response: #{response}"
+          render json: { lease: response }
+        else
+          render status: 422, json: false
+        end
+      end
+
+      def lease_params
+        # TODO: Also require application_id
+        params.require(:lease).permit(
```
callaghanc, line 43: @akegan could we alphabetize these params?

akegan: done!

app/javascript/api/request.js
software-project: Maybe we could wrap it up into a apiRequest and expose methods like get, post and delete?

config/routes.rb
software-project: that is nice 👍

software-project (commented):
@akegan it looks good code wise, but I dont know how leases are working so if we could pair for a minute later on that would be nice.

spec/controllers/api/v1/leases_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+        post :create, params: params
+      end
+      expect(response).to have_http_status(:success)
+      json = JSON.parse(response.body)
+      expect(json['lease'].size).to eq(18) # create returns lease id
+    end
+  end
+
+  describe '#update' do
+    it 'updates a lease successfuly' do
```
callaghanc: @akegan tiny typo "successfuly" -> "successfully"
```suggestion
    it 'updates a lease successfully' do
```

spec/javascript/apiService.test.js
```diff
@@ -0,0 +1,43 @@
+import apiService from '~/apiService'
+// import requestThing from '~/api/requestThing'
```
callaghanc: @akegan can we remove this comment
```suggestion

```

akegan:
@callaghanc fixed your suggestions and fixed merge conflicts, ready for review again

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -50,17 +50,6 @@ def application_api_params
             :primaryApplicantContact,
```
callaghanc, line 49: @akegan copying my comment from slack here to better track on PR: "in the lease controller PR, it looks like we still call `submit_lease` in the short form controller’s submit action - is that intended? my impression was that we were replacing that with an explicit call from react to save the lease"

app/javascript/api/request.js
callaghanc: @akegan could be good to try out @software-project's suggestion here and see if you prefer the result

callaghanc (dismissed):
looks good! check out @software-project's suggestion to your question about the new request setup. otherwise seems to be working

app/javascript/api/request.js
```diff
@@ -0,0 +1,17 @@
...
+
+export const apiCall = async (method, path, data) => {
+  console.log('calling the api call OK')
```
akegan:
```suggestion
```

app/javascript/api/request.js
akegan: @software-project is this what you're suggesting? Just want to make sure I get it right

```
const apiCall = async (method, path, data) => {
  if (process.env.NODE_ENV === 'test') {
    var err = Error('API should not be called in TEST')
    console.error(err)
    throw err
  }
  try {
    const request = await axios[method](`/api/v1${path}`, data)
    return request.data
  } catch (e) {
    console.warn(e)
    return false
  }
}

export const get = async (path, data) => {
  apiCall('get', path, data)
}

export const post = async (path, data) => {
  apiCall('post', path, data)
}

export const delete = async (path, data) => {
  apiCall('delete', path, data)
}

export const put = async (path, data) => {
  apiCall('put', path, data)
}
```

akegan: @software-project just another quick ping on this in case this comment got lost in the shuffle of merging so many PRs yesterday.

software-project: Yes, that's perfect

akegan:
@software-project I added a couple of tests to get us up to the 50% threshold again 🤷‍♀️

spec/controllers/api/v1/leases_controller_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+          application_id: application_without_lease,
+          lease: {
+            'monthly_parking_rent': 100,
+          },
+        }
+        post :create, params: params
+      end
+      expect(response).to have_http_status(:success)
+      json = JSON.parse(response.body)
+      expect(json['lease'].size).to eq(18) # create returns lease id
```
software-project: Why is it 18 here? Shouldn't create add one lease each time?

software-project (approved):
Tested and it looks good. I'm not familiar with business logic here, but I was able to create and update `supplementals` on the application, controller is called as expected. I like how you extracted the lease code and made restful. My only wonder is about that one test.

spec/controllers/api/v1/leases_controller_spec.rb
akegan: it's the length of the lease ID string. I'll add a comment to make that clearer

software-project: oooo, ok, that makes sense now. Thank you

PR #197, opened 2019-03-13
Bugs/lease no preference used #164619509

akegan:
Small release bug
https://www.pivotaltracker.com/story/show/164619509

### DIFF
--- a/app/javascript/components/mappers/domainToApi/lease.js
+++ b/app/javascript/components/mappers/domainToApi/lease.js
@@ -3,9 +3,11 @@ import { currencyToFloat } from '~/utils/utils'
 
 // Lease is sent to backend with domain keys.
 export const leaseFieldMapper = {
+  'id': 'id',
   'lease_start_date': 'lease_start_date',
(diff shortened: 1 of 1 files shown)


callaghanc:
@akegan can you see if you can get tests to pass on this? i re-ran them a couple times but they keep failing

akegan:
just tested locally and tests passed once and failed once 🙁. @callaghanc tests just passed on this one, but I think we really need to work on having better indication of why tests are failing and deal with flakiness.

Locally I had problem that I did catch that was an auth error, I wonder if we could instead re-use the authentication somehow

callaghanc:
@akegan agree that we need to spend a little time on addressing the state of these tests to save us time re-running tests going forward, and also to ensure that there isn't some actual intermittent issue that's being ignored.  let's create a chore for improving Partners e2e tests if we don't already have one. we can merge this PR for now

callaghanc (approved):

akegan:
Ticket made: https://www.pivotaltracker.com/story/show/164644696

PR #199, opened 2019-03-19
Stop E2E tests from hanging on failure #162566752

akegan:
This commit also upgrades jest and removes use of fake timers

Pivotal ticket: https://www.pivotaltracker.com/story/show/162566752

As part of my testing to figure out why tests were hanging, I tried removing use of fake timers, which closes out this ticket too: https://www.pivotaltracker.com/story/show/162182016


akegan:
Testing for this is checking that 
1. If E2E tests have an issue, they fail at ~12-15 minutes (instead of running for over 45 mins)
2. If E2E tests pass, semaphore passes them!

Conveniently, tests are failing everywhere on partners right now 😄

package.json
```diff
@@ -66,10 +66,8 @@
   },
   "scripts": {
     "unit": "jest --testPathPattern=.*.test.js$ --coverage",
-    "e2e": "jest --testPathPattern=.*e2e.* --runInBand",
+    "e2e": "jest --testPathPattern=.*e2e.* --runInBand. --forceExit",
```
akegan, line 69: adding this forceExit is what caused it to actually exit. I tried using [--detectOpenHandles](https://jestjs.io/docs/en/23.x/cli#detectopenhandles) to determine what was wrong (which required the upgrade to 23.x), but it didn't return any results. 

Based on comments on https://github.com/facebook/jest/issues/1456, it seems like it's possible that the issue is not uncommon! and that --runInBand may be the culprit.

I tried running the tests in parallel, but ended up with a lot of salesforce auth errors from trying to sign in at the same time.

software-project (approved):
Looks good. All e2e test passed locally.

akegan:
I'm not going to fix the code climate issues, but I am waiting for semaphore job to pass

akegan:
merging despite code climate duplication warnings

PR #200, opened 2019-03-20
Add character limit to status update modal #162076858

akegan:
https://www.pivotaltracker.com/story/show/162076858


akegan:
pending sign-off from Josh about this solution but otherwise ready for review!

akegan:
got sign-off from josh on this solution, so this is good to review. Tests will fail until we can resolve the snapshot issue.

software-project (approved):
Tested on review app, works good.

PR #201, opened 2019-03-21
Save demographics info to primary applicant #162933550

akegan:
https://www.pivotaltracker.com/story/show/162933550


app/javascript/components/mappers/domainToApi/applicationShape.js
```diff
@@ -21,12 +22,15 @@ export const applicationShape = {
...
     mapFields(applicationShape, {}, application),
     [
       'alternateContact',
       'adaPrioritiesSelected',
       'demographics',
       'lease'
     ])
+  mappedApplication['primaryApplicant'] = extend(mappedApplication['primaryApplicant'], mappedApplication['demographics'])
+  delete mappedApplication['demographics']
+  return mappedApplication
```
akegan, line 35: it looks like we didn't have any testing around this function at all 🤦‍♀️

akegan:
This is ready for review. I added a quick test to verify that primaryApplicant was being mapped correctly.

In doing so, I discovered that we were putting the app member id as the application ID. Salesforce didn't seem to mind either way, so I updated the key to be factually correct. 

I've also commented out the E2E test that relies on snapshots existing. Once tests pass and the PR is removed, We can remove the commenting-out and force merge.

spec/javascript/components/mappers/domainToApi.test.js
```diff
@@ -0,0 +1,71 @@
+// TODO: Expand these tests to use fixtures, test all fields in application.
```
akegan, line 1: I'd like to follow-up this PR with another one that overhauls how we do a lot of our testing

callaghanc:
@akegan while QA'ing this i followed the instructions as written in the pivotal story, and saw that the demographic information was saved correctly in salesforce. then i went back into the paper app i had just entered and re-saved the form. that wiped out the demographic info in salesforce. could you look into this? seems like we are not populating the demographics data on edits of paper apps.

akegan:
This PR is working now, just need to add tests for mapApplication

app/javascript/components/mappers/domainToApi/applicationShape.js
akegan: I've added testing around this mapping now

spec/javascript/fixtures/application.js
```diff
@@ -13,8 +13,8 @@ const application = {
     "Id": "a0n0x000000AbE6AAK",
```
akegan, line 13: These fixture updates are because I started going down the path of calling mapApplication on the SOQL app then checking that it matched the domain app exactly, but abandoned that for now in favor of explicitly testing the demographics changes.

I kept the changes that I made to fixtures that make addresses clearer and add a middle name

spec/javascript/fixtures/short_form_submit_payload.js
```diff
@@ -35,7 +35,7 @@ const payload = {
        phone: '9548883321',
        phoneType: 'Cell',
        maritalStatus: 'Married',
-       applicationId: 'a0n0x000000AbE6AAK' },
+       appMemberId: 'a0n0x000000AbE6AAK' },
```
akegan, line 38: Salesforce doesn't seem to actually care what we send, but we were actually sending the appMemberId instead of the applicationId so I've renamed the field to make it clear. 

If we find later that we do in fact have to send the applicationId as part of the application member, we can add it in.

akegan:
@callaghanc tests should be passing now. We had to make some updates to fixtures post-refresh

callaghanc (approved):

PR #202, opened 2019-03-22
Fix Salesforce query for new files API #164835012

software-project:
https://www.pivotaltracker.com/story/show/164835012


akegan:
@software-project looks like my demographics PR merged ended up causing conflicts in your spec updates

software-project:
@akegan will fix them

akegan:
@software-project looks this PR right now causes code coverage to drop for JS. can you add a quick tests for the file download URL function?

software-project:
@akegan sure, will do

akegan:
@software-project QA looks good except that with the old files the files load as previews in a new window for me, and with the new link they download onto my machine. 

Doesn't seem like a huge deal but I'm wondering if you had any context as to why the behavior is different.

software-project:
@akegan I believe we are using different services. But the question is more for Larry, as he provided me with the url.

.vscode/launch.json
```diff
@@ -0,0 +1,33 @@
+{
```
akegan, line 1: @software-project is this intentional so that we can all debug with vscode? or should it be gitignored?

akegan (approved):
looks good just curious if vscode addition was intentional

.vscode/launch.json
software-project: yes, it's the config that allows to debug the tests in VS. But now I'm thinking maybe they should sit in some kind of docs, rather than here as not everyone might be using vscode. So I'm living it up to you.

akegan: hm. I don't really mind either way. I know catherine at least uses vscode so maybe it's helpful. I'm fine with leaving it in here. Maybe you can open up a separate PR that adds some documentation for how to use this file if you're using vscode

akegan:
QA looks good. ready to merge once tests pass.

akegan:
@software-project yarn unit is failing on semaphore

software-project:
@akegan checking

PR #205, opened 2019-03-28
Make status modal comment limit 255 #162627255

akegan:
https://www.pivotaltracker.com/story/show/162627255


chrisdolendo (approved):
Right on! I wonder if I should update my PR so this is not hardcoded 🤔

PR #206, opened 2019-03-28
Features/create eligibility information group #162770611

software-project:
https://www.pivotaltracker.com/story/show/162770611


spec/javascript/components/applications/application_form/PaperApplicationForm.test.js
```diff
@@ -64,4 +66,35 @@ describe('PaperApplicationForm', () => {
...
+      })
+    })
+
+    describe('sale listing', () => {
+      beforeEach(() => {
+        listing.is_sale = true
+        listing.is_rental = false
+      })
+
+      test('section should be empty', async () => {
```
akegan: looks like test name should be updated

akegan (changes requested):
Did a quick QA and although the checkboxes say they're required, they don't prevent saving an application and don't have any of the red highlighting or anything for validation.

I think we might want to update the wording slightly to match what we have on the webapp now. I'll get back to you with the updated copy shortly.

akegan:
@jrubenoff are you able to help with content on this ticket? I'm not sure how we should phrase the checkbox labels, but as it is right now they lack parallel structure and don't make a ton of sense. 

image

jrubenoff:
Looks like my AC didn't fully match [the original mockup.](https://docs.google.com/document/d/1ySRrQfATVctSFhVCNs4bbYOn3qD9x_LWH_qhUiafAHM/edit#heading=h.i8ji2sl7ey28)

Please:

- Add the following markup before the checkboxes:

```
<div class="form-group">
   <div class="columns">
      <strong class="t-small c-steel" id="prereqs">The applicant has&hellip;</strong>
   </div>
</div>
```
- Add the attribute `aria-labelledby="prereqs"` to each checkbox

jrubenoff:
@software-project For the error messages, could we display one error message for the entire group of checkboxes, instead of one per field?

If so, let's display this message if _any_ box is unchecked:

> The applicant can only qualify for the listing if all of the above are true.

If it's much easier to have one error message per field, let's display this:

> The applicant cannot qualify for the listing unless this is true.

spec/javascript/e2e/ApplicationNewPage.e2e.js
```diff
@@ -303,7 +303,6 @@ describe('ApplicationNewPage', () => {
     expect(values).toContain(TRUNCATED_FIRST_NAME)
     expect(values).toContain(TRUNCATED_LAST_NAME)
     expect(values).toContain(DATE_OF_BIRTH)
-    expect(values).toContain('Vision impairments;Mobility impairments;Hearing impairments')
```
akegan: why did this get removed?

app/javascript/components/applications/application_form/PaperApplicationForm.js
```diff
@@ -121,9 +121,9 @@ class PaperApplicationForm extends React.Component {
     }
     if (listing.is_sale) {
       validations = {
-        is_first_time_homebuyer: validate.isChecked('Field is required')(values.is_first_time_homebuyer),
-        has_completed_homebuyer_education: validate.isChecked('Field is required')(values.has_completed_homebuyer_education),
-        has_loan_preapproval: validate.isChecked('Field is required')(values.has_loan_preapproval),
+        is_first_time_homebuyer: validate.isChecked('The applicant cannot qualify for the listing unless this is true.')(values.is_first_time_homebuyer),
```
akegan: might be worth pulling these validation strings into a variable to avoid copy-pasta

akegan (commented):
one minor request, otherwise looks good. will do QA now

akegan (changes requested):
@software-project while doing QA I found that if you edit an application that has the checkboxes checked, that the boxes aren't checked on edit. 

Looks like the values are being saved in salesforce, but the default/edit values aren't being set correctly in the form 

https://dahlia-lap-full-pr-206.herokuapp.com/applications/a0o0t000000m01qAAA/edit

spec/javascript/e2e/ApplicationNewPage.e2e.js
software-project: Case I should have added it here in the first place. It's a sale listing test.

software-project:
@akegan I focused on `new application` and I even did went to edit. Mappers are added, so it should work as expected now. Also addressed your other comment.

akegan (approved):

PR #213, opened 2019-04-05
Use MultiDateField for lease start date #163398177

akegan:
https://www.pivotaltracker.com/story/show/163398177


akegan:
This PR is ready for review. Super open to feedback, especially on ruby updates!

akegan:
code climate coverage drop is because Ruby coverage is now being included in the total. We'll merge even with this drop

app/models/force/application.rb
```diff
@@ -88,17 +88,11 @@ def to_domain
...
-      field_names = salesforce_fields.keys
-      field_names.each do |field_name|
-        unless %w[Id Name].include?(field_name) || field_name.end_with?('__c')
-          salesforce_fields["#{field_name}__c"] = salesforce_fields[field_name]
-          salesforce_fields.delete(field_name)
-        end
-      end
-
-      salesforce_fields
+    def add_salesforce_suffix(salesforce_fields)
```
software-project: It looks like you are overriding the method and then calling the original, so don't think you need to override it. Just calling it in `to_salesforce` should be enough.

app/models/force/lease.rb
```diff
@@ -22,23 +22,45 @@ class Lease < Force::ObjectBase
...
+      domain_fields = super
+
+      if @fields.domain['lease_start_date']
+        domain_fields['lease_start_date'] = self.class.date_to_domain(@fields.domain['lease_start_date'])
+      end
+
+      domain_fields
+    end
+
+    def add_salesforce_suffix(salesforce_fields)
```
software-project: Same thing here

software-project (approved):
👍

PR #214, opened 2019-04-08
Do not show applications with null lottery numbers in lease-ups table #162001310

akegan:
https://www.pivotaltracker.com/story/show/162001310


chrisdolendo (approved):
Looks good!

PR #215, opened 2019-04-09
Only display valid preferences in lease up apps table and supp app #164963594

chrisdolendo:
Story: https://www.pivotaltracker.com/story/show/164963594

Based on discussions with MSol and Emily, we only display applications on the lease up apps table and preferences on the supp apps page whose preferences are valid (meaning they are **`Receives_Preference__c` == true** or the Receives Preferences is checked off). In addition to this -- moving forward, there must always be a household member assigned to a preferences.

I updated the code to reflect the new rules. Specifically, the SOQL query that pulls the lease up apps table index data so that we only pull valid preferences. As well as updated the supp app page to only show preferences of an application that are valid.


app/javascript/components/applications/application_form/preferences/utils.js
```diff
@@ -10,8 +10,8 @@ export const naturalKeyFromMember = (member) => {
 }
 
 export const memberNameFromPref = (id, householdMembers) => {
-  let member = find(householdMembers, { 'id': id }) || {}
-  return `${member.first_name} ${member.last_name}`
+  let member = find(householdMembers, { 'id': id })
+  if (member) { return `${member.first_name} ${member.last_name}` }
```
callaghanc: @chrisdolendo could we have this return an empty string if no member is found?

callaghanc (dismissed):

callaghanc:
@chrisdolendo could you add a bit of testing to this PR? perhaps for the `PreferencesTable` to test that it shows the correct output when a preference has a HH member and when a preference does not have a HH member?

chrisdolendo:
Thanks @callaghanc! Still waiting on @emilyshore on product rules for this bug, and will write tests when she decides to go ahead with this implementation for this rare use case.

spec/javascript/components/supplemental_application/sections/PreferencesTable.test.js
```diff
@@ -0,0 +1,29 @@
...
+        applicationMembers={applicationMembers}
+        onSave={onSave}
+        fileBaseUrl={fileBaseUrl}
+        onPanelClose={onDismissError}
+        formApi={formApi}
+      />
+    )
+
+    let tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
```
akegan: @chrisdolendo we're trying to avoid adding more tests like this where we're just checking that a snapshot matches because it's so hard to actually check that the snapshot contains what you expect. Can you update this test to explicitly check that your expected preferences are in there and the one's that you're not expecting to be in there aren't in there?

akegan (dismissed):

akegan:
@chrisdolendo QA looks good though, just update that test and should be good to go!

spec/javascript/components/supplemental_application/sections/PreferencesTable.test.js
chrisdolendo: Done!

spec/javascript/components/supplemental_application/sections/PreferencesTable.test.js
```diff
@@ -23,7 +22,6 @@ describe('PreferencesTable', () => {
       />
     )
 
-    let tree = component.toJSON()
-    expect(tree).toMatchSnapshot()
+    expect(wrapper.find('#alice-griffith-housing-development-resident-row').exists()).toBeTruthy()
```
akegan, line 25: @chrisdolendo are there any rows that shouldn't be there that we can check? I'm not clear on what all of the expected prefs are or are not.

akegan (changes requested):

akegan (approved):
looks great! thanks for adding additional test!

PR #219, opened 2019-04-11
Fix error on revoking expired tokens #162316506 

chrisdolendo:
Story: https://www.pivotaltracker.com/story/show/162316506

Instead of catching the error, I found I could just use `.try` instead of `.get` that will not throw an error if the token is already revoked. 

I also implemented a session timeout for Devise after 15 minutes.


akegan:
Just had a mohcd admin session run for ~40 mins and the session is still good, any idea why that might be?

akegan:
looks like it did work for a leasing agent but still not sure why it didn't work for a mohcd admin

chrisdolendo:
@akegan just tested and timeout for admin works, I adjusted to 5.seconds temp to test more easily and will revert back to 30.minutes after approval.

akegan:
@chrisdolendo just did a bit more testing, and it seems like having your session expire through devise doesn't do anything to revoke the salesforce session. If i let my session expire, then click login again I just get redirected without having to enter in my email address and password again.

I'm also wondering how to test the behavior for clicking "sign out" works now if my salesforce session expires. Do you know how long it takes for the salesforce session to expire (Edit: I looked it up, it's currently set to 2 hours)?

Should we be redirecting to the welcome page on session expiration instead of just showing an error when you click to go to another tab?

chrisdolendo:
@akegan Thanks! I will do an overwrite on devise's expire session to revoke the token, thought it would use the same method. Will look.

As for redirecting -- that just depends on the design decision. The current behavior displayed is built-in in Devise. I would say it would be 1-2pts of work, so I would consider creating a separate story for it.

chrisdolendo:
@software-project anything new on this?

software-project:
@chrisdolendo I will have another look on it in a minute.

lib/salesforce_logout_app.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
+
+# Handles Salesforce logouts when user session is timed out.
+class SalesforceLogoutApp < Devise::FailureApp
```
chrisdolendo, line 4: I would like to ignore the diff coverage on this unless someone knows how to effectively test this.

akegan: I think if could be pretty easy to just test that it does what it says it does, right? Check out tests here for examples https://github.com/Exygy/sf-dahlia-web/blob/master/spec/controllers/overrides/passwords_controller_spec.rb. Maybe it's not so easy though!

What's warden?

chrisdolendo:
@software-project 's Solution works well for a leasing agent but still doesn't work on Admin because the `warden.params` indicating that the user is an admin gets overwritten sometimes (aka feature flickers for an admin). The ideal rails implementation for this is to utilize scope on Warden to create an admin scope in managing sessions, but will take a few hours to implement. To close this ticket, I set the session expiry for sessions on Salesforce to 15 minutes which is the same session expiry as Devise. 

Moreover, the `.try` method on revoking the token removes the error which is the original issue of this ticket.

chrisdolendo:
I just test this as an admin and the salesforce session expiry works.

lib/salesforce_logout_app.rb
chrisdolendo: hmm @software-project do you have an idea on how we can test this because it's easily "callable"?

chrisdolendo: @akegan I can't find an easy way to test for this without doing an e2e test and having them wait 15 mins for the user session to timeout. Apparently, controller tests won't really work for this -- https://github.com/plataformatec/devise/issues/3409. We can see if @software-project has something to share.

akegan:
@chrisdolendo did you test whether or not the salesforce session will be extended if the user is active (calling for new data, etc?). 

Working on figuring out what makes sense as our session expiration time now

akegan:
@chrisdolendo looks like 15-30 mins is the standard. Let's go with 30 mins. (maybe slightly longer on salesforce?)

Still checking with larry and msol about impact across the rest of salesforce

akegan:
@chrisdolendo we've gotten sign-off from Msol and larry to keep it at 30 mins

akegan (approved):

akegan:
@chrisdolendo this is ready to merge once you update to 30 mins!

chrisdolendo:
Changed to 30 minutes! Merging.

PR #220, opened 2019-04-11
Features/add additional phone number to appl #165161032

software-project:
https://www.pivotaltracker.com/story/show/165161032


spec/javascript/fixtures/short_form_submit_payload.js
```diff
@@ -25,6 +28,8 @@ const payload = {
        lastName: 'jones',
        middleName: null,
        address: '123 MAIN ST',
+       alternatePhone: null,
```
akegan: can we update fixture to have these values non-null so we can trace them through and make sure they show up where expected in snapshot, etc?

akegan:
one question re: tests, but otherwise QA looks good!

PR #222, opened 2019-04-12
Require language on paper application #162621696 

software-project:
https://www.pivotaltracker.com/story/show/162621696


app/javascript/components/applications/application_form/ApplicationLanguageSection.js
```diff
@@ -11,8 +11,14 @@ const ApplicationLanguageSection = ({ formApi, editValues }) => {
...
       <div className='row'>
         <div className='small-4 columns'>
-          <label>Language Submitted In</label>
-          <Select field='application_language' options={applicationLanguageOptions} />
+          <Field.Select
+            label='Language Submitted In'
+            blockNote='(required)'
+            id='application_language'
+            field='application_language'
+            errorMessage={(_, error) => error}
```
akegan, line 19: Seems like we could cover this if we had a test that tried to submit without filling out any of the required fields. We might already have a test for some of them that you could add this too

software-project: I'm almost done with the test for it.

spec/javascript/components/applications/application_form/PaperApplicationForm.test.js
```diff
@@ -66,6 +67,23 @@ describe('PaperApplicationForm', () => {
         `${errorMsgSel}.error`
       expect(wrapper.find(errorElementsSel).text()).toEqual('Please enter a valid dollar amount.')
     })
+
+    test('should validate language', async () => {
```
akegan: looks like from above that the thought for these tests is to have the test name be the field name?

So then it's

describe PaperApplicationForm
  describe should validate fields correctly: 
     test Annual income
     _test Language_

not sure if that's the best way to go about it, but I think we should at least make the naming consistent and have it make sense

akegan (commented):
One small comment about test naming again. Otherwise QA and everything looks good

akegan (approved):

PR #223, opened 2019-04-12
Setup sentry #165160753

akegan:
https://www.pivotaltracker.com/story/show/165160753

*Task before merge*: Remove errors added to test sentry


app/controllers/api_controller.rb
```diff
@@ -17,6 +21,8 @@ def render_error(opts = {})
     else
       message = 'Not found.'
     end
+    p 'yo'
```
chrisdolendo: let's remove this

app/javascript/components/listings/ListingsPage.js
```diff
@@ -51,6 +52,7 @@ const layout = {
 }
 
 const ListingsPage = ({ page, listings }) => {
+  Sentry.captureException(new Error('This is my fake error message with sourcemaps hopefully'))
```
chrisdolendo: Do we want to keep this message on there or update?

app/views/layouts/application.html.slim
```diff
@@ -27,4 +28,5 @@ html
     == render 'layouts/icons'
 
     javascript:
-      var SALESFORCE_BASE_URL = "#{{current_user&.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL']}}"
\ No newline at end of file
+      var SALESFORCE_BASE_URL = "#{{current_user&.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL']}}"
```
chrisdolendo: I think we need to remove this JS because we brought it in differently

chrisdolendo (changes requested):
Just some cleanup of the code.

Also getting failed semaphore tests with Bundler.. `You must use Bundler 2 or greater with this lockfile.`

app/javascript/components/listings/ListingsPage.js
akegan: This is part of the task before merge. Once you approve, I'll remove these errors before merging

akegan:
@chrisdolendo I've updated semaphore job to work with bundler v2.0.0, looks like it works OK with master too.

I've also made other suggested updates! Semaphore will still until I remove errors that need to be removed before merge

chrisdolendo (approved):

PR #226, opened 2019-04-17
Memoize lending institutions on class level #165403854

software-project:
https://www.pivotaltracker.com/story/show/165403854


app/services/force/custom_api/lending_institutions_service.rb
```diff
@@ -5,32 +5,36 @@ module CustomApi
     # Provide Salesforce custom API interactions for rental assistances
     class LendingInstitutionsService < Force::Base
       def lending_institutions
+        self.class.lending_institutions(@client)
```
akegan, line 8: i think that my future self would probably appreciate a comment somewhere in this file explicitly stating that the the values are being memoized and will be reloaded on server restart

software-project: Sure, will add a comment

akegan (approved):
verified that log was logged once on server start, then not again, even with refresh of page, or logout and log back in

PR #230, opened 2019-04-23
Downgrade webpack-dev-server to previous version #165548965

akegan:
https://www.pivotaltracker.com/story/show/165548965


callaghanc (approved):
i've tested this locally and this fixes the issue with running webpack-dev-server

PR #238, opened 2019-05-03
Upgrade paper application to react form #162457306

akegan:
https://www.pivotaltracker.com/story/show/162457306


akegan:
Starting to do an initial QA:

(not sure if you've already fixed these)
1. When saving application with validation, alert box shows up, but page doesn't scroll to the alert box
image
2. Alert box doesn't close when "x" is clicked on 

3. HH member shows up as required even if none of the fields are populated. This does not match behavior on full 
image

akegan:
Looks like we're missing the red border around fields for all validation errors

before (full):
image

after (this branch):
image

akegan:
OK last QA comment.

We lost the behavior of checking that all 3 dob fields are touched before showing error. Now if you type in a month you get a red error saying you have an invalid date even if you haven't touched day or year yet
image

I will try to fix after my next meeting (should be out by 3:45)

akegan:
oh dear @software-project dob validation looks worse now. I'll take a stab at fixing it

akegan:
we also appear to have lost email validation

app/javascript/components/applications/application_form/AddressForm.js
```diff
@@ -1,20 +1,20 @@
...
+import { FieldWrapper } from '~/utils/form/final_form/Field'
 import { isEmpty } from 'lodash'
 import { maxLengthMap } from '~/utils/formUtils'
 
-const buildField = (memberType, nestedField, fieldMap, fieldKey) => {
+const buildField = (memberType, name, fieldMap, fieldKey) => {
   // to do: refactor this to use props
   if (memberType === 'primaryApplicant') {
-    return fieldMap[fieldKey]
-  } else if (memberType === 'householdMember') {
```
callaghanc, line 10: @akegan looks like we used to expect "householdMember" as a value for `memberType` here, but now we don't - what's the context around that change?

app/javascript/components/applications/application_form/AddressForm.js
```diff
@@ -1,20 +1,20 @@
 import React from 'react'
-import { Text } from 'react-form'
+import { FieldWrapper } from '~/utils/form/final_form/Field'
 import { isEmpty } from 'lodash'
 import { maxLengthMap } from '~/utils/formUtils'
 
-const buildField = (memberType, nestedField, fieldMap, fieldKey) => {
+const buildField = (memberType, name, fieldMap, fieldKey) => {
```
callaghanc: @akegan could we update the variable `name` to have a slightly more descriptive name? i'm not too familiar with the paper app form so i'm not sure what `name` is here. same for use of `name` in `AddressForm` component below.

akegan: looks like name is used in the case of FieldArrays to pass down what the name is for form item in the value in the array. 

For example. for a household member, we'd expect it to be e.g. "household_members[0]", then the e.g. first name field would have the value "household_members[0].first_name" as it's "name". Maybe fieldName would make more sense?

app/javascript/components/applications/application_form/AddressForm.js
akegan: I didn't make this change, but it seems like it was an update to again accomodate the field array syntax. Maybe we should actually put this first, so the logic is: use the name if you have it, and fall back to applicant or alternate_contact if the name isn't defined.

@software-project do you have any more context?

software-project: This is due to different way that forms are working with final form. HouseholdMembers are using FormArray. But in fact I can simplify it further.

callaghanc: @software-project Ok that's a lot simpler! Could we also differentiate the var names `fieldName`, `fieldMap`, and `fieldKey` a little bit more? Seems like the field in `fieldName` is like a level higher than the field in `fieldMap` and `fieldKey`. If that's the case, it could be helpful to indicate that in the var names. Does something like `subFieldMap` and `subFieldKey` make sense? Or `addressFieldMap` and `addressFieldKey`? Something to show that there's this parent field with nested address fields.

app/javascript/components/applications/application_form/preferences/PreferencesSection.js
```diff
@@ -33,41 +34,44 @@ class PreferencesSection extends React.Component {
...
+          {({ fields }) =>
+            <React.Fragment>
+              { fields.map((name, index) => {
+                return (
+                  <div className='border-bottom margin-bottom--2x' key={name}>
+                    <PreferenceForm {...{index, name, form, listingPreferences, fullHousehold}} />
+                  </div>)
+              })
+              }
+              <div className='row'>
```
callaghanc, line 58: @software-project does the row with the Add Preference button need to be included within the `FieldArray` component?

spec/javascript/components/applications/__snapshots__/ApplicationEditPage.test.js.snap
```diff
@@ -1134,12 +1134,12 @@ Array [
                               className=""
                               id="household_members_0_first_name"
                               maxLength={40}
-                              name="household_members[0].first_name"
+                              name="undefined.first_name"
```
akegan: @software-project looks like a bug was introduced here

app/javascript/components/applications/application_form/preferences/PreferencesSection.js
software-project: @callaghanc we paired on this one with Chris. We had had some hard times to make it work and moving it here worked. Maybe there is some other way, to keep it outside, but we couldn't find it.

akegan:
@software-project looks like application save is totally broken now .

Doesn't look like your fault though:
```
<< API Error >> Faraday::ClientError, APEX_ERROR: API_LeaseUp_ApplicationService.RecordLockException: List index out of bounds: 0 
```

software-project:
> @software-project looks like application save is totally broken now .
> 
> Doesn't look like your fault though:
> 
> ```
> << API Error >> Faraday::ClientError, APEX_ERROR: API_LeaseUp_ApplicationService.RecordLockException: List index out of bounds: 0 
> ```

@akegan when are you getting this exception. It looks more like Salesforce issue.

akegan:
@software-project yeah it was salesforce issue. Fixed now

akegan:
heads up @software-project e2e tests are failing on application new page save

software-project:
@akegan issue is with latest changes to validation. Household member is not validated at all. Will try to fix.

app/javascript/components/applications/application_form/HouseholdMemberForm.js
```diff
@@ -1,72 +1,72 @@
...
 import formOptions from '~/components/applications/application_form/formOptions'
+import { FieldWrapper, SelectField } from '~/utils/form/final_form/Field'
+import { MultiDateField } from '~/utils/form/final_form/MultiDateField'
 
 const {
   relationshipOptions
 } = formOptions
 
-const HouseholdMemberForm = ({ i, formApi }) => {
+const HouseholdMemberForm = ({fieldName, index, form}) => {
```
callaghanc: @software-project it's not clear here what type of information `fieldName` is. can we update the variable name or add a comment to further explain?

app/javascript/utils/form/validations.js
```diff
@@ -102,4 +102,26 @@ validate.any = (...fns) => (value) => {
       map(fns, fn => fn(value))))
 }
 
+validate.isValidDOB = (person, errors, isPrimaryApplicant = false) => {
```
callaghanc, line 105: @software-project how would you feel about using an options object for the `isPrimaryApplicant` argument? e.g. `opts = {}`. then the call is `isValidDOB(person, errors, {isPrimaryApplicant: true})`, which indicates what that last argument is for, instead of `isValidDOB(person, errors, true)`, which doesn't indicate what `true` will do and makes the dev have to go look up the function

app/javascript/components/applications/application_form/HouseholdMemberForm.js
software-project: @callaghanc what name do you find more descriptive? `formNamePrefix`, `fieldNamePrefix`? It's basically nested field name prefix. It's required for final for 'FieldArray'

akegan:
@software-project looks like e2e test for selecting live/work pref is broken. Also branch is out of date with master

app/javascript/components/applications/application_form/HouseholdMemberForm.js
callaghanc: @software-project what values do we expect for `fieldName`? is it going to be "householdMember", or is it going to be the natural key?

software-project: It's household member object name in the array, eg. `household_members[2]`

callaghanc: ah ok thanks, so will the value of `fieldName` always be `` `household_members[${index}]` ``, similar to how it was before? or is there another possible value for `fieldName` that wouldn't be equal to `` `household_members[${index}]` ``?

software-project: yes

callaghanc: ok cool, then what do you think about passing ``fieldName={`household_members[${index}].middle_name`}`` to the fields in this form and removing the `fieldName` prop from `HouseholdMemberForm` entirely? that makes one less thing to think about here, since `fieldName` isn't really bringing any info that isn't already in `index`

software-project: that would work, sure.

callaghanc: cool i'll push that change

akegan:
I can't officially approve because I opened the PR but just did another round of solid QA and it loooks good!!! let's merge! @callaghanc @software-project you can approve and merge!

akegan:
OK did some more QA around HH members and still looks good! @callaghanc please approve and merge! 🎉

callaghanc (approved):
approved based on @akegan's qa

PR #239, opened 2019-05-06
[Hotfix] Stop flagged app comment field from losing focus on typing #165830985

akegan:
https://www.pivotaltracker.com/story/show/165830985

More details are on ticket, but TL;DR is that a react-table update changed the way tables were re-rendered and introduced issues if you pass columns as a function or build a component inside your ReactTable render rather than passing columns as a fully created object.


callaghanc (changes requested):
@akegan to extend e2e test to include setting review status and checking that save worked

callaghanc (approved):

PR #242, opened 2019-05-13
Removed loader and added saving text #164967254

LoganArnett:
This PR addresses Pivotal [#164967254](https://www.pivotaltracker.com/story/show/164967254)

I also added in a `start` script that allows us to run the project with only 1 terminal window and one command for both services using `npm-run-all`. I also added a note to the README in relation to this script as well as a little help hint that I didn't realize I needed when installing


akegan:
@LoganArnett heads up looks like unit tests are failing. You may just need to update snapshots with `yarn run unit -u`

akegan (changes requested):
Did a quick QA locally, all looks good. Just one more AC to meet:

2. Disable status modal buttons when saving as well.

Now that we don't have the full page loading state, we want to make sure that users aren't clicking around and trying to save multiple ways while the page is loading.

Here are the buttons to disable:

Button where it says "Processing" here at the bottom of the page:
image

and these two buttons at the top of the page too ("Processing" and "Add a comment":
image

LoganArnett:
Updated this with the disabled state for the dropdowns and the add comment button

akegan:
@LoganArnett looks like you need to update snapshots again after this change

akegan:
also just found one more button to disable, whoops. It's in the status history at the bottom of the page. 
image

So many buttons to save on one page!

LoganArnett:
True, didnt think about that, sending the updated snapshots now

LoganArnett:
Checking on the e2e issue now

LoganArnett:
Looks like the flaky tests like you mentioned, all passing now

akegan (approved):

akegan:
Looks good, just needs snapshots updated again

PR #246, opened 2019-05-14
Upgrade supplemental application to react final form #165735950

software-project:
https://www.pivotaltracker.com/story/show/165735950


app/javascript/components/applications/application_form/formOptions.js
```diff
@@ -1,21 +1,30 @@
...
+    _.map(options, (option) => (
+      {
+        value: option.hasOwnProperty('value') ? option.value : option,
+        label: option.label || option
+      }
+    ))
+  )
 }
 
+// TODO: Remove the use of labelize on all the below options arrays once the
```
chrisdolendo: Do we need to create a chore for this so we don't forget it?

callaghanc: You could add it as an AC/task on the Pivotal story if you feel that would suffice.

akegan: added! https://www.pivotaltracker.com/story/show/165735996

PR #247, opened 2019-05-15
Add SubStatus for Application 158365756

LoganArnett:
This PR handles this [ticket](https://www.pivotaltracker.com/story/show/158365756).

It adds in the Sub Status dropdown to the `StatusModalWrapper` component

### Demo:
demo


akegan:
looks good! 

Only notes:
1. Please add review instructions to the ticket
2. Maybe extend the e2e test a bit to actually select a substatus if there are any and save
3. Would it be feasible to add a unit test to check that the substatuses update when you update the status? [Here's an example](https://github.com/Exygy/sf-dahlia-lap/blob/master/spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js#L96) with some tests that don't involve snapshots, but I'm open to input on the best plan for testing!

LoganArnett:
Updated the review instructions, and added sub status to the unit and e2e testing for status modal

akegan (approved):
Looks good!

PR #248, opened 2019-05-20
Add validation to check for valid naturalKey selection #165942084

akegan:
https://www.pivotaltracker.com/story/show/165942084

Strategy ended up being:
1. Keep field-level validation for normal checking if app member is present with validate.isRequired.
2. If there is a value for naturalKey on the preference, then check and make sure it's valid on the form level.

### DIFF
--- a/app/javascript/components/applications/application_form/PaperApplicationForm.js
+++ b/app/javascript/components/applications/application_form/PaperApplicationForm.js
@@ -14,7 +14,8 @@ import AlertBox from '~/components/molecules/AlertBox'
 import validate from '~/utils/form/validations'
 import { Form } from 'react-final-form'
 import arrayMutators from 'final-form-arrays'
-import { isEmpty } from 'lodash'
+import { includes, isEmpty } from 'lodash'
+import { naturalKeyFromMember } from './preferences/utils.js'
 
 class PaperApplicationForm extends React.Component {
   constructor (props) {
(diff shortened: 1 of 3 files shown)


app/javascript/components/applications/application_form/PaperApplicationForm.js
```diff
@@ -53,6 +54,22 @@ class PaperApplicationForm extends React.Component {
...
+    if (values.preferences) {
+      // Combine household members with primary applicant to get list of all the natural keys
+      const allMembers = [values.applicant].concat(values.household_members ? values.household_members : [])
+      const naturalKeys = allMembers.map((member) => naturalKeyFromMember(member))
+      errors.preferences = []
+      values.preferences.map(
+        (pref, i) => {
+          errors.preferences = errors.preferences.concat({})
+          if (pref.naturalKey && !includes(naturalKeys, pref.naturalKey)) {
+            errors.preferences[i].naturalKey = 'This field is required'
```
akegan, line 66: Note, the required message actually varies for all of these fields generally, but I wasn't able to get this validation to work sufficiently at the field level so had to bump up to form-level. 

I think we could: Keep message as "This field is required" in this use case, or set the error messages at a higher level and try to match the validation message to the given preference. 

Got sign-off from emily on this decision

spec/javascript/components/applications/application_form/PaperApplicationForm.test.js
```diff
@@ -9,7 +9,8 @@ import application from '../../../fixtures/domain_application'
 import lendingInstitutions from '../../../fixtures/lending_institutions'
 
 const mockSubmitApplication = jest.fn()
-const applicationWithInvalidAnnualIncome = clone(application)
+
+let testApplication = null
```
akegan, line 13: this refactor is because originally I had tried to add a test to this file, but interactivity of preferences wasn't working right for me 🙁

app/javascript/components/applications/application_form/PaperApplicationForm.js
```diff
@@ -53,6 +54,22 @@ class PaperApplicationForm extends React.Component {
       errors.alternate_contact.last_name = validate.isPresent('Please enter a Last Name')(values.alternate_contact.last_name)
       errors.alternate_contact.email = validate.isValidEmail('Please enter a valid Email')(values.alternate_contact.email)
     }
+
+    // Secondary preference application member validation: Check if selected app member value is still a valid natural key.
+    if (values.preferences) {
+      // Combine household members with primary applicant to get list of all the natural keys
+      const allMembers = [values.applicant].concat(values.household_members ? values.household_members : [])
```
software-project: I think we need to filter `allMembers` and use them then they are valid. Right now form breaks when I click `Add Preference` when HH member is missing date or when I delete it.

software-project (changes requested):

app/javascript/components/applications/application_form/PaperApplicationForm.js
akegan: excellent bug hunting! will investigate

akegan:
@software-project fixed pushed, please review again

software-project (approved):
That's perfect now!

PR #252, opened 2019-05-22
Memory Usage config updates #163738628

LoganArnett:
This PR is for [ticket #163738628](https://www.pivotaltracker.com/story/show/163738628). This ticket was for investigating the memory usage escalating in Production. 

While running the app locally in either development or production mode I was able to see a stable memory usage of roughly 300mb:
image

This PR also includes the `oj` gem for Optimized JSON parsing

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -15,7 +15,7 @@ port        ENV.fetch('PORT') { 3000 }
 
 # Specifies the `environment` that Puma will run in.
 #
-environment ENV.fetch('RAILS_ENV') { 'production' }
+environment ENV.fetch('RAILS_ENV') { 'development' }
 
 # Specifies the number of `workers` to boot in clustered mode.
 # Workers are forked webserver processes. If using threads and workers together
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -55,11 +55,6 @@ preload_app!
 #   ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
 # end
 #
-# before_fork do
-#   require 'puma_worker_killer'
-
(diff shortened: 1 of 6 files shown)


config/scout_apm.yml
```diff
@@ -0,0 +1,34 @@
+# This configuration file is used for Scout APM.
+# Environment variables can also be used to configure Scout. See our help docs at http://docs.scoutapm.com#environment-variables for more information.
+common: &defaults
+
+  # key: Your Organization key for Scout APM. Found on the settings screen.
+  # - Default: none
+  key: s4ehrWvArTiWqSpyvLHZ
```
akegan: is this OK to be public?

config/puma.rb
```diff
@@ -53,6 +55,11 @@
 #   ActiveRecord::Base.establish_connection if defined?(ActiveRecord)
 # end
 #
+# before_fork do
```
akegan: is this here so we can un-comment if we decide to use it?

akegan (commented):
just a couple of questions, otherwise looks good.

config/scout_apm.yml
LoganArnett: That is a good question, it should most likely be moved to an ENV, and I assume that you will want to use the key for your account as the main one

config/puma.rb
LoganArnett: removed this for now

config/scout_apm.yml
LoganArnett: moved to `SCOUT_KEY='SCOUT_KEY_HERE'`

akegan (approved):
looks good

PR #256, opened 2019-05-24
Currency fields #165205282

software-project:
https://www.pivotaltracker.com/story/show/165205282


akegan (changes requested):
A couple of issues with QA:
1. Rental assistance amounts don't have currency field, don't load with $ format built-in
1. Lease values (rent, etc) don't load with $ format built in (should be able to add it in the lease model
1. Please update paper application to be consistent if it's not too much work

software-project:
@akegan should be ready for a review

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -216,6 +217,8 @@ class SupplementalApplicationPage extends React.Component {
       rentalAssistance,
       this.state.persistedApplication.id
     )
+    // show price in right format
+    rentalAssistance.assistance_amount = formUtils.formatPrice(rentalAssistance.assistance_amount)
```
akegan, line 221: @software-project I wonder if we could get this line covered with a snapshot test that verifies that currencies are formatted with $, etc.

spec/fixtures/model/force/lease/lease_domain.json
```diff
@@ -2,11 +2,11 @@
   "application_id": "a0o0P00000FFc2yQAD",
   "id": "a130U0000001n4wQAA",
   "lease_start_date": {"year": "2028", "month": "01", "day": "01"},
-  "monthly_parking_rent": 100,
-  "monthly_tenant_contribution": 800,
+  "monthly_parking_rent": "$100.00",
```
akegan: maybe increase some of these sample currencies to >=$1,000.00 to catch the lack of commas issue.

akegan (changes requested):
More QA notes:
- For some reason, when currencies are loaded for leases and rental assistances they are not formatted with commas (they do have the dollar sign and cents).
- @software-project should I move changing paper app currencies to currency field to a new ticket, or can you include in this PR?
- See comment below re: tests. 

Pic for currency formatting issue on load:
image

akegan:
@software-project please also add review instructions to the ticket

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
software-project: It's already covered in rental-assistance e2e test. I modified it just now to also test for format with comma.

akegan:
getting weird behavior for non-number characters 
weirdCurrencyBehavior

Looks like in the [react final form example](https://codesandbox.io/s/3rp260ly51?from-embed)it also gets rid of letters if you type them in but doesn't have the odd behavior around commas

akegan:
otherwise looks good! if the weird behavior from chars after commas is unavoidable I can check with Emily to see if we can keep it as is

software-project:
@akegan should be working fine now.

akegan (approved):
OK looks good now. Thanks!

PR #261, opened 2019-05-31
Bug - Open House Inconsistencies 166098541

LoganArnett:
This PR covers ticket [#166098541](https://www.pivotaltracker.com/story/show/166098541) regarding open house information not being parsed correctly.

It appears that this was due to the helper function forcing the fields to lowercase so even though we saw a date it was incorrect as the field was pulling from `date` instead of `Date` and the times were attempting to pull from `start_time` and `end_time` instead of `Start_Time` and `End_Time`.


spec/javascript/components/listing/__snapshots__/ListingPage.test.js.snap
```diff
@@ -84,24 +84,7 @@ exports[`ListingPage individual fields Additional Information fields 1`] = `
...
       <div
         className="content-card"
-      >
-        <div
-          className="margin-bottom--half"
-        >
-          <h4
-            className="t-sans t-small t-bold no-margin"
-          >
-            Legal Disclaimers
```
akegan: @LoganArnett do you know why so many lines were removed in this snapshot?

akegan (changes requested):
Update. Looks like snapshot lines missing are a regression.

Looking at [mosso listing](https://dahlia-lap-full-pr-261.herokuapp.com/listings/a0W0P00000H3hJsUAJ) you can see the difference 
here's mosso on full:
image

vs this PR:
image

LoganArnett:
Weird, I will update with master and see what is going on

LoganArnett:
@akegan found the issue here and updated

akegan:
@LoganArnett one last request, can you update the fixture that we have so that the open house info is correctly displayed in snapshot? looks like after latest change there's no test update so no test coverage for the open house info

akegan (approved):
Looks good to merge! 👍

PR #264, opened 2019-06-11
Block New Application Unless Lottery is Not Yet Run #164899513

LoganArnett:
This PR covers [ticket #164899513](https://www.pivotaltracker.com/story/show/164899513) for redirecting on new applications for a listing where the lottery_status is anything other than `Not Yet Run`

### DIFF
--- a/app/controllers/listings/applications_controller.rb
+++ b/app/controllers/listings/applications_controller.rb
@@ -13,6 +13,9 @@ module Listings
     end
 
     def new
+      if @listing.Lottery_Status != 'Not Yet Run'
+        redirect_to "/listings/#{@listing.Id}"
+      end
       @lending_institutions = listing_service.sale?(@listing) ? lending_institutions : {}
     end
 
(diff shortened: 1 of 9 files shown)


app/controllers/listings/applications_controller.rb
```diff
@@ -13,6 +13,9 @@ def index
     end
 
     def new
+      if @listing.Lottery_Status != 'Not Yet Run'
+        redirect_to "/listings/#{@listing.Id}"
```
chrisdolendo: you can probably use something more lexical like `listing_url` + param instead of interpolated string

chrisdolendo: You could use the semantic `listing_url` + param as opposed to the interpolation. Also might be cleaner to make this a before action.

chrisdolendo (changes requested):

chrisdolendo:
:+1: Looks good! just had a one suggestion.

app/controllers/listings/applications_controller.rb
LoganArnett: Updated

chrisdolendo (approved):
Look great! Thanks for the great tests

PR #268, opened 2019-06-24
Fix failing E2E tests #166881191

akegan:


spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
```diff
@@ -8,12 +10,32 @@ const testStatusModalUpdate = async (page) => {
...
   // Enter a comment into the status modal comment field
   await page.type('#status-comment', COMMENT)
   // Submit the status modal form
   await page.click('.form-modal_form_wrapper button.primary')
+
+  // Verify that no form-field errors are present on save
+  const errors = await page.$$eval('span.error', divs => divs.map(d => d.textContent))
+  expect(errors).toStrictEqual([])
+  // Verify that no salesforce save errors present on save
+  try {
```
akegan: @LoganArnett I'm having a hard time getting this to work. 

My goal is to have a useful error that pops up if there's an issue saving substatuses, ideally with the specific status and substatus that's been sent logged. 

The issue is that asserting that something isn't present while allowing for some time to get a response isn't working great

LoganArnett: I am not sure if this is the best way to handle this. Another option would be to use another add-on that can actually listen to `http` requests. Something like Sinon gives you the ability to do something like:
```
page.on('response', response => {
    const endpoint = response.url().split('/').pop();
    if (endpoint === "/update_status") {
        const responseBody = await response.json();
        console.log(responseBody)
        // Here we can decipher whether it was an error or not form the API side
    }
})

Comment (LoganArnett) [reply spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js]: This is the simpler smaller option that is available through Puppeteer but Sinon would allow us to mock and digest more API requests if we see a need for it

Comment (akegan) [reply spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js]: @LoganArnett what's the simpler smaller option?

**Comment by akegan** [comment]
This is passing both on semaphore-full and semaphore-prod now, so it's ready for review!

File: spec/javascript/e2e/ApplicationNewPage.e2e.js [hunk 2]
```diff
@@ -29,6 +29,9 @@ describe('ApplicationNewPage', () => {
 
     // Save the application
     await page.click('.save-btn')
+    // Verify that no errors are present on save
```
LoganArnett, line 32: Everything looks good but this block of code is in here a number of times so maybe we should make a simple helper function even just at the top of this file so we do not have the duplicate blocks throughout

LoganArnett (changes requested):

akegan:
@LoganArnett ready for re-review

LoganArnett (approved):
Looks good to me

PR #269, opened 2019-06-25
Fix flagged app listing dropdown #166685619

akegan:
[Pivotal Ticket](https://www.pivotaltracker.com/story/show/166685619)

Sorted by name rather than lottery date, set default filter to "all", clean up code


LoganArnett (approved):
Tested on the deployed version and looks good

PR #270, opened 2019-07-02
Prevent status updates if sup form is invalid #166453291]

chrisdolendo:


app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -19,8 +19,19 @@ import { withContext } from './context'
 import StatusModalWrapper from '~/components/organisms/StatusModalWrapper'
 import validate from '~/utils/form/validations'
 
-const StatusUpdateSection = withContext(({ store }) => {
+const StatusUpdateSection = withContext(({ store, form, checkForValidationErrors }) => {
```
LoganArnett: Since these are being passed in here could we not just pass a prop of `isFormInvalid` or `isFormValid` with this check being run from the container component instead of calling it within these child components?

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -29,8 +40,8 @@ const StatusUpdateSection = withContext(({ store }) => {
         substatus={recentStatusUpdate.substatus}
         comment={recentStatusUpdate.comment}
         date={recentStatusUpdate.date}
-        onStatusDropdownChange={openUpdateStatusModal}
-        onAddCommentClick={openAddStatusCommentModal}
+        onStatusDropdownChange={() => onStatusDropdownChange(form)}
```
LoganArnett: Since these are referencing the `form` from the props we could just as easily still use them without the additional arrow function here and call to the `isFormInvalid` prop:
```js
const onStatusDropdownChange = () => !isFormInvalid ? openUpdateStatusModal() : null

onStatusDropdownChange={onStatusDropdownChange}
```

LoganArnett: And same for the comment

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -104,11 +115,17 @@ const DemographicsSection = () => {
   )
 }
 
-const StatusHistorySection = withContext(({ store }) => {
+const StatusHistorySection = withContext(({ store, form, checkForValidationErrors }) => {
   const { statusHistory, openAddStatusCommentModal, loading } = store
+
+  const onAddCommentClick = (form) => {
+    let invalidForm = checkForValidationErrors(form)
```
LoganArnett: Same comment here for the prop to already pass the boolean

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -104,11 +115,17 @@ const DemographicsSection = () => {
...
+    if (!invalidForm) { openAddStatusCommentModal() }
+  }
+
   return !isEmpty(statusHistory) && (
     <ContentSection.Sub title='Status History' borderBottom={false}>
-      <StatusList items={statusHistory} onAddComment={openAddStatusCommentModal} commentDisabled={loading} />
+      <StatusList items={statusHistory} onAddComment={()=> {onAddCommentClick(form)}} commentDisabled={loading} />
```
LoganArnett: And same structure here

LoganArnett (changes requested):

LoganArnett:
Looks good, my change wont work because I missed that we were within the form render function but other than that it looks much cleaner 👍

LoganArnett (approved):

PR #271, opened 2019-07-08
Add "required" block notes to required fields #166852285

akegan:
https://www.pivotaltracker.com/story/show/166852285


app/javascript/utils/form/final_form/Field.js
```diff
@@ -99,7 +99,7 @@ export const SelectField = ({ fieldName, label, blockNote, validation, id, optio
             }}
             id={id || `form-${fieldName}`}
             name={input.name}
-            className={classNames(className || 'form-group', (meta.error && meta.touched && 'error') || '')}>
+            className={classNames(className, (meta.error && meta.touched && 'error') || '')}>
```
akegan, line 102: this fixed row spacing issues from https://www.pivotaltracker.com/story/show/166909855

akegan:
@LoganArnett I'm wondering if you have any ideas for making the required block note show up on the gender/sexual orientation other dynamically if the user selects not listed. Seems like right now the component doesn't know anything about state (it's determined in [this file](https://github.com/Exygy/sf-dahlia-lap/blob/master/app/javascript/components/applications/application_form/PaperApplicationForm.js#L45)). It's probably not worth making any big changes to make it possible, but if there's an easy fix then maybe it is worth it

spec/javascript/components/applications/application_form/PaperApplicationForm.test.js
```diff
@@ -118,35 +118,24 @@ describe('PaperApplicationForm', () => {
           onSubmit={() => (null)}
         />
       )
-      wrapper.find('form').first().simulate('submit')
-
-      await wait(100)
-
-      expect(wrapper.text()).toContain('Ethnicity is required')
```
akegan, line 125: these pieces were already tested in unit test above. I removed them for clarity

spec/javascript/components/applications/__snapshots__/ApplicationsPage.test.js.snap
```diff
@@ -105,7 +105,7 @@ Array [
                 className="form-group"
               >
                 <select
-                  className="form-group"
+                  className=""
```
LoganArnett, line 108: If these no longer need classes we can remove here and line 169

spec/javascript/components/lease_ups/__snapshots__/LeaseUpApplicationsPage.test.js.snap
```diff
@@ -427,7 +427,7 @@ exports[`LeaseUpApplicationsPage Should render LeaseUpTable 1`] = `
                                           id="form-preference"
                                         />
                                         <select
-                                          className="form-group"
+                                          className=""
```
LoganArnett, line 430: Same className removal here

app/javascript/components/applications/application_form/PaperApplicationForm.js
```diff
@@ -79,8 +81,11 @@ class PaperApplicationForm extends React.Component {
     if (values.demographics) {
       const { demographics } = values
       errors.demographics = {}
-      errors.demographics.gender_other = this.checkNotListed(demographics, 'gender', 'Gender is required')
-      errors.demographics.sexual_orientation_other = this.checkNotListed(demographics, 'sexual_orientation', 'Sexual Orientation is required')
+
+      errors.demographics.gender_other = this.checkNotListed(demographics, 'gender') ? 'Gender is required' : undefined
+      this.state.genderSpecifyRequired = this.checkNotListed(demographics, 'gender')
```
LoganArnett: When setting state here we should still use `this.setState` instead of direct assignment. This is more stable with our components lifecycle methods. One easy way could be to do something like
``` js
const isGenderSpecifyRequired = this.checkNotListed(demographics, 'gender');
const isOrientationOtherRequired = this.checkNotListed(demographics, 'sexual_orientation');
errors.demographics.gender_other = isGenderSpecifyRequired ? 'Gender is required' : undefined
errors.demographics.sexual_orientation_other = isOrientationOtherRequired ? 'Sexual Orientation is required' : undefined
this.setState({ genderSpecifyRequired: isGenderSpecifyRequired, orientationOtherRequired: isOrientationOtherRequired })
```

LoganArnett (changes requested):

spec/javascript/components/lease_ups/__snapshots__/LeaseUpApplicationsPage.test.js.snap
akegan: this is just the snapshot, I tried setting it to null or undefined and it still shows up as a line

app/javascript/components/applications/application_form/PaperApplicationForm.js
akegan: update made!

LoganArnett (approved):

PR #274, opened 2019-07-16
Fix security vulnerabilities #166827461

chrisdolendo:
Issues:
* Fstream - getting downgraded by the latest axios
* Upgrading Lodash downgrades just-extend from 4.0 back to 3.0


akegan:
@chrisdolendo I think we're good to go with my grunt updates. We still have an issue with lodash version on applause (which is a grunt dep), but we've also moved those to dev dependencies so they're not installed on production. 

will you give this another once-over and maybe re-confirm that we're OK on the packages and then we can merge!

akegan (approved):

PR #276, opened 2019-07-25
Chores/speed up e2e tests 167210348

LoganArnett:
https://www.pivotaltracker.com/story/show/167210348

Updating e2e tests for speed updates with shared browser options and removing image and css from rendering

This PR is for [ticket #167210348](https://www.pivotaltracker.com/story/show/167210348)


LoganArnett:
@chrisdolendo this is ready for review, it looks like the shared sessions and removing CSS has shaved between 2-3 minutes off of the e2e portion

chrisdolendo:
Approaches implemented to speed up E2E:

- Shared sessions
- Removing CSS
- Combine more tests, e.g. have scenario fail 3 ways, then succeed rather than having separate tests for similar scenarios

spec/javascript/e2e/ApplicationNewPage.e2e.js
```diff
@@ -1,23 +1,35 @@
...
+import SetupBrowserAndPage from '../utils/SetupBrowserAndPage'
+let testBrowser
 
 describe('ApplicationNewPage', () => {
-  test('should create a new application successfully', async () => {
-    let browser = await puppeteer.launch({ headless: HEADLESS })
-    let page = await browser.newPage()
+  test('should fail if required fields are missing, and create application if not', async () => {
+    let { browser, page } = await SetupBrowserAndPage()
+    testBrowser = browser
```
akegan, line 16: would be feasible to have an explicit setup and teardown of the browser outside of the individual tests? Is that feasible?

akegan:
@LoganArnett were you up for adding an env var and default to the consts file so that it's easier to turn css on and off

spec/javascript/utils/SetupBrowserAndPage.js
```diff
@@ -0,0 +1,35 @@
+import puppeteer from 'puppeteer'
+import { HEADLESS as headless } from '../support/puppeteer/consts'
+
+const IgnoreImageAndCSSLoad = async (page) => {
+  await page.setRequestInterception(true)
+
+  page.on('request', (req) => {
+    if (req.resourceType() === 'stylesheet' || req.resourceType() === 'font' || req.resourceType() === 'image') {
```
akegan, line 8: just to clarify from this morning, do you have a sense of the effect of this piece vs the combination of tests and sharing of browsers between individual tests?

akegan (commented):
Just a couple of questions, would like to see the easy configurability of css on/off though

spec/javascript/e2e/ApplicationNewPage.e2e.js
LoganArnett: I am not sure, I will check to see if there is a way to handle that cleaner

spec/javascript/utils/SetupBrowserAndPage.js
LoganArnett: This piece appears to remove about 1 minute overall since the majority of the tests can have it be removed, the shared browser sessions removes about the same

spec/javascript/e2e/ApplicationNewPage.e2e.js
LoganArnett: I looked into a few options but I am not sure that our current setup supports it and it may take another decent chunk of time to refactor to a `beforeAll/beforeEach` approach with something like `jest-puppeteer`

LoganArnett:
@akegan updated with the constant here for individual override and mass override

spec/javascript/e2e/ApplicationNewPage.e2e.js
akegan: ok thanks for looking into it

spec/javascript/support/puppeteer/consts.js
```diff
@@ -41,3 +41,5 @@ export const DECLINE_TO_STATE = 'Decline to state'
 
 // Change to false to see tests running on browser locally
 export const HEADLESS = true
+// Change to false to fall back to including css
+export const REMOVECSS = true
```
akegan: @logan can you update to actually make it settable via env var too so we can set it on semaphore?

akegan (approved):
I think we're good to merge this, env var issue shouldn't come up too much because I anticipate mostly using the env var on semaphore, rather than using it in my local env file to flip the switch on css

PR #277, opened 2019-07-26
Fix status dropdown update on supp app page #167503966

chrisdolendo:
https://www.pivotaltracker.com/story/show/167503966


akegan:
change looks good. can you add a test for this? might be easy to add it to a snapshot test that we already have or something

akegan (changes requested):
see request for tests above

akegan (approved):
looks good! merge away!

PR #280, opened 2019-08-12
Set all fields to touched on status dropdown validation check #167854715

LoganArnett:
This PR adds in a touchAllFields helper

### DIFF
--- a/app/javascript/utils/form/validations.js
+++ b/app/javascript/utils/form/validations.js
@@ -125,4 +125,8 @@ validate.isValidDate = (date, errors, options = {}) => {
   return errors
 }
 
+export const touchAllFields = (form, touchedState) => {
+  Object.keys(touchedState).map(fieldName => form.blur(fieldName))
+}
+
 export default validate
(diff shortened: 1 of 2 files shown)


app/javascript/utils/form/validations.js
```diff
@@ -125,4 +125,8 @@ validate.isValidDate = (date, errors, options = {}) => {
   return errors
 }
 
+export const touchAllFields = (form, touchedState) => {
```
akegan, line 128: I'm not super clear on what touchedState is here

LoganArnett: `touchedState` is the current touched `state` from the form of all fields, so it is just an object with all of the field names and a boolean value, i.e. 

```js
{
  name: false,
  username: true,
  password: false
}
```

final-form keeps track of these at all times so on our validation calls we can simply grab that object from the form props and loop through the names and blur them ourselves with `form.blur`

akegan (approved):
did QA making sure that lease start date is now validating on save

PR #281, opened 2019-08-13
Chores/functional components refactor #167213234

LoganArnett:
This PR covers [ticket #167213234](https://www.pivotaltracker.com/story/show/167213234) for the functional component refactor


spec/javascript/support/puppeteer/consts.js
```diff
@@ -40,6 +40,6 @@ export const LENDING_AGENT_ID = process.env.E2E_LENDING_AGENT_ID || '0030P00002C
 export const DECLINE_TO_STATE = 'Decline to state'
 
 // Change to false to see tests running on browser locally
-export const HEADLESS = true
+export const HEADLESS = false
```
akegan: whoops this probably shouldn't have snuck in there
```suggestion
export const HEADLESS = true
```

akegan:
looks like there are a couple of merge conflicts and snapshots probably need to be rebuilt after merging your memory leak PR

akegan:
I marked the code climate issues as won't fix. Since we're not touching the functionality of the two files at all in this PR I don't think it makes sense to refactor them in this PR

LoganArnett:
Updated the merge conflicts and pushed

akegan:
unit test failing:
```
 console.error node_modules/react-dom/cjs/react-dom.development.js:506
      Warning: An update to ExpandableTableRow inside a test was not wrapped in act(...).
      
      When testing, code that causes React state updates should be wrapped into act(...):
      
      act(() => {
        /* fire events that update state */
      });
      /* assert on the output */
...
```

LoganArnett:
Yea the unit test failing came in after merging the latest branch, there is something funny going on with how that input is getting set,

akegan:
found an issue with missing flagged apps button, doesn't look like it's a change here, probably from the memory leak one. 
image
image

Options are:
1. I can fix it in the bug PR i started last night
2. You can fix it in this PR if you think that'll help avoid merge conflicts. 

I think it definitely deserves some updated testing for that functionality too. 

**Edit**: Nevermind, looks like it's not a bug, but that once lottery is complete we don't allow updates, so we hide the save button but the dropdown still shows up. I'll create a separate ticket to fix that.

app/javascript/components/molecules/TabsMenu.js
```diff
@@ -46,45 +46,41 @@ const Tab = ({ title, url, active, onKeyDown, onFocus, onClick, linkRefs }) => {
...
-  tabRefs = arrayUtils.cycle([])
+const TabsMenu = ({ items, currentUrl }) => {
+  const tabRefs = arrayUtils.cycle([])
 
-  handleKeyDown = (e) => {
+  const handleKeyDown = (e) => {
     keyboard.forEvent(e)
-      .on('leftArrow', (e) => this.tabRefs.prev().focus())
-      .on('rightArrow', (e) => this.tabRefs.next().focus())
+      .on('leftArrow', () => tabRefs.prev().focus())
```
akegan, line 54: not super clear on what the value of this is @callaghanc do you have any idea? Seems like if we tab through, then can use the arrow key to switch back and forth between highlighting the tabs? do you know if that was added as an accessibility item? or something else?

app/javascript/components/applications/application_form/EligibilitySection.js
```diff
@@ -1,161 +1,112 @@
...
-          id='lending_agent'
-          fieldName='lending_agent'
-          options={lenders}
-          validation={validate.isPresent('Please select a lender.')}
-        />
-      </div>
-    )
-  }
-
-  checkboxesMarkup = () => {
```
akegan, line 113: curious to hear what the motivation is around switching from this "markup" strategy to putting all the html in the return directly.

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -33,11 +33,7 @@ class LeaseUpApplicationsPage extends React.Component {
       loading: false
     }
   }
-
-  constructor (props) {
-    super(props)
-    this.eagerPagination = new EagerPagination(ROWS_PER_PAGE, SERVER_PAGE_SIZE)
-  }
+  eagerPagination = new EagerPagination(ROWS_PER_PAGE, SERVER_PAGE_SIZE)
```
akegan, line 36: does this need to have a const attached to it?

app/javascript/components/molecules/DropdownMenu.js
```diff
@@ -2,41 +2,37 @@ import React from 'react'
 import DropdownMenuItem from '../atoms/DropdownMenuItem'
 import formUtils from '~/utils/formUtils'
 
-class DropdownMenu extends React.Component {
-  handleOnChange = (e, value, label) => this.props.onChange && this.props.onChange(value, label)
+const DropdownMenu = ({ items, value, style, classes, onChange }) => {
+  const handleOnChange = (e, val, label) => onChange && onChange(val, label)
 
-  handleOnKeyDown = (e, value, label) => {
+  const handleOnKeyDown = (e, val, label) => {
```
akegan, line 8: I'm not super clear on what this functionality with keyboards was for dropdowns before, do you?

akegan (commented):
OK, I have officially reviewed all of the files. I left some scattered questions, but for the most part it looks good. 

 @LoganArnett once you solve the unit test issue I'll review those changes and do a final QA then we should be good to go.

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
LoganArnett: No this is technically still a class component so adding as a property of the class like this is the same as `this.eagerPagination`

app/javascript/components/applications/application_form/EligibilitySection.js
LoganArnett: This structure of calling multiple class functions directly like this is another performance issue. The other option would be to rip these out of the class and create small functional components and pass them as a component instead of a function. That way they are handled by react and the virtual dom in a more efficient way and can be loaded ahead of time instead of having these non component functions being re-called on each render

app/javascript/components/molecules/DropdownMenu.js
LoganArnett: These keycodes are `enter` and `space` so I assume that this was added so that you can select an item without clicking it with the mouse

app/javascript/components/molecules/TabsMenu.js
callaghanc: @akegan yes the idea is to allow the arrow keys to navigate through the tabs, as an accessibility compliance feature. here's the original PL story for creating the tabs: https://www.pivotaltracker.com/story/show/155753715. and here's the story where the tabs were implemented in partners: https://www.pivotaltracker.com/story/show/154785586. both stories link to these accessibility guidelines which explain the arrow key navigation: https://www.w3.org/TR/2017/NOTE-wai-aria-practices-1.1-20171214/examples/menubar/menubar-1/menubar-1.html#.

app/javascript/components/applications/application_form/EligibilitySection.js
akegan: ok, this change makes sense. thanks for the explanation!

akegan (approved):
Looks good to me. I did some thorough QA of all of the various bits of functionality.

Let's wait for @callaghanc to finish her review before merging though

app/javascript/components/SpreadsheetIndexTable.js
```diff
@@ -8,15 +8,12 @@ import IndexTableCell from './IndexTableCell'
...
-      expanded: {},
-      loading: {},
-      // [...array] == clone array
-      editData: [...props.results],
-      persistedData: [...props.results]
-    }
+  state = {
+    expanded: {},
+    loading: {},
+    // [...array] == clone array
```
callaghanc, line 14: @LoganArnett could we remove this comment?

callaghanc:
@akegan I connected with @LoganArnett to take a look at how the `useEffect` hook was working on the `Dropdown` component, and we noticed some odd behavior. the effect should only be happening on mount and unmount for each component, but on the lease up applications page we saw the effect firing more often than that (see attached screenshot). @LoganArnett is going to look into this and see if he can find out why it's happening. if it's a small fix, let's include it in this PR, but if it turns out that the issue predates the refactor and is something more complicated, maybe to do with how we're putting the lease up applications page together, then we can break out a new story for that and merge this PR.
Screen Shot 2019-08-22 at 9 32 31 AM

akegan:
@LoganArnett does look like this PR has merge conflicts. Let me know if you have any questions about what the changes are or how to resolve them

akegan:
OK I think this is ready to go! Just did a bit more smoke testing and checked out the changes you pulled into master and confirmed that AMI calculations are still working. @callaghanc did you want to give an official approving review to this too?

PR #282, opened 2019-08-16
Fix AMI calculation for supplemental application #167937039

akegan:
https://www.pivotaltracker.com/story/show/167937039

Still WIP. Need to add tests, and maybe refactor a bit.


app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -26,7 +26,8 @@ const getAmiCharts = (chartsToLoad) => {
 const getAmis = async (chartsToLoad) => {
   const promises = map(chartsToLoad, chart => getAMIAction({ chartType: chart.ami_chart_type, chartYear: chart.ami_chart_year }))
   const amis = await Promise.all(promises)
-  return amis
+  // convert list of ami lists into one big array
```
akegan: when amis is returned it's either
[[{}, {}, {}]] if there's only one kind of AMI chart, and [[{},{},{}], [{},{},{}], ...] if there are multiple AMI charts.

It might make sense to move this logic into where we actually find which AMI chart to use in the confirmed HH income file.

LoganArnett:
@akegan Everything appears fine code wise, but for the failing test I am not able to find a matching id for the element that it is trying to find

akegan:
looking into failing tests now

akegan:
@LoganArnett e2e test should be fixed. It actually caught a bug too!

LoganArnett:
Looks good to me @akegan

PR #288, opened 2019-09-05
Add styles and message for invalid preferences #168173940

software-project:
https://www.pivotaltracker.com/story/show/168173940


app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -86,7 +86,7 @@ const LeaseUpApplicationsTable = ({ listingId, dataSet, onLeaseUpStatusChange, o
     const trClassName = classNames(
       'rt-tr-status',
       statusClassName,
-      { 'is-invalid': rowInfo && isInvalid(rowInfo.original) }
+      { 'is-invalid is-disqualified': rowInfo && isInvalid(rowInfo.original) }
```
akegan: we might be able to drop the is-invalid piece? or do we need to update pattern lib to not make it greyed out?

software-project: We either need to update pattern lib or leave is-invalid in for grayed out text.

akegan: does is-invalid do anything other than grayed-out text? we actually want to get rid of the grayed out text piece

software-project: only that, will remove it then

akegan (dismissed):
@software-project looks like removing the is-invalid class also got rid of the "NRHP Invalid" piece with the roll-over behavior

image

Seems like we should maybe update the pattern lib to have is-invalid just change the background color and not gray out the letters?

software-project:
@akegan I suppose we need it then and styles update. Added it back.

akegan:
changes look good with pattern lib update, but I want to confirm the final style with @emilyshore before we merge.

akegan:
@callaghanc latest master of pattern lib is imported so this is ready for review!

callaghanc (approved):

PR #299, opened 2019-10-21
Add general pref applications to default lease up applications view #168008976

akegan:


akegan:
not working right now, only shows general apps if all prefs is selected.

app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -19,9 +14,20 @@ def index
         status_last_updated_dates = application_status_service.last_updated_dates(application_ids)
         set_status_last_updated(status_last_updated_dates, applications.records)
       end
+
       render json: applications
     end
 
+    # def lease_up_get_general_preference_applications
```
akegan: Don't forget to remove these

app/javascript/components/lease_ups/leaseUpAppPrefModel.js
```diff
@@ -16,8 +16,8 @@ export const buildLeaseUpAppPrefModel = (applicationPreference) => {
     lease_up_status: application.processing_status,
     status_last_updated: application.status_last_updated,
     preference_order: applicationPreference.preference_order,
-    preference_record_type: listingPreference.record_type_for_app_preferences,
-    preference_lottery_rank: applicationPreference.preference_lottery_rank,
+    preference_record_type: applicationPreference.preference_name.includes('General') ? applicationPreference.preference_name : listingPreference.record_type_for_app_preferences,
```
akegan: We should add a comment here to explain what we're doing since the behavior is weird.

app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
```diff
@@ -64,9 +64,9 @@ const LeaseUpTableContainer = ({
       rowData.address = result.residence_address
     }
 
-    rowData.preference_rank = `${result.preference_record_type} ${result.preference_lottery_rank}`
+    rowData.preference_rank = `${result.preference_record_type} ${result.preference_lottery_rank || result.preference_all_lottery_rank}`
```
akegan: have we already mapped it so this file doesn't have to know about the all vs not-all rank?

app/services/force/soql/preference_service.rb
```diff
@@ -50,6 +60,15 @@ def app_preferences_for_listing_query(opts)
                .transform_results { |results| massage(results) }
       end
 
+      def app_preferences_for_listing_single_query(opts)
```
akegan: we should add a comment to explain this too since it's pretty hacky. Let me know if you want me to write up the language from it.

akegan (commented):
Filters other than the prefs filter don't seem to be working at all with what's currently applied

akegan:
@LoganArnett here's a gif of the first name and app number filters not working for me
lease_up_filters

akegan:
I'm still having issues searching for names, it looks like searching for a name in general works, but not e.g. portia toole.

I also just got stuck in a weird state where most of the applications aren't showing even though I don't have any filters set.
image

Refreshing the page didn't seem to help either 🤔

akegan:
Actually maybe the filtering is working but the base query is just off so we're not getting all of the results back? I signed out and back in and am still getting that limited subset (14 records total)

app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
akegan: pinging again on this comment too

app/javascript/components/lease_ups/leaseUpAppPrefModel.js
akegan: pinging on this comment again too

app/javascript/components/lease_ups/leaseUpAppPrefModel.js
```diff
@@ -16,8 +16,8 @@ export const buildLeaseUpAppPrefModel = (applicationPreference) => {
     lease_up_status: application.processing_status,
     status_last_updated: application.status_last_updated,
     preference_order: applicationPreference.preference_order,
-    preference_record_type: listingPreference.record_type_for_app_preferences,
-    preference_lottery_rank: applicationPreference.preference_lottery_rank,
+    preference_record_type: applicationPreference.preference_name && applicationPreference.preference_name.includes('General') ? applicationPreference.preference_name : listingPreference.record_type_for_app_preferences,
```
akegan, line 20: we should add a comment here to explain too

app/javascript/components/lease_ups/leaseUpAppPrefModel.js
LoganArnett: Updated

app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
LoganArnett: removed

akegan:
OK. I think this is good to go! 😱. Because I created the PR in the first place I can't approve. but @LoganArnett when you get back you can approve and merge

LoganArnett (approved):

PR #300, opened 2019-10-22
updating with pattern library

LoganArnett:
This is for [ticket #168998006](https://www.pivotaltracker.com/story/show/168998006)

s


akegan (approved):
looks good

PR #301, opened 2019-11-02
Refresh all VCR Cassettes #168299150

akegan:
https://www.pivotaltracker.com/story/show/168299150

Cassette recreation is still a bit funny because:
1. Our applications we use as fixtures update over time and still don't 100% match from prod to full (and are therefore overwritten on refresh)
1. Lending institution memoization is weird
1. Some tests require e.g. starting with an application without a lease, then adding a lease

I will do some more thinking on how to make it easier to redo all cassettes on Monday


LoganArnett:
Are we supposed to be putting the value of SALESFORCE_INSTANCE_URL directly in the `yml` files?

LoganArnett (approved):
Everything looks good

PR #305, opened 2019-11-22
Chores/remove full page refresh #167213157

LoganArnett:
This PR is for ticket [#167213157](https://www.pivotaltracker.com/story/show/167213157)


app/controllers/api/v1/short_form_controller.rb
```diff
@@ -11,6 +11,11 @@ def submit
     if short_form_validator.valid?
       application = custom_api_application_service.submit(application_api_params)
       logger.debug "application submit response: #{application}"
+      if params[:supplemental]
```
akegan, line 15: let's add a comment here along the lines of: If we've submitted a supplemental application, fetch additional data from salesforce not included in the default custom API submit response.

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -163,7 +168,15 @@ def application_db_params
           )
   end
 
+  def application_query_params
```
akegan: where does this get used?

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -185,7 +186,7 @@ const SupplementalApplicationContainer = ({ store }) => {
             <div className='button-pager'>
               <div className='button-pager_row align-buttons-left primary inset-wide'>
                 <StatusDropdown
-                  status={application.processing_status}
+                  status={statusHistory && statusHistory[0] ? statusHistory[0].status : application.processing_status}
```
akegan: what was the motivation behind this change?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -149,9 +159,12 @@ class SupplementalApplicationPage extends React.Component {
       })
       this.setState({loading: false})
     } else {
-      this.updateStatusModal({loading: false, isOpen: false})
-      // Reload the page to fetch the field update comment just created.
-      window.location.reload()
+      console.log(setApplicationsDefaults(mapApplication(updatedApplication)))
```
```suggestion
```

app/javascript/components/supplemental_application/actions.js
```diff
@@ -6,14 +6,20 @@ import { isEmpty, find, isEqual, every } from 'lodash'
...
   const applicationApi = domainToApi.buildApplicationShape(application)
-  promises.push(apiService.submitApplication(applicationApi))
+  promises.push(apiService.submitApplication(applicationApi, true))
   const responses = await Promise.all(promises)
 
-  return every(responses, (promise) => promise !== false)
+  if (every(responses, (promise) => promise !== false)) {
+    const [ { lease }, { application } ] = responses
+    if (application) {
+      application.lease = lease
```
akegan, line 24: what's the motivation behind this logic?

spec/controllers/api/v1/leases_controller_spec.rb
```diff
@@ -22,7 +22,7 @@
       end
       expect(response).to have_http_status(:success)
       json = JSON.parse(response.body)
-      expect(json['lease']).to eq(true) # update returns true
+      expect(json['lease']).not_to eq(false)
```
akegan: I'd rather have this explicitly check for some of the params if it's supposed to be returning the lease now.

spec/controllers/api/v1/leases_controller_spec.rb
```diff
@@ -55,7 +55,7 @@
...
```
akegan: same here

spec/javascript/components/supplemental_application/actions.test.js
```diff
@@ -43,7 +43,7 @@ describe('updateApplication', () => {
     const applicationDomain = mapApplication(supplementalApplication)
     const response = await updateApplication(applicationDomain)
 
-    expect(response).toEqual(true)
+    expect(response).not.toEqual(false)
```
akegan: same here re: more explicit expectations, can we expect that it equals the input application or something. Same goes for other tests in this file

spec/javascript/e2e/SupplementalApplicationPage/lease_information.e2e.js
```diff
@@ -30,9 +30,6 @@ describe('SupplementalApplicationPage lease section', () => {
     // Click save
     await supplementalApplicationSteps.savePage(page)
 
-    // Wait for page to load
-    await page.waitForNavigation()
```
akegan, line 34: exciting!

akegan (changes requested):
It looks like test coverage is over 50% for this PR but it drops the overall test coverage, are you up for picking up a couple of lines that are showing up as not-covered and adding a test or two for them?

QA issue:
- I tried saving the app with an open rental assistance tab and the rental assistances disappeared 
image

app/javascript/components/supplemental_application/actions.js
LoganArnett: The responses come back as 200's no matter what, but they return either a value or false, so we check to see the responses are not just false like before and then return the application with information otherwise false

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
LoganArnett: removed

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
LoganArnett: Originally I misunderstood the refresh action, this comment helped to see what I was missing and had thought I accounted for in another way, I updated this to update the status directly to also get that back on the pull

app/controllers/api/v1/short_form_controller.rb
LoganArnett: removed

app/controllers/api/v1/short_form_controller.rb
LoganArnett: added

spec/controllers/api/v1/leases_controller_spec.rb
LoganArnett: updated

spec/controllers/api/v1/leases_controller_spec.rb
LoganArnett: updated

spec/javascript/components/supplemental_application/actions.test.js
LoganArnett: updated

LoganArnett:
replicated and realized the rental assistances had not been added to the fetch on update so I added that in

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -11,8 +11,10 @@ def submit
     if short_form_validator.valid?
       application = custom_api_application_service.submit(application_api_params)
       logger.debug "application submit response: #{application}"
+      # if submitting a supplemental application we will re-fetch updated data from salesforce
       if params[:supplemental]
         application = soql_application_service.application(application[:id] || application_api_params[:id])
+        application['rental_assistances'] = soql_rental_assistance_service.application_rental_assistances(application[:id] || application_api_params[:id])
```
akegan: is the lease pulled in in the application above?

akegan: or it looks like it comes from a separate request?

akegan: looks like in the supplementals controller we add the rental assistance into the application object the same way you're adding it here, so that's OK for now. Ideally I'd like to keep the application object as just the application object and treat rental assistances and leases separately, or at least group rental assistances and leases together since rental assistances are actually tied to a lease

akegan: actually, can we just go ahead and at least move the rental assistances piece into the applications service for now so it's treated the same way as leases? pulling it out of here and the supplementals controller and putting it in the soql applications controller instead? would that break anything?

akegan:
A couple of issues with saving rental assistances with an open panel:
1. If successful, the rental assistance shows up in the table, but a new panel is then opened that is blank. The panel should close on successful save.
1. I had an intermittent issue were one of my rental assistances saved via an open panel didn't show up in the table when the save appeared to be complete. It then later showed up when I was saving another rental assistance with an open panel

akegan:
I just ran into an issue where after save the lease wasn't there either, it did show up when I refreshed the page.

app/controllers/api/v1/rental_assistances_controller.rb
```diff
@@ -24,7 +24,8 @@ def update
         )
 
         if response
-          render json: true
+          updated_rental_assistances = soql_rental_assistance_service.application_rental_assistances(params[:application_id])
```
akegan: one small nit is that I've been updating other similar methods to use the bang (!) methods to actually raise an error that we can see in the logs instead of having this 422 status response. That way if something goes wrong, the user experience isn't actually that different, and I can see the logs in papertrail

app/javascript/components/molecules/ExpandableTable.js
```diff
@@ -30,7 +36,7 @@ const ExpandableTableRow = ({ row, rowKeyIndex, numColumns, expanderRenderer, ex
   )
 }
 
-const ExpandableTable = ({ columns, rows, rowKeyIndex, expanderRenderer, expandedRowRenderer, originals, classes }) => (
+const ExpandableTable = ({ columns, rows, rowKeyIndex, expanderRenderer, expandedRowRenderer, originals, classes, toggleAllRows }) => (
```
akegan: what does toggle all rows mean here? should it be "closeAllRows"?

app/javascript/components/supplemental_application/actions.js
```diff
@@ -1,32 +1,42 @@
...
-  const responses = await Promise.all(promises)
 
-  if (every(responses, (promise) => promise !== false)) {
-    const [ { lease }, { application } ] = responses
+  const initialResponses = await Promise.all([
+    updateLease(application['lease'], primaryApplicantContact, application['id']),
+    apiService.submitApplication(applicationApi, true)
+  ])
+
+  const rentalAssistanceResponses = await Promise.all(updateUnsavedRentalAssistances(application, prevApplication))
```
akegan: so here we're saying, first we're going to update the rental assistances, wait for a response, then update the lease and application? Due to potential collisions with creating a lease through updating a lease and saving new rental assistances without a lease already on the application and the same time?

If that's the case, can we add a comment explaining that that's why we're keeping them separate? Otherwise I think it's weird enough that it will be really confusing when we end up back in this function again.

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -250,7 +249,7 @@ const RentalAssistanceForm = ({ values, onSave, loading, onClose, applicationMem
   )
 }
 
-const RentalAssistance = ({ store, form }) => {
+const RentalAssistance = ({ store, form, submitting }) => {
```
akegan, line 252: I'm not super clear on what this submitting variable is here. I was kinda expecting it to be like a loading state, but now I'm just a bit confused.

spec/controllers/api/v1/short_form_controller_spec.rb
```diff
@@ -6,45 +6,55 @@
...
       end
     end
+    describe 'with supplemental param' do
+      it 'receives a successful response with updated application from Salesforce' do
+        VCR.use_cassette('api/v1/short-form/submit/with-supplemental-param') do
+          post :submit, params: { application: base_application, supplemental: true }
+        end
+        expect(response).to have_http_status(:success)
+        json = JSON.parse(response.body)
+        # puts json['application']
```
akegan: need to remove this

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -138,14 +139,6 @@ describe('SupplementalApplicationPage', () => {
 
   test('it saves a live/work application preference panel', async () => {
     const withValidPreferences = cloneDeep(supplementalApplication)
-    const payload = cloneDeep(mockShortFormSubmitPayload())
```
akegan: what was the motivation behind removing this piece?

akegan (commented):
A few more questions and a couple of issues. Sorry this is taking so long!

I'll do some QA now to see check if the functionality is working.

akegan:
Found a QA issue while adding a new rental assistance on portia tool. I:

1. Loaded the application
2. Started a new rental assistance, partially filled it out, but did not hit save on the rental assistance
3. Clicked save at the bottom of the page. 

I got this issue:
image

akegan:
Another issue:
I just tried to add a lease and a rental assistance at the same time to this application that didn't have either https://dahlia-lap-full-pr-305.herokuapp.com/applications/a0o0P00000ItllUQAR/supplementals

Neither of the details showed up once the button stopped loading. They did show up when I refreshed the page. You might be able to reproduce on e.g. https://dahlia-lap-full-pr-305.herokuapp.com/applications/a0o0P00000ItlleQAB/supplementals

app/javascript/components/supplemental_application/sections/RentalAssistance.js
LoganArnett: This is being passed down from the main form, it is to let the other pieces know that the form is currently `submitting` https://final-form.org/docs/final-form/types/FormState#submitting, it is an trigger already managed by the form state instead of adding our own hook or state

app/javascript/components/molecules/ExpandableTable.js
LoganArnett: updated to closeAllRows, will be pushed with updates shortly, also applies to a comment below

app/javascript/components/supplemental_application/actions.js
LoganArnett: Currently we await the lease and application updates first, then update the rentalAssistances

This was based on our discussion where Lease and Application would take priority, I can add a comment to explain the order but I am still not familiar enough with the back end data structures and how they relate in terms of potential collisions other than what we have discussed

spec/controllers/api/v1/short_form_controller_spec.rb
LoganArnett: removed

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
LoganArnett: that payload was never used in any of the expects, we created and merged it below but we do not use it anywhere else in the test

akegan (approved):
looks like it's working now! let's merge once tests pass! 🎉. Looks like e2e tests are failing, but on pre-lottery stuff

akegan (approved):
really actually works now. Merging it now, good work!

PR #306, opened 2019-11-22
adding in use effect and page tracking from onFetch

LoganArnett:
This PR is for ticket [#169606899](https://www.pivotaltracker.com/story/show/169606899)

s

### DIFF
--- a/app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
+++ b/app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
@@ -1,4 +1,4 @@
-import React from 'react'
+import React, { useEffect, useState } from 'react'
 import ReactTable from 'react-table'
 import { trim } from 'lodash'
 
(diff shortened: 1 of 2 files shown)


app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -107,14 +113,18 @@ const LeaseUpApplicationsTable = ({ listingId, dataSet, onLeaseUpStatusChange, o
...
       pages={pages}
       columns={columns}
       getTdProps={getTdProps}
       getTrProps={getTrProps}
       defaultPageSize={rowsPerPage}
       sortable={false}
       loading={loading}
-      onFetchData={onFetchData}
+      onFetchData={(state, instance) => {
+        setCurrentPage(state.page)
```
akegan, line 125: can you explain a bit what's happening here? ReactTable has its own page in its state, right, so here we're saying, to update the react table's state to have the current page?

I found this answer for our issue, is that where you got the idea? https://spectrum.chat/react-table/general/pagination-reset-on-data-changed~2ef35446-d56a-4098-ac11-3e344f3af085

LoganArnett: Similar solution yes, I didn't actually see this prior, I read the SSR docs for the `manual` flag that we have set. We keep track of the page whenever data changes in the `onFetchData` call so we can set it by default with `useState` and then track it in state directly when the onFetch function fires

LoganArnett:
@akegan the update you made earlier allows this PR to pass

akegan (approved):
cool, looks good

PR #308, opened 2019-12-02
Feature/remove select one #169929680

LoganArnett:
This is for ticket [#169929680](https://www.pivotaltracker.com/story/show/169929680)


app/javascript/utils/formUtils.js
```diff
@@ -49,6 +49,11 @@ export const maxLengthMap = {
   income: 16
 }
 
+export const reorderEmptyOptions = (options) => {
+  options.some((option, i) => isEmpty(option.value) && options.unshift(option) && options.splice(i + 1, 1))
```
LoganArnett: Loop through and find an empty value, unshift it to the front, and splice the old index out

akegan (commented):
QA'd and looked at code and looks good. Only concern is that it looks like snapshot changes from the react-final-form update bled into this PR.

Semaphore is also failing because of the update to a higher version of ruby in the other PR. We can get that one in first.

akegan (approved):
approving because I think this'll be good to go once semaphore passes, please just keep an eye on the snapshots and make sure the changes make sense

akegan:
@LoganArnett this is good to merge!

PR #312, opened 2019-12-11
Add clearer error handling to confirmed preference save issue #170144565

akegan:
This PR also updates the way that we send preference updates to only include the total monthly rent if the preference is rent burdened

### DIFF
--- a/app/controllers/api/v1/applications_controller.rb
+++ b/app/controllers/api/v1/applications_controller.rb
@@ -16,6 +16,7 @@ module Api
         response = rest_application_service.update(application_params.merge(id: params[:id]))
         if response
           render json: true
+        end
       end
 
       private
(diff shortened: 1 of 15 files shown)


app/controllers/api/v1/applications_controller.rb
```diff
@@ -16,8 +16,6 @@ def update
         response = rest_application_service.update(application_params.merge(id: params[:id]))
         if response
           render json: true
-        else
```
akegan, line 19: this is handled already by rails, and this way we provide a bit more info to the console.logs on the front end

akegan:
still a bit of a work in progress, just need to add some rails tests

akegan:
@LoganArnett this PR is ready for review

LoganArnett (approved):
Working as expected 👍

PR #314, opened 2019-12-26
Only show supp app from lease up view #170198513

akegan:
https://www.pivotaltracker.com/story/show/170198513

This PR also:
- Consolidated the two listing services
- Fixed "act" error messages in tests
- Hides listings that are showing in the lease ups page from the pre-lottery pages. 
- Removes unused developer caching.


app/services/force/soql/listing_service.rb
```diff
@@ -14,6 +14,54 @@ def lease_up_listings
           WHERE Status__c = 'Lease Up' AND Is_Applicant_List_for_Leaseup__c = FALSE
         )))
       end
+
+      def pre_lottery_listings
+        # Show only listings that are not present in the lease up tab.
+        parsed_index_query(%(
+          SELECT #{query_fields(:index)} FROM Listing__c
+          WHERE Status__c != 'Lease Up' OR Is_Applicant_List_for_Leaseup__c = TRUE
```
akegan, line 22: this is the only change here, the rest of it is copied over from the other listing service file

spec/vcr/applications_controller/edit/non_lease_up_application.yml
```diff
@@ -21,7 +21,7 @@ http_interactions:
       message: OK
     headers:
       Date:
-      - Fri, 01 Nov 2019 21:38:48 GMT
```
akegan: all of these cassette changes are because moving the listing service queries to the other file ended up adding more spaces to the urls. I thought about trying to figure out a way to ignore it, but there wasn't one that was immediately apparent so I updated the cassettes instead.

app/javascript/components/layouts/CardLayout.js
```diff
@@ -12,7 +12,6 @@ const CardLayout = ({ children, pageHeader, tabSection, toolbar }) => {
     document.body.classList.add('bg-snow')
   }, [])
 
-  if (tabSection && !tabSection.currentUrl) { tabSection.currentUrl = window.location.pathname }
```
LoganArnett, line 15: What did this line accomplish or not accomplish? Just trying to clarify

akegan: this was selecting the active tab based on the current url, so for example if you're on /supplementals  have the supp app tab highlighted as active, and if you're on the plain application page, have the application page highlighted.

I had to update it because I added a param to the application url that has ?leaseup=true on it, so this line wasn't working. I had a different line of code that had the full url, then split it on the hostname, but I thought it made more sense to just explicitly set which tab was active wherever you're specifying the tabs.

LoganArnett (approved):

PR #317, opened 2020-01-13
Add yes-no parking dropdown to supp app #169077844

LoganArnett:
This PR is for ticket [#169077844](https://www.pivotaltracker.com/story/show/169077844)


akegan:
@LoganArnett I just tested and it looks good except we don't need the "Select One..." option.

app/javascript/apiService.js
```diff
@@ -102,11 +102,9 @@ const deleteRentalAssistance = async (rentalAssistanceId) => {
   return request.destroy(`/rental-assistances/${rentalAssistanceId}`)
 }
 
-export const createOrUpdateLease = async (lease, applicationId) => {
-  const leaseId = lease['id']
-  const data = {
-    lease: lease
-  }
+export const createOrUpdateLease = async (leaseToCreateOrUpdate, applicationId) => {
```
akegan, line 105: this looks like changes that bled over from your previous PR?

app/controllers/api/v1/rental_assistances_controller.rb
```diff
@@ -34,7 +34,8 @@ def update
         )
 
         if response
-          render json: true
+          updated_rental_assistances = soql_rental_assistance_service.application_rental_assistances(params[:application_id])
```
akegan: this also looks like changes from your other PR

app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
```diff
@@ -0,0 +1,53 @@
...
+
+import FormGrid from '~/components/molecules/FormGrid'
+import { SelectField, CurrencyField } from '~/utils/form/final_form/Field.js'
+import validate from '~/utils/form/validations'
+
+const monthlyRentFieldName = 'lease.monthly_parking_rent'
+const Yes = 'Yes'
+const No = 'No'
+
+const validateLeaseCurrency = (value) => {
```
akegan: this is duplicated from the other file. I wonder if it can be moved from to the validations file, or if there's another way to avoid this code duplication

app/javascript/utils/form/final_form/Field.js
```diff
@@ -106,6 +107,7 @@ export const SelectField = ({ fieldName, label, blockNote, validation, id, optio
               id={id || `form-${fieldName}`}
               name={input.name}
               className={classNames(className, (meta.error && meta.touched && 'error') || '')}
+              {...(selectValue && { value: selectValue })}
```
akegan, line 110: did we not have the option before of having a default value for a select field?

akegan (changes requested):
One comment about duplicate code, plus a couple of changes that look like they bled over from your refresh PR.

Other than these + the "select one", looks good.

app/javascript/apiService.js
LoganArnett: Not sure what you mean here, I updated this so that the variable `lease` was not being used so many times for potential side effects

akegan: ok thanks for the clarification! i wasn't sure if it was a rebase issue or something new.

app/javascript/utils/form/final_form/Field.js
LoganArnett: I did not see this in place, at least in the way React recommends so I added this

app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
LoganArnett: Brought this out into a util for both files

app/controllers/api/v1/rental_assistances_controller.rb
LoganArnett: removed

app/controllers/api/v1/rental_assistances_controller.rb
```diff
@@ -34,8 +34,7 @@ def update
         )
 
         if response
-          updated_rental_assistances = soql_rental_assistance_service.application_rental_assistances(params[:application_id])
```
akegan: Do you remember if this was something that was supposed to be in or out? Seems helpful just wasn't clear on the motivation

app/controllers/api/v1/rental_assistances_controller.rb
akegan: wondering which state is the right state to be in. Do you remember which it was supposed to be from the end of your last PR?

akegan (commented):
Select One... is still there, plus one comment below. Other changes look good

akegan (approved):
QA looks good. Yarn unit tests are failing, but you can merge once you get them passing, or I can fix them later tonight.

Another unintended consequence is that this will now create a lease on save just based on that default value, but I think that we'll just have to deal with that when we're a bit more explicit about creating or deleting a lease. For now it's OK

PR #319, opened 2020-01-22
Update first and last name queries with like syntax on lease ups - #169929870

LoganArnett:
This PR is for ticket number [#169929870](https://www.pivotaltracker.com/story/show/169929870)

### Example
s

### Performance Before
s

### Performance After
s


akegan (approved):
Looks good to me. looks like unrelated e2e test failed so I clicked re-run on them. you should be good to merge once those tests pass

akegan:
One other small request, can you update the PR title to use the present tense rather than present participle? (e.g. *Update* name query to use like syntax for lease ups) vs. updat*ing*). Same goes for your other PR.

PR #321, opened 2020-01-28
Format Currencies and Dates #169613899

LoganArnett:
This PR is for [ticket #169613899](https://www.pivotaltracker.com/story/show/169613899)


app/javascript/components/mappers/domainToApi/application.js
```diff
@@ -31,12 +30,12 @@ export const applicationFieldMapper = {
   reserved_senior: 'hasSenior',
   application_submission_type: 'applicationSubmissionType',
   processing_status: 'processingStatus',
-  householdAssets: (source) => currencyToFloat(source.household_assets),
```
akegan, line 34: I'm pretty sure that with these you can just actually flip it to match e.g. line 33 above, no need for the lambda fn

app/javascript/components/mappers/domainToApi/lease.js
```diff
@@ -8,9 +7,9 @@ export const leaseFieldMapper = {
   'no_preference_used': 'no_preference_used',
   'preference_used': 'preference_used',
   'unit': 'unit',
-  total_monthly_rent_without_parking: (source) => currencyToFloat(source.total_monthly_rent_without_parking),
```
akegan, line 11: same here re: flipping the order to avoid the lambda fn

app/javascript/components/mappers/domainToApi/lease.js
```diff
@@ -1,5 +1,4 @@
 import { createFieldMapper } from '~/utils/objectUtils'
-import { currencyToFloat } from '~/utils/utils'
```
akegan, line 2: is this still in use anywhere? If not can we delete it?

app/javascript/utils/form/validations.js
```diff
@@ -136,4 +138,15 @@ export const validateLeaseCurrency = (value) => {
   )
 }
 
+// format currency in one call instead of in different mappers
+export const convertCurrency = (values) => {
```
akegan, line 148: please add some unit tests for this function. We should have some on the old currency conversion function that you can repurpose.

app/javascript/utils/form/final_form/Field.js
```diff
@@ -10,7 +10,12 @@ const {
 
 // Make react-final-form does include empty values on submit.
 // Source: https://github.com/final-form/react-final-form/issues/130#issuecomment-425482365
-const identity = value => (value)
+const identity = (value, name, isCurrency = false) => {
+  if (value && isCurrency) {
+    return value.toString().replace(/[^\d|.]/g, '')
```
akegan, line 17: can you clarify what this is doing here? I think it's worth adding a comment too

app/javascript/utils/form/final_form/Field.js
```diff
@@ -65,25 +70,22 @@ export const InputField = ({ fieldName, label, blockNote, validation, placeholde
   </Field>
 )
 
-export const CurrencyField = ({ fieldName, validation, id, label, placeholder, maxLength, disabled }) => (
-  <Field name={fieldName} validate={validation} component='input' format={formUtils.formatPrice} parse={identity} formatOnBlur>
+export const CurrencyField = ({ fieldName, validation, id, label, placeholder, maxLength, disabled, isDirty = true }) => (
+  <Field name={fieldName} validate={validation} component='input' format={formUtils.formatPrice} parse={(value, name) => identity(value, name, true)} formatOnBlur={isDirty}>
```
akegan, line 76: Can you think of a way to make the identity fn a little clearer here rather than just passing "true" as the 3rd param? I'm not sure I can but I'm just curious

akegan (changes requested):
We have some currency conversion going on in mapping in ruby too. If you think this might be a better solution than that we'll have to get rid of those conversions too. 

The experience of entering the currencies is pretty different. I'm going to check in with Chris to see if we want to move forward with this updated experience or if we'd rather keep the old one. 

# QA Notes
Submitting a paper application failed with the error `APEX_ERROR: System.JSONException: Character $ is neither a decimal digit number, decimal point, nor "e" notation exponential mark. at [line:1, column:131]`

Submitting a lease up application does seem to work.

app/javascript/utils/form/validations.js
LoganArnett: Added a test around this to illustrate

app/javascript/utils/form/final_form/Field.js
LoganArnett: added a comment

app/javascript/utils/form/final_form/Field.js
LoganArnett: unfortunately I cannot see another way except for passing a string and checking that value but the boolean is more performant

app/javascript/components/mappers/domainToApi/lease.js
LoganArnett: removed and removed test

app/javascript/components/mappers/domainToApi/application.js
LoganArnett: updated

app/javascript/components/mappers/domainToApi/lease.js
LoganArnett: updated

app/javascript/components/mappers/domainToApi/lease.js
```diff
@@ -7,10 +7,10 @@ export const leaseFieldMapper = {
   'no_preference_used': 'no_preference_used',
   'preference_used': 'preference_used',
   'unit': 'unit',
-  total_monthly_rent_without_parking: (source) => source.total_monthly_rent_without_parking,
```
akegan: looking at this more it looks like we could maybe get rid of this lease mapping completely? Do you thin that would break anything?

app/javascript/utils/form/final_form/Field.js
```diff
@@ -12,6 +12,8 @@ const {
 // Source: https://github.com/final-form/react-final-form/issues/130#issuecomment-425482365
 const identity = (value, name, isCurrency = false) => {
   if (value && isCurrency) {
+    // Here we replace anyting that is not a numeric digit
```
akegan:
```suggestion
    // Here we remove anything that is not a numeric digit
```

app/javascript/utils/fieldSpecs.js
```diff
@@ -2,9 +2,15 @@ import { each, replace, get, toLower, includes, isString, camelCase, startCase,
 import moment from 'moment'
 
 import utils from '~/utils/utils'
+import formUtils from '~/utils/formUtils'
 
 export const getFormatType = (field) => {
-  if (includes(toLower(field), 'date')) { return 'date' } else { return null }
+  if (includes(toLower(field), 'date')) {
+    return 'date'
```
akegan, line 9: this seems a bit sketchy to me. I'm not sure what the new parking price PR used for a field name but I don't think that it has rent or income

akegan (commented):
Couple of questions, plus saving rental assistances is borked. I get  "JSON_PARSER_ERROR: Cannot deserialize instance of currency from VALUE_STRING value $111.00 or request may be missing a" when I try to save a rental assistance 

Rental assistances table also doesn't have values formatted as a currency:
image

app/javascript/utils/fieldSpecs.js
LoganArnett: Yea I was following a similar helper but looking at better options

app/javascript/components/mappers/domainToApi/lease.js
LoganArnett: removed and working

app/javascript/utils/fieldSpecs.js
LoganArnett: updated to what we discussed

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -54,7 +54,7 @@ class SupplementalApplicationPage extends React.Component {
 
   handleSaveApplication = async (application) => {
     this.setLoading(true)
-
+    console.log(application)
```
```suggestion
```

app/javascript/utils/fieldSpecs.js
```diff
@@ -26,6 +24,7 @@ const dateIsJson = (date) => {
 }
 
 export const formatValue = (value, type) => {
+  console.log(value, type)
```
akegan:
```suggestion
```

app/javascript/utils/fieldSpecs.js
```diff
@@ -26,6 +24,7 @@ const dateIsJson = (date) => {
...
   if (type === 'date') {
```
akegan: wondering why you added the currency format in the [expandable table](https://github.com/Exygy/sf-dahlia-lap/pull/321/files/d4c085152d5e9836f818b3abdea54d281f4e7bdd..0b943dbbf1c6a365a1cfb5a86215c6dd99fb90cf#diff-618b03bb9c05415bd74152c0747eb2cdR19) rather than here. Not a big deal. just curious because it looks like dates are formatted here.

akegan (commented):
Couple of console.logs to remove and one question about that last change. Starting another round of QA now.

app/javascript/utils/fieldSpecs.js
LoganArnett: It looks like these are not formatted using this same fieldSpecs helpers, at least not the rental_assistances, they are built inline right before declaring the table and bypass the other formatting

akegan:
# QA Results
## Currency formatting on views
Listing view is missing currency formatting on Deposit Min/Max. See e.g. [Edwin Lee Apartments](https://dahlia-lap-full-pr-321.herokuapp.com/listings/a0W0P00000HbryHUAR)
image

## Supp app
Rent burdened confirmed pref needs same update rental assistance had for saving preferences individually
image


Otherwise looks good!

akegan (approved):
OK looks good! merge away!

PR #322, opened 2020-01-29
Switch from pricing to qualifying ami #170706912

akegan:
We don't appear to actually be using this field for anything, but we need to keep it up to date with changes on Salesforce so changes don't break. 

We'll wait to deploy this until after the salesforce changes are deployed to prod

[Pivotal Ticket](https://www.pivotaltracker.com/story/show/170706912)


akegan:
quick update, won't be able to get tests to pass on this until Larry's change is deployed to full so the field is available to update VCRs

LoganArnett (approved):
LGTM 👍

PR #325, opened 2020-02-13
[Hotfix] Search for user to update based on uid not email #171262865

akegan:
[Pivotal Ticket](https://www.pivotaltracker.com/story/show/171262865)

Please help me think of ways that this might be a security concern and why we were searching based on uid and username in the first place.


chrisdolendo (approved):
Tested with review instructions and investigated potential security issues. Looks good! :+1:

PR #327, opened 2020-02-19
Filter by accessibility #169193985

LoganArnett:
This PR is for ticket [#169193985](https://www.pivotaltracker.com/story/show/169193985)

s


app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -4,6 +4,7 @@ module Api::V1
   # Lease Up Applications controller for access via the API
   class LeaseUpApplicationsController < ApiController
     def index
+      puts "Hello #{lease_up_apps_params}"
```
akegan:
```suggestion
```

app/javascript/components/lease_ups/LeaseUpApplicationsFilter.js
```diff
@@ -39,6 +44,9 @@ const LeaseUpApplicationsFilter = ({ onSubmit, preferences = [], loading = false
                 <div className='filter-group_item'>
                   <SelectField fieldName='status' options={StatusOptions} placeholder='Status' />
                 </div>
+                <div className='filter-group_item'>
+                  <SelectField fieldName='accessibility' options={accessibilitFilterOptions} placeholder='Accessibility Requests' />
```
akegan: Typo in accessibilityOptions both above and here
```suggestion
                  <SelectField fieldName='accessibility' options={accessibilityFilterOptions} placeholder='Accessibility Requests' />
```

app/services/force/soql/application_service.rb
```diff
@@ -10,7 +10,7 @@ class ApplicationService < Force::Base
 
       def applications(opts = { page: 0 })
         query_scope = applications_query(opts)
-
+        puts "HELLO #{opts}"
```
akegan:
```suggestion
```

app/services/force/soql/preference_service.rb
```diff
@@ -19,7 +19,7 @@ def app_preferences_for_application(application_id)
 
       def app_preferences_for_listing(opts)
         query_scope = app_preferences_for_listing_query(opts)
-
+        puts query_scope
```
akegan:
```suggestion
```

akegan (changes requested):
A couple of other issues from looking at the gif:
1. Need to replace address column with "Accessibility Requests" column with values
1. I think we need to combine hearing and vision into one option for now until we have a multi-select option. This is because many units are for folks with hearing or vision impairments. This likely complicates the query, but hopefull isn't too hacky. If doing a multiselect would be easier that's an option too.

app/services/force/soql/preference_service.rb
```diff
@@ -19,20 +19,25 @@ def app_preferences_for_application(application_id)
...
-        puts query_scope
+
         query_scope.query
       end
 
       private
 
       def buildAppPreferencesFilters(opts)
+        if opts[:accessibility]
+          accessibility_filters = opts[:accessibility].split(', ')
```
akegan, line 31: I'd like a test for this piece to make it clearer what the expected input and output are if possible. Maybe break it into a separate function and add a unit test? 
A comment might be helpful here too.

app/services/force/soql/preference_service.rb
```diff
@@ -26,12 +26,18 @@ def app_preferences_for_listing(opts)
...
+          accessibility_string = accessibility_filters.length == 1 ? "'#{accessibility_filters[0]}'" : "'#{accessibility_filters[0]}', '#{accessibility_filters[1]}'"
+        end
+
         filters = ''
         filters += "and Preference_All_Name__c like '%#{opts[:preference]}' " if opts[:preference].present?
         filters += "and Application__r.Name like '%#{opts[:application_number]}%' " if opts[:application_number].present?
         filters += "and Application__r.Applicant__r.First_Name__c like '%#{opts[:first_name]}%' " if opts[:first_name].present?
         filters += "and Application__r.Applicant__r.Last_Name__c like '%#{opts[:last_name]}%' " if opts[:last_name].present?
         filters += "and Application__r.Processing_Status__c = " + (opts[:status] == 'No Status' ? 'NULL' : "'#{opts[:status]}'") if opts[:status].present?
+        filters += "and Application__r.Has_ADA_Priorities_Selected__c INCLUDES (#{accessibility_string}) " if opts[:accessibility].present?
```
akegan, line 41: I'm not super clear on how this gives us vision OR hearing impairments, is it just some sort of SOQL magic?

LoganArnett: It is a query to say look for any of these that include vision impairments OR hearing impairments, so it checks against the string on each one individually

https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_querying_multiselect_picklists.htm

LoganArnett: Your logic thought is correct, that is the way I thought of it in SQL but yes using SOQL logic

akegan: OK, that makes sense, sort of! can you add that the comma means or in your comment about coming up with the accessibility_string above?

LoganArnett: added comment and unit test, was trying to add an FE test as well but was not successful

akegan (approved):
OK looks good to merge!

PR #331, opened 2020-02-28
Features lease up default listings #171143212

chrisdolendo:


akegan:
@chrisdolendo rspec tests are failing on this branch. Snapshot tests might break because of the changes too, so make sure you check those too

akegan:
otherwise, functionality looks good

spec/controllers/api/v1/short_form_controller_spec.rb
```diff
@@ -5,7 +5,7 @@
 RSpec.describe Api::V1::ShortFormController, type: :controller do
   login_admin
 
-  listing_id = 'a0W0P00000GbyuQ' # LAP Test Listing - Yellow Acres
+  listing_id = 'a0W0P00000F8YG4UAN' # LAP Test Listing - Yellow Acres
```
akegan: looks like this is the wrong listing. Should be yellow acres lease-up listing, but you've updated it to automated test listing, our pre-lottery test listing

akegan:
Update: Larry added a validation that blocks new applications from being submitted if the lottery is in progress or complete. In this case we're trying to add new applications to the lease up test listing. I'll refactor the tests to take that into account

akegan:
I went ahead and switched over to using automated test listing after all, and just made it a bit more explicit in the tests. i should be touching these spec tests in the mapping PR that I'm working on too, so I'll try to make the tests a bit more robust then.

akegan (approved):

software-project (approved):

PR #335, opened 2020-03-05
Add rake task for renaming heroku apps #171070561 

software-project:


Gemfile
```diff
@@ -33,6 +33,9 @@ gem 'jbuilder', '~> 2.5'
 # Use Capistrano for deployment
 # gem 'capistrano-rails', group: :development
 
+# Heroku API
+gem 'platform-api', github: 'envoy/platform-api', branch: 'review_apps'
```
akegan, line 37: can this be just in the development group? or does it need to be in the production environment?

lib/tasks/heroku.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'update_app_name'
+
+namespace :heroku do
+  desc 'Renames all review apps to kepp app name convension dahlia-lap-full-pr-XXX'
+  task rename_apps: :environment do
+    UdpateAppName.call
```
akegan:
```suggestion
    UpdateAppName.call
```

lib/update_app_name.rb
```diff
@@ -0,0 +1,58 @@
+# frozen_string_literal: true
+
+# Service for updating Heroku review app's name, to keep app name convension: 'dahlia-lap-full-pr-XXX'.
+class UdpateAppName
```
akegan:
```suggestion
class UpdateAppName
```

lib/tasks/heroku.rake
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+require 'update_app_name'
+
+namespace :heroku do
+  desc 'Renames all review apps to kepp app name convension dahlia-lap-full-pr-XXX'
```
akegan:
```suggestion
  desc 'Renames all review apps to keep app name convention dahlia-lap-full-pr-XXX'
```

lib/update_app_name.rb
```diff
@@ -0,0 +1,58 @@
...
+  SKIPPED_APPS = %w[dahlia-lap-full dahlia-lap-nonepic dahlia-lap-qa dahlia-lap-production dahlia-lap-preprod].freeze
```
akegan: this scares me a little bit, to be renaming all of them instead of renaming just the one that's being created. What's the reasoning behind going through all apps instead of the app that's being deployed?

I wouldn't want to accidentally affect a different project too. I know it's filtered by dahlia-lap but still seems a bit sketchy

lib/update_app_name.rb
```diff
@@ -0,0 +1,58 @@
...
+
+  def new_app_name(app)
+    if app['pr_number'].present?
+      "dahlia-lap-full-pr-#{app['pr_number']}"
+    else
+      failback_name
+    end
+  end
+
+  def failback_name
```
akegan:
```suggestion
  def fallback_name
```

akegan: I'd replace all "failback" with "fallback" if possible

akegan (changes requested):
Couple of minor typos, plus wondering if there's a way to limit the scope to just updating the app that's currently being deployed, that would at least limit the changes to dahlia-lap code that's running the script. Would still have to filter out e.g. dahlia-lap-full.

Gemfile
software-project: I had to double check this. Heroku is running deploy bundle `--with-out development` so that would not work.

lib/update_app_name.rb
software-project: Sure, I understand your concerns. Our first approach was to load review apps via pipeline. But review apps endpoint doesn't return app names. And there is no filtering in the Heroku API: https://devcenter.heroku.com/articles/platform-api-reference#app-list 
But maybe a safer filtering approach would be to get all there review apps via pipeline, then all apps, and continue filtering by ID, rather then name? Well first via ID then name to get the incorrect apps. 
Chris was initially in contact with heroku support and thats the way they suggested. Ideally we would rather get ID of the app in question during the deployment, but I dont think its possible.

akegan: yeah that would make me more comfortable. Are you up for trying that out?

software-project: Sure, will do

akegan:
OK! the PR looks good, how do we test it?

software-project:
> OK! the PR looks good, how do we test it?

Review apps should already be renamed by the script if it worked. If not, you can try running `rake heroku:rename_apps` via heroku console on that review app. Please make sure that correct token is setup in ENV.

software-project:
@akegan I tested it from local host. It looks like it renamed the review apps correctly. So last thing would be to create a new PR basing on this one to check if it's triggered correctly?

akegan (approved):
OK this is good to go. I tested it by creating a new PR. Connection with github is still not working but I'm working with Ben from Exygy to get it sorted out and I don't think it's related to this PR. We can go ahead and merge it. I'll update the callback urls to add the fallback pattern too and we should be back to being able to use review apps! 🎉

PR #337, opened 2020-03-05
Fix missing supplemental application tab on discard changes #171364103

LoganArnett:
This PR is for ticket [#171364103](https://www.pivotaltracker.com/story/show/171364103)

### Before
s
### After
s


akegan (approved):
Good to merge once you update snapshots and tests pass

PR #345, opened 2020-04-04
Change ApplicationsTable to work with single listing #171270123

software-project:
Fixes issue [#171270123](https://www.pivotaltracker.com/n/projects/1405352/stories/171270123)


akegan:
@software-project can we sort by created date to show most recent applications first (both for this new table and the normal applications table. Since leasing agents can't sort and they can't see more than 105 pages of applications 🤦, they are probably most interested in new apps.

Right now it looks like it's sorted by Application # ascending. Application numbers are sequential so could either sort by application submitted date or by application # descending.

spec/javascript/components/applications/ListingApplicationsPage.test.js
```diff
@@ -1,44 +1,31 @@
...
+  test('should render succesfully', (done) => {
     const wrapper = renderer.create(
       <ListingApplicationsPage
         listing={listing}
-        applications={applications} />
+      />
     )
 
     expect(wrapper.toJSON()).toMatchSnapshot()
+    wait(2000)
```
akegan: why does this wait have to be here? can you use act() instead?

akegan:
looks pretty good just have a couple of questions in comments above!

software-project:
> @software-project can we sort by created date to show most recent applications first (both for this new table and the normal applications table. Since leasing agents can't sort and they can't see more than 105 pages of applications 🤦, they are probably most interested in new apps.
> 
> Right now it looks like it's sorted by Application # ascending. Application numbers are sequential so could either sort by application submitted date or by application # descending.

I think we should be able to sort via SOQL query. Should this be also a default way of sorting (also for `/applications`)?

akegan:
sure. I think that sounds good

akegan (approved):
sorting by date looks great!

PR #349, opened 2020-04-28
Confirm At Least 1 AMI Chart #172512100

LoganArnett:
This PR is for ticket [#172512100](https://www.pivotaltracker.com/story/show/172512100)

### DIFF
--- a/app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
+++ b/app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
@@ -36,7 +36,6 @@ const ConfirmedHouseholdIncome = ({ listingAmiCharts, form, visited }) => {
 
   useEffect(() => {
     const getAmiCharts = async () => {
-      console.log(listingAmiCharts)
       if (listingAmiCharts.length > 0) {
         let chartTypes = listingAmiCharts.map((chart) => {
           const value = `${chart.ami_chart_type} - ${chart.ami_chart_year}`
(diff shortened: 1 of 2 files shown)


app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -51,22 +51,23 @@ const ConfirmedHouseholdIncome = ({ listingAmiCharts, form, visited }) => {
 
   useEffect(() => {
     const getAmiCharts = async () => {
-      let chartTypes = []
       let amis = await getAmis(listingAmiCharts, totalHouseholdSize)
```
akegan: we should be able to remove this query entirely depending on what's available in listingAmiCharts already.

akegan (changes requested):
I think we can remove an excess API call here now before we merge

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
LoganArnett: I pushed a removal here and found that the getAmis was only used here and in a file testing the function so I removed those instances as well

app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js
```diff
@@ -43,19 +29,17 @@ export const getAmiPercent = ({income, ami}) => {
...
   const [ selectedType, setSelectedType ] = useState(currentAmiChartType || '')
 
   useEffect(() => {
     const getAmiCharts = async () => {
-      let amis = await getAmis(listingAmiCharts, totalHouseholdSize)
-
-      if (amis.length > 0) {
-        let chartTypes = amis.map((chart) => {
-          const value = `${chart.name} - ${chart.year}`
+      console.log(listingAmiCharts)
```
akegan:
```suggestion
```

akegan (approved):
Need to remove a stray console.log, but otherwise looks good. You can ignore the circleCI builds for now.

akegan:
test comment for slack integration

akegan:
another test comment for slack integration

PR #350, opened 2020-04-29
Switch from semaphore to CircleCI #170582273

software-project:
https://www.pivotaltracker.com/story/show/170582273


spec/javascript/utils/SetupBrowserAndPage.js
```diff
@@ -15,7 +15,7 @@ const IgnoreImageAndCSSLoad = async (page) => {
 }
 
 const SetupBrowser = async () => {
-  let browser = await puppeteer.launch({ headless })
```
akegan, line 18: this looks not right. headless and removecss should be separate things and we should still be indicating whether or not it's headless?

.circleci/config.yml
```diff
@@ -0,0 +1,159 @@
...
+            wget localhost:3000
+            cat index.html
+      - run: yarn e2e
+  coverage:
+    executor: ruby_node
+    steps:
+      - checkout
+      - *attach_workspace
+      # - run: ./tmp/cc-test-reporter format-coverage -t lcov -o tmp/codeclimate.frontend.json coverage/frontend/.resultset.json
+      # - run: ./tmp/cc-test-reporter format-coverage -t simplecov -o tmp/codeclimate.backend.json coverage/backend/.resultset.json
```
akegan:
```suggestion
```

.circleci/config.yml
```diff
@@ -0,0 +1,159 @@
...
+    steps:
+      - checkout
+      - *attach_workspace
+      - *restore_cache
+      - puppeteer/install
+      - run:
+          name: install yarn dependencies
+          command: yarn install --check-files
+      - run: bundle --path vendor/bundle
+      - run: bundle exec rake assets:precompile
```
akegan, line 110: Now that we are persisting state correctly across jobs, we probably don't need to re-install packages, right?

software-project: I'll give it a try

spec/javascript/utils/SetupBrowserAndPage.js
akegan: this is still an issue. we need to be able to run non-headless locally. I'll fix it

PR #352, opened 2020-05-07
Fix AMI percent input bug #172688620

james-wills-sf:
[pivotal ticket #172688620](https://www.pivotaltracker.com/story/show/172688620)

[review app link](https://dahlia-lap-full-pr-352.herokuapp.com)

## Summary
This change makes the supplemental application form submit properly when
the AMI percent input text has a percent sign in it. Additionally, this
change makes the behavior of the percent input closer to that of the
currency input, where a percent sign is automatically added after a
user defocuses the view, and the input only allows numbers (no decimal
points) so a failure case ends up being pretty much impossible.

## Screenshot (before): page error when submitting with percent sign in input
- 2020-05-06 at 5 14 42 PM

## Screencast (after): AMI percentages input
Note that the user is only allowed to type digits [0-9], if you type anything else it won't type anything. This is the same behavior the currency inputs have
AMIpercentage

## Screencast (after): Currency input (just to show that this didn't break)
currencyform

### DIFF
--- a/app/javascript/utils/form/final_form/Field.js
+++ b/app/javascript/utils/form/final_form/Field.js
@@ -8,14 +8,28 @@ const {
   labelize
 } = formOptions
 
+const FieldType = {
+  Text: 0,
+  Currency: 1,
+  Percent: 2
+}
+Object.freeze(FieldType)
+
 // Make react-final-form does include empty values on submit.
 // Source: https://github.com/final-form/react-final-form/issues/130#issuecomment-425482365
-const identity = (value, name, isCurrency = false) => {
-  if (value && isCurrency) {
-    // Here we remove anything that is not a numeric digit
-    // or a decimal place while editing the field
-    return value.toString().replace(/[^\d|.]/g, '')
+const identity = (value, name, fieldType = FieldType.Text) => {
+  if (value) {
+    switch (fieldType) {
+      case FieldType.Currency:
+        // Here we remove anything that is not a numeric digit
+        // or a decimal place while editing the field
+        return value.toString().replace(/[^\d|.]/g, '')
+      case FieldType.Percent:
+        // Only allow digits for percent fields
+        return value.toString().replace(/[^\d]/g, '')
+    }
   }
+
   return value
 }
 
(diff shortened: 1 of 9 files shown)


app/javascript/utils/form/final_form/Field.js
```diff
@@ -78,7 +92,37 @@ export const CurrencyField = ({ fieldName, validation, id, label, placeholder, m
...
+        <FieldError meta={meta} />
+      </div>
+    )}
+  </Field>
+)
+
+/**
+ * Field that allows only whole integer percent values.
+ */
+export const PercentField = ({ fieldName, validation, id, label, placeholder, maxLength, disabled, isDirty = true }) => (
```
james-wills-sf, line 106: This is basically just copied and pasted from `CurrencyField`, except for the parse and format props in `Field`.

app/javascript/utils/formUtils.js
```diff
@@ -14,20 +14,31 @@ const toOptions = (items) => {
...
-// Formats a numer to currency in format eg. $1,000.00
+// Formats a number to currency in format eg. $1,000.00
 // If the field is empty, return null to prevent salesforce issues.
 const formatPrice = (value) => {
   if (!value) return null
   let valueString = value.toString().replace(/[^.|\d]/g, '')
 
   // return value if value is not valid number
-  if (parseFloat(valueString)) {
+  if (!isNaN(parseFloat(valueString))) {
```
james-wills-sf, line 24: fix issue where `0` didn't trigger formatting in currency. Now it does

akegan (commented):
One quick note from QA: looks like you can't type in "." so you can't do fractional percents. I'm not sure if they round or not. will ask in slack.

akegan:
PR otherwise looks great! should be good to go once we resolve the int vs float question

james-wills-sf:
@akegan I added a commit that updates the form to accept float values if you want to take another look! 

AMIPercentFloat

akegan (approved):
Looks good! Once you merge you can pull lastest master into the release branch. Master also contains https://github.com/SFDigitalServices/sf-dahlia-lap/pull/349 which I don't think is on the release branch but is related so I think we can pull both into the release branch

PR #353, opened 2020-05-11
Fix invalid percent formatting bug #172688620

james-wills-sf:
[pivotal ticket](https://www.pivotaltracker.com/n/projects/1405352/stories/172688620)

Previously, if a user entered 5.5.5 into the AMI percent field, the
formatter would auto-format it to be 5.5% because of how parseFloat()
behaves. This change makes it so that 5.5.5 will trigger a validation
error and will stop the page from saving successfully.

- 2020-05-11 at 3 48 19 PM

### DIFF
--- a/spec/javascript/utils/form/validations.test.js
+++ b/spec/javascript/utils/form/validations.test.js
@@ -160,12 +160,10 @@ describe('validate', () => {
         expect(validate.isValidCurrency(VALIDATION_MSG)('$2000.53')).toEqual(undefined)
         expect(validate.isValidCurrency(VALIDATION_MSG)('2,000')).toEqual(undefined)
       })
-      // Maybe validation should be enhanced for this case?
       test('when a currency string with too many commas is entered', () => {
         expect(validate.isValidCurrency(VALIDATION_MSG)('$2000,')).toEqual(undefined)
+        expect(validate.isValidCurrency(VALIDATION_MSG)('$2000,.52')).toEqual(undefined)
         expect(validate.isValidCurrency(VALIDATION_MSG)('$2,0,0,0')).toEqual(undefined)
-        expect(validate.isValidCurrency(VALIDATION_MSG)('$,,')).toEqual(undefined)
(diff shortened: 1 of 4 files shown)


akegan:
actual review app: https://dahlia-lap-full-pr-353.herokuapp.com/

akegan (approved):
This change + QA looks good. Currency fields have this same issue. Wondering if it makes sense to fix in this PR too or open up a separate ticket.

james-wills-sf:
@akegan it was a simple enough fix that I added it as a new commit to this PR, can you take another look?

currencyform2

spec/javascript/utils/form/validations.test.js
```diff
@@ -160,15 +160,20 @@ describe('validate', () => {
         expect(validate.isValidCurrency(VALIDATION_MSG)('$2000.53')).toEqual(undefined)
         expect(validate.isValidCurrency(VALIDATION_MSG)('2,000')).toEqual(undefined)
       })
-      // TODO: Enhance currency validation to exclude these cases.
-      test('when any string with $s and ,s is entered', () => {
-        expect(validate.isValidCurrency(VALIDATION_MSG)('$2000$')).toEqual(undefined)
-        expect(validate.isValidCurrency(VALIDATION_MSG)('$2,000,')).toEqual(undefined)
-        expect(validate.isValidCurrency(VALIDATION_MSG)('$')).toEqual(undefined)
+      // Maybe validation should be enhanced for this case?
```
akegan: not sure. Users are prevented from entering these values at least on Chrome

james-wills-sf: @akegan yep it's across all browsers, (it happens in the `identity()` function in `Field.js`. I'll just update the comment to explain why we don't care about that case

Edit: I just removed the comment entirely and updated the regex to at least not allow things like `,`, `$,`, `,0.00`

akegan (approved):
looks good and QA looks good too

PR #354, opened 2020-05-12
Fix submission type #171270123

software-project:
Fix for [#171270123](https://www.pivotaltracker.com/n/projects/1405352/stories/171270123) and https://www.pivotaltracker.com/story/show/172804251


akegan (commented):
In other cases I thought we were able to distinguish between the value we display and the actual value of the filter. Would that be possible here?

software-project:
Well, I just found another bug. Select a listing, click filter, change to any listing and then filter again. We have two ways of fixing it I suppose. One in the controller or in React. Which would you prefer?

james-wills-sf:
I'm going to take over this PR since Martin is out of hours for the day.

james-wills-sf:
@akegan @software-project updated this to fix the issue on the frontend.

It turns out, when we used "null" as a value for one option, the actual input created looks like this:
```
<select ...>
  <option>Any Listing</option>
  <option value="<listing-id>">Some listing name</option>
  ...
</select>
```

And when the first option is selected on submit it uses the label, "Any Listing", since there is no value attr. I fixed it by passing in an empty string as value instead of null, so the html looks like this instead:
```
<select ...>
  <option value="">Any Listing</option>
  <option value="<listing-id>">Some listing name</option>
  ...
</select>
```

And then I scrub the filters for empty strings before passing them to the api.

One other nice thing this fixes is before we would pass filters like `application-name: ""` if you typed something in that text field then deleted it. The scrubber works on all fields so that won't happen anymore.

applicationfilter

spec/javascript/components/applications/__snapshots__/ListingApplicationsPage.test.js.snap
```diff
@@ -201,7 +201,7 @@ Array [
                   <option
                     value=""
                   >
-                    Select One...
```
james-wills-sf, line 204: The type field used to default to "Select One..." with "Any Type" at the bottom of the list. I considered this a bug since select one and any type have the same functionality, so now we just default to "Any Type"

## Screenshot (before)
- 2020-05-12 at 2 24 29 PM

## Screenshot (after)
- 2020-05-12 at 2 24 10 PM

akegan (approved):
These changes look good. Just noticed that we have the same bug in our lease up applications table too. Not sure if it's worth fixing in this PR too or we could make a separate ticket, up to you

james-wills-sf:
@akegan I went ahead and fixed lease up too.

I also did a code search to see if this is happening anywhere else and I only found one place in [ConfirmedHouseholdIncome.js](https://github.com/SFDigitalServices/sf-dahlia-lap/blob/master/app/javascript/components/supplemental_application/sections/ConfirmedHouseholdIncome.js#L47), but it appears that they're manually handling select changes so I am not going to mess with it.

leaseupapplicationfilter

akegan (approved):
looks good! thanks for making the update

PR #356, opened 2020-05-18
Separate AMI chart and year dropdowns #172864201

james-wills-sf:
[#172864201](https://www.pivotaltracker.com/n/projects/1405352/stories/172864201)

## Summary

Previously we combined AMI charts and years, but recently we found that
agents want to use more recent years than our options provided. This
change makes it so that agents will always be able to choose the current
year, even if it's not included in our AMI chart data. Additionally we
split up the year and AMI chart dropdowns because there is no longer a
1:1 mapping between the two.

### A couple decisions I made around how years are parsed

- I allow non-year strings and just append them to the beginning of the year dropdown
  - Why? I could see a case where there are options like "2020 adjusted" or something.
- I consider years more than 10 years away from the current year to be treated like non-years
  - Why? I didn't want something like "1000" to be entered and result in a dropdown with a thousand options

## Screencast
amichartyears

### DIFF
--- a/app/javascript/components/applications/application_form/formOptions.js
+++ b/app/javascript/components/applications/application_form/formOptions.js
@@ -1,11 +1,12 @@
-import { isEmpty, map } from 'lodash'
+const isNullOrEmptyString = (value) => value === undefined || value === null || value === ''
 
 const labelize = (options, attrs = {}, noPlaceholder = false) => {
-  if (isEmpty(options)) return []
+  if (options.length === 0) return []
 
   let emptyInitialOptionPresent =
-    isEmpty(options[0]) ||
-    (options[0].hasOwnProperty('value') && isEmpty(options[0].value)) || noPlaceholder
+    noPlaceholder ||
+    options[0] === '' ||
+    (options[0].hasOwnProperty('value') && isNullOrEmptyString(options[0].value))
 
   let labelizedOptions = []
   if (!emptyInitialOptionPresent) {
@@ -17,13 +18,13 @@ const labelize = (options, attrs = {}, noPlaceholder = false) => {
   }
 
   return labelizedOptions.concat(
-    map(options, (option) => (
-      {
+    options.map(option => {
+      return {
         value: option.hasOwnProperty('value') ? option.value : option,
-        label: option.label || option,
+        label: option.hasOwnProperty('label') ? option.label : option,
         ...(option.disabled && { disabled: option.disabled })
       }
-    ))
+    })
   )
 }
 
--- a/app/javascript/components/applications/application_form/formOptions.js
+++ b/app/javascript/components/applications/application_form/formOptions.js
@@ -1,10 +1,13 @@
+import { isEmpty } from 'lodash'
+
 const isNullOrEmptyString = (value) => value === undefined || value === null || value === ''
 
 const labelize = (options, attrs = {}, noPlaceholder = false) => {
-  if (options.length === 0) return []
+  if (isEmpty(options)) return []
 
   let emptyInitialOptionPresent =
     noPlaceholder ||
+    isEmpty(options) ||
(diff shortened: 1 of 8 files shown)


app/javascript/components/applications/application_form/formOptions.js
```diff
@@ -1,11 +1,12 @@
-import { isEmpty, map } from 'lodash'
+const isNullOrEmptyString = (value) => value === undefined || value === null || value === ''
 
 const labelize = (options, attrs = {}, noPlaceholder = false) => {
-  if (isEmpty(options)) return []
+  if (options.length === 0) return []
 
   let emptyInitialOptionPresent =
-    isEmpty(options[0]) ||
-    (options[0].hasOwnProperty('value') && isEmpty(options[0].value)) || noPlaceholder
```
james-wills-sf, line 8: `isEmpty(options[0].value)` here is a bug. [Lodash isEmpty](https://lodash.com/docs/4.17.15#isEmpty) isn't meant to be called on strings and null values don't work as expected.

app/javascript/components/applications/application_form/formOptions.js
```diff
@@ -17,13 +21,13 @@ const labelize = (options, attrs = {}, noPlaceholder = false) => {
   }
 
   return labelizedOptions.concat(
-    map(options, (option) => (
```
james-wills-sf, line 20: I don't think there's any need for us to use lodash map anymore

app/javascript/components/applications/application_form/formOptions.js
```diff
@@ -17,13 +21,13 @@ const labelize = (options, attrs = {}, noPlaceholder = false) => {
...
-      {
+    options.map(option => {
+      return {
         value: option.hasOwnProperty('value') ? option.value : option,
-        label: option.label || option,
+        label: option.hasOwnProperty('label') ? option.label : option,
```
james-wills-sf, line 27: The old code was causing a bug where a null label in an option like this:

```
{
  value: "value",
  label: null
}
```

Would labelize into
```
{
  value: "value",
  label: {
    value: "value",
    label: null
  }
}
```

Which breaks during render

spec/javascript/utils/amiYearUtils.test.js
```diff
@@ -0,0 +1,90 @@
...
+  test('should handle multiple recent past year options properly', async () => {
+    expect(getYearsWithMockOptions(2019, 2020)).toEqual([2019, 2020])
+    expect(getYearsWithMockOptions(2020, 2019)).toEqual([2019, 2020])
+    expect(getYearsWithMockOptions(2018, 2020)).toEqual([2018, 2019, 2020])
+    expect(getYearsWithMockOptions(2018, 2019)).toEqual([2018, 2019, 2020])
+    expect(getYearsWithMockOptions(2019, 2019, 2020)).toEqual([2019, 2020])
+    expect(getYearsWithMockOptions(2019, 2020, 2020)).toEqual([2019, 2020])
+  })
+
+  test('should handle single close future year options properly', async () => {
```
akegan, line 55: Not totally sure about this behavior but shouldn't ever really come up so should be OK?

akegan (approved):
looks good! normally for release updates we try to branch off of the release branch instead of off of master to avoid adding more complication to getting the release out, but in this case, the only other commit on master not in the release is the circle CI code so it should be OK.

spec/javascript/utils/amiYearUtils.test.js
james-wills-sf: Yeah I wasn't sure either, I thought maybe if the charts were updated in december or something they might have the future year?

akegan: yeah it seems like a graceful way to handle future dates in case they show up

PR #358, opened 2020-05-26
Remove heroku rename pipeline script #172829373

software-project:
Fixes [Setup predictable urls for heroku review apps](https://www.pivotaltracker.com/n/projects/1405352/stories/172829373/comments/214632953)
As Heroku now allows to set predictable url for PR, we want to set it back to how it used to me and use review apps prefixed as: `dahlia-lap-full-pr356.herokuapp.com`


akegan (changes requested):
I just took a look at the PR where we added these scripts in the first place and it looks like there are some other pieces left that we could take out
1. Remove the [platform api](https://github.com/SFDigitalServices/sf-dahlia-lap/pull/335/files#diff-8b7db4d5cc4b8f6dc8feb7030baa2478R37) gem
2. We could get rid of postdeploy.sh and go back to just running rake db:migrate

akegan (approved):
looks good now, thanks

PR #363, opened 2020-06-12
Replace usages of master with main #173314185

james-wills-sf:
[#173314185](https://www.pivotaltracker.com/n/projects/1405352/stories/173314185)


akegan (approved):
looks good. thanks for taking this on

PR #367, opened 2020-06-24
[Partners] Run e2e production tests #173302542

james-wills-sf:
[#173302542](https://www.pivotaltracker.com/n/projects/1405352/stories/173302542)

## Summary
Add tests that will run only on the "production" branch with prod env
variables.

## Steps taken to verify
- [x] added this pr name to the e2e-production conditional to test that it runs the production tests (and skips the rest)

- [x] Verify that regexes work, and that release branches will also perform the production e2e workflow.

- [x] removed this pr name to the e2e-conditional to test that regular branches only build in dev.

- [x] sshed into e2e-production box and verified that it's using prod credentials.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -189,7 +189,6 @@ workflows:
           filters:
             branches:
               only:
-                - /^chores.*$/
                 - /^releases.*$/
                 - /^release.*$/
                 - production
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -160,7 +160,6 @@ workflows:
             branches:
               ignore:
                 - production
-                - chores/circle-ci-on-production-#173302542
       - frontend:
           requires:
             - build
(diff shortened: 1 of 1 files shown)


james-wills-sf:
@akegan this and the webapp pr: https://github.com/SFDigitalServices/sf-dahlia-web/pull/1359 should be ready to review.

I set it up so that any PR starting with release or releases will automatically run the e2e-production workflow along with the dev ones. Does that sound good to you?

akegan (approved):
looks good. still not totally clear on the logistics, but looks like it should work!

james-wills-sf:
@akegan the logic is basically this:

1. We added an extra job that's only for production and release branches that runs the same e2e tests as dev but adds the `partners-prod` context, which is just a bunch of prod env variable overwrites.

2. We also don't run dev tests for the production branch.

### Job graphs by branch type:
#### regular branches run dev tests only


#### release branches run dev and e2e-production tests


#### `production` branch runs e2e-production tests only

PR #369, opened 2020-06-25
[Partners] simplify circleci config

james-wills-sf:
I learned how to share filters across jobs in my webapp pr, this just
uses the same simplification in partners.


akegan (approved):
looks good. nice simplification

PR #370, opened 2020-06-26
Bugs/change to put request #172115886

software-project:
Fixes: https://www.pivotaltracker.com/story/show/172115886

### DIFF
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -1,15 +1,23 @@
 import apiService from '~/apiService'
 import Alerts from '~/components/Alerts'
-import { isEmpty, find, isEqual, every, reject } from 'lodash'
+import { isEmpty, find, isEqual, every, reject, forEach } from 'lodash'
 import { convertCurrency } from '~/utils/form/validations'
 
 export const updateApplication = async (application, prevApplication) => {
   const primaryApplicantContact = application.applicant && application.applicant.id
   const applicationId = application['id']
+  const changedFields = {id: application['id']}
+
+  forEach(application, (value, key) => {
+    if (!isEqual(prevApplication[key], value)) {
+      changedFields[key] = value
+    }
+  })
+
   // await lease and base application updates first
   const initialResponses = await Promise.all([
     updateLease(application['lease'], primaryApplicantContact, applicationId),
-    apiService.submitApplication(application, true)
+    apiService.submitApplication(changedFields, true)
   ])
   // next we update rental assistances if applicable
   await Promise.all(updateUnsavedRentalAssistances(application, prevApplication))
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -8,7 +8,6 @@ export const updateApplication = async (application, prevApplication) => {
   const primaryApplicantContact = application.applicant && application.applicant.id
   const applicationId = application['id']
   const changedFields = filterChanged(prevApplication, application)
-  changedFields['id'] = application['id']
 
   // await lease and base application updates first
   const initialResponses = await Promise.all([
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -1,25 +1,8 @@
 import apiService from '~/apiService'
 import Alerts from '~/components/Alerts'
-import { isEmpty, find, isEqual, every, reject, forEach, isPlainObject } from 'lodash'
+import { isEmpty, find, isEqual, every, reject } from 'lodash'
 import { convertCurrency } from '~/utils/form/validations'
-
-const filterChanged = (prev, current) => {
-  const changedFields = {}
-  forEach(current, (value, key) => {
-    if (!isEqual(prev[key], value)) {
-      if (isPlainObject(value)) {
-        const obj = filterChanged(prev[key], value)
-        if (!isEmpty(obj)) {
-          obj['id'] = value.id
-          changedFields[key] = obj
-        }
-      } else {
-        changedFields[key] = value
-      }
-    }
-  })
-  return changedFields
-}
+import { filterChanged } from '~utils/utils'
 
 export const updateApplication = async (application, prevApplication) => {
   const primaryApplicantContact = application.applicant && application.applicant.id
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -1,18 +1,31 @@
 import apiService from '~/apiService'
 import Alerts from '~/components/Alerts'
-import { isEmpty, find, isEqual, every, reject, forEach } from 'lodash'
+import { isEmpty, find, isEqual, every, reject, forEach, isPlainObject } from 'lodash'
 import { convertCurrency } from '~/utils/form/validations'
+
+const filterChanged = (prev, current) => {
+  const changedFields = {}
+  forEach(current, (value, key) => {
+    if (!isEqual(prev[key], value)) {
+      if (isPlainObject(value)) {
+        const obj = filterChanged(prev[key], value)
+        if (!isEmpty(obj)) {
+          obj['id'] = value.id
+          changedFields[key] = obj
+        }
+      } else {
+        changedFields[key] = value
+      }
+    }
+  })
+  return changedFields
+}
 
 export const updateApplication = async (application, prevApplication) => {
   const primaryApplicantContact = application.applicant && application.applicant.id
   const applicationId = application['id']
-  const changedFields = {id: application['id']}
-
-  forEach(application, (value, key) => {
-    if (!isEqual(prevApplication[key], value)) {
-      changedFields[key] = value
-    }
-  })
+  const changedFields = filterChanged(prevApplication, application)
+  changedFields['id'] = application['id']
 
   // await lease and base application updates first
   const initialResponses = await Promise.all([
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -2,7 +2,7 @@ import apiService from '~/apiService'
 import Alerts from '~/components/Alerts'
 import { isEmpty, find, isEqual, every, reject } from 'lodash'
 import { convertCurrency } from '~/utils/form/validations'
-import { filterChanged } from '~utils/utils'
+import { filterChanged } from '~/utils/utils'
 
 export const updateApplication = async (application, prevApplication) => {
   const primaryApplicantContact = application.applicant && application.applicant.id
--- a/app/javascript/components/supplemental_application/sections/DemographicsInputs.js
+++ b/app/javascript/components/supplemental_application/sections/DemographicsInputs.js
@@ -3,6 +3,7 @@ import { range } from 'lodash'
 import { SelectField } from '~/utils/form/final_form/Field.js'
 
 import FormGrid from '~/components/molecules/FormGrid'
+import formUtils from '~/utils/formUtils'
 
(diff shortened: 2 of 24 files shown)


app/javascript/components/supplemental_application/actions.js
```diff
@@ -2,14 +2,17 @@ import apiService from '~/apiService'
 import Alerts from '~/components/Alerts'
 import { isEmpty, find, isEqual, every, reject } from 'lodash'
 import { convertCurrency } from '~/utils/form/validations'
+import { filterChanged } from '~utils/utils'
```
james-wills-sf: nit: should this be `~/utils/utils`?

app/javascript/components/supplemental_application/sections/DemographicsInputs.js
```diff
@@ -16,6 +17,7 @@ const DemographicsInputs = () => {
             <SelectField id='demographics-dependents'
               fieldName='number_of_dependents'
               options={numberOfPeopleOptions}
+              format={formUtils.formatNumber}
```
james-wills-sf, line 20: What does it mean to format a SelectField, isn't it a dropdown? Would it make more sense to do this formatting as a preprocessing step?

app/javascript/utils/formUtils.js
```diff
@@ -59,6 +59,12 @@ const scrubEmptyValues = (obj, scrubEmptyStrings = false) => {
   return filterValues(obj, valueIsNonEmpty)
 }
 
+const formatNumber = (value) => {
```
james-wills-sf, line 62: Nit: Can you add a unit test for this?

app/javascript/utils/utils.js
```diff
@@ -21,6 +21,28 @@ export const formatPercent = (value) => {
...
+  if (!prev) {
+    return current
+  }
+  const changedFields = {}
+  forEach(current, (value, key) => {
+    if (!isEqual(prev[key], value)) {
+      if (isPlainObject(value) && prev[key] !== null) {
+        const obj = filterChanged(prev[key], value)
+        if (!isEmpty(obj)) {
+          obj['id'] = value.id
```
james-wills-sf: If the object doesn't have an id will this set `{ id: undefined }`?

spec/javascript/utils/utils.test.js
```diff
@@ -21,3 +34,30 @@ describe('convertCurrency', () => {
     expect(updatedFormValues.lease.another_nested_level.additional_rent).toEqual(425.99)
   })
 })
+
+describe('filterChanged', () => {
```
james-wills-sf, line 38: I just thought of a few cases and I'm not sure what the behavior will be, can you add them?

1) a test to check that IDs will be updated properly when entire objects are swapped
```
prevApp = {
  id: 'appid',
  primary_applicant: {
    id: 'primary1',
    name: 'joe',
  }
}

newApp = {
  id: 'appid',
  primary_applicant: {
    id: 'primary2',
    name: 'james',
  }
}

expectedFilteredApp = { 
  id: 'appid',
  primary_applicant: {
    id: 'primary2',
    name: 'james',
  }
}
```

2) a test with arrays
```
prevApp = {
  id: 'appid',
  applicant_contacts: [
   {
      id: 'contact1',
      name: 'joe',
    },
   {
      id: 'contact2',
      name: 'james',
    },
  ],
  applicant_contacts_unchanged: [
   {
      id: 'contact1',
      name: 'joe',
    },
   {
      id: 'contact2',
      name: 'james',
    },
  ],
  other_changed_list: [ 1, 2, 3 ],
  other_unchanged_list: [ 1, 2, 3 ],
}

newApp = {
  id: 'appid',
  applicant_contacts: [
   {
      id: 'contact3',
      name: 'chris',
    },
   {
      id: 'contact4',
      name: 'amy',
    },
  ],
  applicant_contacts_unchanged: [
   {
      id: 'contact1',
      name: 'joe',
    },
   {
      id: 'contact2',
      name: 'james',
    },
  ],
  other_changed_list: [ 1, 2, 3, 4 ],
  other_unchanged_list: [ 1, 2, 3 ],
}

expectedFilteredApp = {
  id: 'appid',
  applicant_contacts: [
   {
      id: 'contact3',
      name: 'chris',
    },
   {
      id: 'contact4',
      name: 'amy',
    },
  ],
  other_changed_list: [ 1, 2, 3, 4 ],
}
```

3: empty object
```
prevApp = { }
newApp = { }
expectedFilteredApp: { }
```

4: object without an id and no changes
```
prevApp = { a: '1' }
newApp = { a: '1' }
expectedFilteredApp: { }
```

5: object without an id and changes
```
prevApp = { a: '1' }
newApp = { a: '2' }
expectedFilteredApp: { a: '2' }
```

6: nested object without changes (when parent does have changes)
```
prevApp = { id: 'abc', a: '1', b: { id: 'abcd', c: '2' } }
newApp = { id: 'abc', a: '2', b: { id: 'abcd', c: '2' } }
expectedFilteredApp: { id: 'abc', a: '2' }
```

7: nested object with changes (when parent does not have other changes)
```
prevApp = { id: 'abc', a: '1', b: { id: 'abcd', c: '2' } }
newApp = { id: 'abc', a: '1', b: { id: 'abcd', c: '3' } }
expectedFilteredApp: { id: 'abc', b: { id: 'abcd', c: '3' } }
```

app/services/force/custom_api/application_service.rb
```diff
@@ -47,9 +47,16 @@ def application_snapshot(id)
         application
       end
 
-      def submit(custom_api_attrs)
+      def submit(custom_api_attrs, method)
         application_params = application_defaults.merge(custom_api_attrs)
-        result = api_post('/LeasingAgentPortal/shortForm', application_params)
+        application_params = application_params.except('primaryApplicant') if application_params['primaryApplicant'].blank?
```
james-wills-sf, line 52: Why are this and the exceptions on line 56 necessary?

config/routes.rb
```diff
@@ -53,7 +53,8 @@
       resources :rental_assistances, path: '/rental-assistances', only: %w[index create update destroy]
 
       scope '/short-form' do
-        post 'submit' => 'short_form#submit'
+        # put 'submit' => 'short_form#submit'
```
james-wills-sf: nit: remove commented out code

spec/javascript/utils/utils.test.js
```diff
@@ -1,4 +1,6 @@
 import { convertCurrency } from '~/utils/form/validations'
+import { filterChanged } from '~utils/utils'
```
james-wills-sf: nit: should this be `~/utils/utils`?

spec/vcr/api/v1/short-form/submit/pre-lottery/with-supplemental-param.yml
```diff
@@ -45,17 +49,16 @@ http_interactions:
       - chunked
     body:
       encoding: ASCII-8BIT
-      string: '{"access_token":"<<ACCESS_TOKEN>>","instance_url":"https://<<SALESFORCE_HOST>>","id":"https://<<SALESFORCE_HOST>>/id/00D1D0000009nrYUAQ/005U00000066jl9IAA","token_type":"Bearer","issued_at":"1590532069542","signature":"DXXasvQj8CwCocjhGb7x7tfLBTy2CYuA7JJnOEOubiI="}'
-    http_version: null
-  recorded_at: Tue, 26 May 2020 22:27:49 GMT
+      string: '{"access_token":"<<ACCESS_TOKEN>>","instance_url":"https://<<SALESFORCE_HOST>>","id":"https://test.salesforce.com/id/00D1F000000I2yHUAS/005U00000066jl9IAA","token_type":"Bearer","issued_at":"1593681327685","signature":"5/hgfYw8iVreqIa2CIgS/9vJ4PmeohwrrvXs79cHWxg="}'
```
james-wills-sf, line 52: Why not use the SALESFORCE_HOST variable here? curious because I think I was the one to make that change

james-wills-sf (commented):
@software-project reviewed, I mostly have nit comments and questions.

Do you know if any of the e2e tests are using the new put requests?

app/javascript/components/supplemental_application/actions.js
software-project: I believe so, I put that function to `app/javascript/utils/utils.js`

app/javascript/components/supplemental_application/sections/DemographicsInputs.js
software-project: I believe thats a fix that Logan made. The issue was that select returned values as strings and we needed them to be numbers. That fixed the issue.

app/javascript/utils/formUtils.js
software-project: Ok, will add a test for this one.

app/javascript/utils/utils.js
software-project: I don't think this case can happen in current scenario as this is used only for editing defined object. Or more precisely for objects related to `application`, like `primary_applicant`. If we were to add element to an array it would be handled by line 38. But if you like I can future proof it. But should id be set to `null` in that case or not added at all?

app/services/force/custom_api/application_service.rb
software-project: `primaryApplicant` gets pulled in short_form_controller.rb#10 but it's empty, so to avoid messing with SF I'm removing it here but only if we are not actually updating it. Regarding line 56, we dont't want to edit them and I recall having SF errors when they were present. On the other hand we want those fields for POST request, so we can't filter them on controller params.

config/routes.rb
software-project: will do

spec/javascript/utils/utils.test.js
software-project: Same case as above.

spec/javascript/utils/utils.test.js
software-project: 👍

app/javascript/utils/utils.js
james-wills-sf: I'd say it should not be added at all, but you're right that practically this is unlikely

app/javascript/components/supplemental_application/actions.js
james-wills-sf: right I just meant the inclusion of the `/` after the `~`

software-project: I'm not sure what are you suggesting but if to change to `import { filterChanged } from '~utils'`, thats not going to work.

james-wills-sf: @software-project no I'm suggesting to change to `import { filterChanged } from '~/utils/utils'` since that seems to be the dominant style (see line 4)

`~/utils/utils` vs `~utils/utils`

app/javascript/utils/utils.js
software-project: what should I do? Leave it as is, future proof or sth else?

james-wills-sf: I think we should future-proof it. I can see a scenario where a nested object doesn't have an id in the future

spec/vcr/api/v1/short-form/submit/pre-lottery/with-supplemental-param.yml
software-project: This is a VCR cassette it's auto generated.

software-project:
@james-wills-sf I've updated this PR according to your review. Also I've found one issue with demographics (can't reset values to null) and I've already let Larry know.

james-wills-sf (approved):
@software-project thanks for adding new tests, LGTM

james-wills-sf:
@software-project just tested it out and looked at the network requests, lgtm!

PR #379, opened 2020-08-20
Change leaseups radio fields to checkboxes [DAH-24]

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAHLIAFE-24


app/javascript/utils/form/final_form/Field.js
```diff
@@ -194,6 +194,40 @@ export const CheckboxField = ({ fieldName, label, blockNote, validation, id, ari
   </Field>
 )
 
+export const TextCheckboxField = ({ fieldName, label, blockNote, validation, id, ariaLabelledby, form, trueFalue = 'Yes', falseValue = 'No' }) => {
```
james-wills-sf: Can you add a comment above this component explaining the difference between it and the regular CheckboxField?

james-wills-sf: Also typo: `trueFalue` should be `trueValue`

app/javascript/utils/form/final_form/Field.js
```diff
@@ -194,6 +194,40 @@ export const CheckboxField = ({ fieldName, label, blockNote, validation, id, ari
...
+    <Field
+      name={fieldName}
+      validate={validation}
+      type='input'
+      component='input'
+    >
+      {({ input, meta }) => (
+        <React.Fragment>
+          <div className={classNames('form-group', (meta.error && meta.touched && 'error') || '')} >
+            <Input
```
james-wills-sf: I don't really understand what this hidden input does as it doesn't seem to be attached to the fieldName, can you explain what it does?

app/javascript/utils/form/final_form/Field.js
```diff
@@ -194,6 +194,40 @@ export const CheckboxField = ({ fieldName, label, blockNote, validation, id, ari
...
+  }
+  return (
+    <Field
+      name={fieldName}
+      validate={validation}
+      type='input'
+      component='input'
+    >
+      {({ input, meta }) => (
+        <React.Fragment>
```
james-wills-sf: nit: I think you only need `React.Fragment` when you're wrapping a list of components. Since you're just wrapping a single div here you don't need it.

Also I'm not sure what version this was introduced in, but the new way to do this is just with `<>...</>`
like:
```
return (
  <>
    <div />
    <div />
    <div />
  </>
);
```

james-wills-sf (changes requested):
@software-project I was curious about the necessity of the hidden Input, so I removed it locally and it seems like it works fine without it. I added the change as a commit because it seemed like the easiest way for you to review the change. Feel free to revert it if you find an issue with it, but take a look and test it out. I think it works just as well this way.

I have a few nit comments, and I'm wondering if you can add an e2e test for this? Don't worry about a test to check modifying the value, but at least that the correct boxed are checked on load with existing data that is a mix of Yes's and No's.

If the e2e aren't set up in a way that makes this relatively easy don't worry about it.

app/javascript/utils/form/final_form/Field.js
```diff
@@ -194,6 +194,39 @@ export const CheckboxField = ({ fieldName, label, blockNote, validation, id, ari
   </Field>
 )
 
+export const TextCheckboxField = ({ fieldName, label, blockNote, validation, id, ariaLabelledby, form, trueValue = 'Yes', falseValue = 'No', initialValue = null }) => {
```
james-wills-sf, line 203: Can you add a comment to this component that explains how this is different from the `CheckboxField` component?
something like
```
Checkbox that is backed by a string value (ex: has_developental_disability: "Yes"/"No")
```

app/javascript/utils/form/final_form/Field.js
```diff
@@ -194,6 +194,39 @@ export const CheckboxField = ({ fieldName, label, blockNote, validation, id, ari
...
+    <Field
+      name={fieldName}
+      validate={validation}
+      type='input'
+      component='input'
+      initialValue={initialValue}
+    >
+      {({ input, meta }) => (
+        <div className={classNames('form-group', (meta.error && meta.touched && 'error') || '')} >
+          <Input
```
james-wills-sf, line 217: Can you add a comment explaining why we need the hidden input?

app/javascript/components/molecules/FormGrid.js
```diff
@@ -25,11 +25,13 @@ FormGrid.Row.defaultProps = {
...
 
 // NOTE: Grid is hardcoded. This should not be generic element
-FormGrid.Item = ({ children }) => (
-  <div className='form-grid_item small-12 medium-6 large-3 column'>
-    {children}
-  </div>
-)
+FormGrid.Item = ({ children, wide }) => {
+  return (
+    <div className={`form-grid_item small-12 medium-6 ${wide ? 'large-6' : 'large-3'} column`}>
```
james-wills-sf: nit: for long lists of classnames with conditional values like this I think it's preferrable to use the classnames util:

```
classNames(
  'form-grid_item',
  'small-12',
  'medium-6',
  'column',
  { 'large-6': wide, 'large-3': !wide },
  { wide ? 'large-6' : 'large-3' } // another option that does the same thing
)
```

james-wills-sf (approved):
@software-project a couple of nit comments but this LGTM

app/javascript/components/molecules/FormGrid.js
akegan: what is this change for? Can we pull it out and I can fix the layout in the next PR if necessary?

software-project: `wide` option is to spread grid columns to the full width of the container as it is in Figma designs.

PR #381, opened 2020-08-25
[DAHLIAFE-26] Update tab and card styling 

akegan:
https://sfgovdt.jira.com/browse/DAHLIAFE-26

Pattern lib PR: https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/139

### This PR
1. Updates tabs to match figma and look reasonable for medium + small screen sizes
1. Removes the "App card" border container around things to get the full horizontal white line
1. Makes currency and percent fields have a border radius that matches other fields

### This PR does not
1. Make any changes to general padding/styling of form fields -> I'm assuming that that'll make the most sense to do when we re-arrange other fields

### Pieces to test
1. Make sure tabs still work
1. See tabs on both supp app + listing page (where you click back and forth between listing + applications)


akegan:
@james-wills-sf I didn't really end up adding much more scope to this PR, but I think it's a manageable scope to start with.

akegan:
@cpalmatier can you take a look at the review app here and see if the tab and card behavior looks OK to you? the 2 examples I could find were on a sample supp app, and if you click in to view a listing on the pre-lottery section

app/javascript/utils/form/final_form/Field.js
```diff
@@ -89,6 +89,7 @@ export const CurrencyField = ({ fieldName, validation, id, label, placeholder, m
           id={id || `form-${fieldName}`}
           fieldName={fieldName} />
         <input {...input}
+          type='text'
```
james-wills-sf, line 92: What does adding this line do? Is this what triggers the border radius?

app/javascript/components/organisms/TabsSection.js
```diff
@@ -2,13 +2,13 @@ import React from 'react'
 import TabsMenu from '../molecules/TabsMenu'
 import TabCard from './TabCard'
 
-const TabsSection = ({ children, background, padding, ...props }) => {
+const TabsSection = ({ children, background, padding, tabBackground, ...props }) => {
```
james-wills-sf: is `tabBackground` ever being passed in? Seems like we're just defaulting to white now?

If the tab background is always going to be white I'd rather we just remove the prop options because the way it's done now is not my favorite (the prop is a string that needs to turn into a valid class when `bg-` is prepended, just seems fragile).

james-wills-sf (commented):
@akegan added a couple comments

app/javascript/utils/form/final_form/Field.js
akegan: yes! the border radius styling depends on having the input type defined

app/javascript/components/organisms/TabsSection.js
akegan: yeah I think that's reasonable. I thought we'd be using it in different contexts but we're not

akegan:
@cpalmatier here's a gif of what the tabs look like on the listings view: 
listings_tabs

and on the supp app view:
supp app tab view

cpalmatier:
@akegan looks good to me, thanks!

james-wills-sf (approved):
@akegan lgtm!

PR #382, opened 2020-08-25
Add sidebar to component library #DAHLIAFE-37

james-wills-sf:
Jira ticket: [[#DAHLIAFE-37](https://sfgovdt.jira.com/browse/DAHLIAFE-37)]

- Add StatusPill component
- Add LeaseUpSidebarButtons component
- Add LeaseUpSidebar component


james-wills-sf:
Ugh okay looks like the `standard` library bump causes a ton of new lint errors. I'll investigate and decide what to do.

james-wills-sf:
@akegan this is ready to review now, along with a small [pattern library change](https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/140)

app/javascript/utils/statusUtils.js
```diff
@@ -47,11 +47,31 @@ export const LEASE_UP_SUBSTATUS_OPTIONS = {
...
 }
 
+export const getStatusPillClass = status => {
+  const statusOption = find(LEASE_UP_STATUS_OPTIONS, { value: status })
+  return statusOption ? statusOption.pillStyle : 'status-pill-no-status'
+}
+
+export const getStatusPillLabel = status => {
+  const statusOption = find(LEASE_UP_STATUS_OPTIONS, { value: status })
+  return statusOption ? statusOption.label : 'Comment'
```
akegan: This was actually a mistake in the figma design I think. There will always be a status. Adding a comment just creates another field update comment with the same status/substatus and a different comment.

app/javascript/utils/statusUtils.js
```diff
@@ -1,13 +1,13 @@
...
-  { value: 'Waitlisted', label: 'Waitlisted', style: 'is-waitlisted', commentRequired: true },
-  { value: 'Disqualified', label: 'Disqualified', style: 'is-disqualified', commentRequired: true },
-  { value: 'Approved', label: 'Approved', style: 'is-approved', commentRequired: true },
-  { value: 'Lease Signed', label: 'Lease Signed', style: 'is-leased' }
+  { value: 'Processing', label: 'Processing', style: 'is-processing', pillStyle: 'status-pill-processing', commentRequired: true },
+  { value: 'Withdrawn', label: 'Withdrawn', style: 'is-withdrawn', pillStyle: 'status-pill-withdrawn' },
+  { value: 'Appealed', label: 'Appealed', style: 'is-appealed', pillStyle: 'status-pill-appealed', commentRequired: true },
+  { value: 'Waitlisted', label: 'Waitlisted', style: 'is-waitlisted', pillStyle: 'status-pill-waitlisted', commentRequired: true },
+  { value: 'Disqualified', label: 'Disqualified', style: 'is-disqualified', pillStyle: 'status-pill-disqualified', commentRequired: true },
+  { value: 'Approved', label: 'Approved', style: 'is-approved', pillStyle: 'status-pill-approved', commentRequired: true },
```
akegan: might be a little too late, but what's the value in having a separate style and pillStyle attribute?

app/javascript/components/molecules/LeaseUpSidebar.js
```diff
@@ -0,0 +1,48 @@
...
+    <div className='sidebar-content'>
+      {sidebarButtons(false)}
+      <h3 className='status-history-label'>Status History</h3>
+      <StatusItems statusItems={statusItems} />
+      {sidebarButtons(true)}
+    </div>
+  )
+}
+
+LeaseUpSidebar.propTypes = {
```
akegan, line 65: I didn't realize that we were using propTypes.

akegan (commented):
Just a question about pill styling

app/javascript/components/molecules/LeaseUpSidebar.js
james-wills-sf: @akegan some components use them in partners, some don't. I prefer them but can remove if you don't want to use them. I like it for objects that are passed around, and that you can easily see that the "status" string relates to the list in statusUtils

akegan: yeah they seem helpful! just hadn't seen us using them before. I'd be up for using them more. Would help with eventual conversion to typescript too

app/javascript/utils/statusUtils.js
james-wills-sf: Well one's for a button and ones for a pill, the different classes apply different css attrs. I could calculate them from a single value like `style: 'approved'` and then in status pill use it like ``status-pill-${option.style}`` but I kind of feel like it's more searchable this way.

Can also update the `style` field to be `buttonStyle` to make it more clear

app/javascript/components/molecules/LeaseUpSidebar.js
james-wills-sf: Yep I was thinking the same thing, but didn't want to be too idealistic haha

app/javascript/utils/statusUtils.js
james-wills-sf: I thought this would come up, I'll ask chris about it. We might want it to show as "comment" if multiple history items have the same status in a row.

app/javascript/utils/statusUtils.js
akegan: i was thinking that maybe something like

`.status-pill .is-approved {}` could give the right style, but maybe that's a bad idea. I agree that having it be searchable is helpful

app/javascript/utils/statusUtils.js
akegan: When I brought it up when looking at the designs it sounded like he was fine with just doing what we do now, which is just showing the status and sub-status that's there, but we can re-confirm

james-wills-sf: If you already asked and that's what we show now, that's fine with me! The designs also show duplicate statuses one after the other.

akegan:
Should the pattern include the show/hide functionality, or will we add that when it's implemented in the supp app?

james-wills-sf:
@akegan I knew I was forgetting something! I'll add the show/hide link.

james-wills-sf:
@akegan updated to address your pill styling comment, and I added the show/hide status link

Edit: also there's another pattern library update ready for review: https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/141

showhidestatusupdates

akegan (approved):
Looks good.
Last minor question is whether or not we have the show/hide functionality on small screens. Looks like it's not there now for small screens but I figure that that seems intentional

Last minor comment is that the button seems a bit big for show/hide so the button highlighting when it's active is pretty big 
image

james-wills-sf:
@akegan you're right about the show/hide link. Designs show it in the small screen too. Updated that.

I think I'm going to leave the big button for now. It won't look as huge in the actual sidebar since it doesn't extend the whole width of the screen.

akegan:
cool, good to merge then!

PR #383, opened 2020-08-26
Chores/update standard linting to use Fragment shorthand syntax

james-wills-sf:
I want to be able to use the new `<React.Fragment />` shorthand syntax (`</>`) in Partners, so we can do
```
return (
  <>
    <div />
    <div />
  </>
)
```
Instead of:
```
return (
  <React.Fragment>
    <div />
    <div />
  </React.Fragment>
)
```

We're already on a new enough react/babel version to handle this, but the `babel-eslint` and `standard` libraries we use to lint our code complain about it, so those needed to be updated.

When updating the `standard` library, it also has some new rules that need to be fixed, so I also ran `yarn lint:fix` to fix our code where necessary.

So to sum up the changes in this PR:
1. bump `babel-eslint` and `standard` libraries to the lowest versions that work with the fragment shorthand syntax
2. Ran `yarn lint:fix` to update the code to pass the new code standards
3. Replaced all existing usages of `<Fragment />` or `<React.Fragment />` with the  shorthand syntax (`</>`).


akegan (approved):
Looks good, thanks for the update

PR #384, opened 2020-08-27
Add column layout to supp app page [DAH-36]

akegan:
DAH-36


app/javascript/components/molecules/FormGrid.js
```diff
@@ -26,7 +26,7 @@ FormGrid.Row.defaultProps = {
 
 // NOTE: Grid is hardcoded. This should not be generic element
 FormGrid.Item = ({ children }) => (
-  <div className='form-grid_item small-12 medium-6 large-3 column'>
+  <div className='form-grid_item small-12 medium-12 large-6 column'>
```
akegan, line 29: this impacts the paper application page too but I think it looks OK

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -17,13 +16,23 @@ import LeaseInformationInputs from './sections/LeaseInformationInputs'
...
+
 import validate, { touchAllFields } from '~/utils/form/validations'
 import ParkingInformationInputs from './sections/ParkingInformationInputs'
 import { convertPercentAndCurrency } from '../../utils/form/validations'
+import AsymColumnLayout from '../organisms/AsymColumnLayout'
 
 const StatusUpdateSection = withContext(({ store, formIsValid }) => {
-  const { statusHistory, openUpdateStatusModal, openAddStatusCommentModal, loading } = store
-  let recentStatusUpdate = statusHistory && statusHistory[0] ? statusHistory[0] : { status: null, comment: null, date: null }
+  const {
```
akegan, line 27: this was autoformatted and I thought it does look better this way

app/javascript/components/molecules/FormGrid.js
akegan, line 29: this impacts the paper application page too but I think it looks OK

app/javascript/components/molecules/ContentSection.js
```diff
@@ -9,14 +9,14 @@ const ContentSection = ({ title, description, children }) => (
 )
 
 ContentSection.Header = ({ title, description }) => (
-  <div className='app-inner header-wide'>
+  <div className='app-inner'>
```
akegan, line 12: removing all of these x-wide classes get rid of some pretty heavy left-padding to help the columns look OK. This impacts the paper app too but again I think it's OK

.vscode/settings.json
```diff
@@ -1,3 +1,7 @@
 {
-  "prettier.requireConfig": true
-}
+    "standard.validate": [
```
akegan, line 2: I'm thinking we should probably remove this from being tracked in git unless we want to force folks to use the same linting extensions

akegan (commented):
ended up being a lot of formatting changes but hopefully they're all helpful? I can revert if they'll cause issues. 

I left confirmed household untouched for now so I don't cause conflicts with @software-project 's PR

This PR:
1. Moves most fields on the supp app into their own FormGrid.Row
1. Introduces the AsymColumnLayout pattern with sidebar and main content
1. Updates content section styles to remove some padding
1. Formats a bunch of files. (I ran prettier on them)

.vscode/settings.json
james-wills-sf: Yeah let's add the whole .vscode folder to the gitignore. I think we should be using the same linting rules but we don't need to impose limitations on what extensions are used to highlight + fix them (unless we find that one works best for some reason)

app/javascript/components/molecules/ContentSection.js
```diff
@@ -39,7 +38,7 @@ ContentSection.Content = ({ children, borderBottom, paddingBottomNone, marginTop
   )
 }
 
-ContentSection.Sub = ({ title, description, borderBottom = true, children }) => (
+ContentSection.Sub = ({ title, description, borderBottom = false, children }) => (
```
james-wills-sf, line 41: Not asking you to change anything, but one rule we followed at my last job that I liked is that optional boolean props _always_ default to false. I recommend we make that a general style rule across partners too

using this component for example, the two options for calling it with borderbottom look like:
```
// with a border
<ContentSection.Sub borderBottom />
```
or
```
// without a border
<ContentSection.Sub />
```

When the prop defaults to true your options look like:
```
// with a border
<ContentSection.Sub />
```
or
```
// without a border
<ContentSection.Sub  borderBottom={false} />
```

Which is just eeeever so slightly more noise since you can't use the "prop name defaults to true" shorthand.

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -162,62 +202,105 @@ const SupplementalApplicationContainer = ({ store }) => {
...
-            </div>
-          </form>
-          <StatusModalWrapper
-            {...statusModal}
-            onClose={handleStatusModalClose}
-            onSubmit={(submittedValues) => handleStatusModalSubmit(submittedValues, convertPercentAndCurrency(form.getState().values))}
-          />
-        </>
+              <AsymColumnLayout.Container>
+
```
james-wills-sf, line 264: nit: remove newline

app/javascript/components/supplemental_application/sections/DemographicsInputs.js
```diff
@@ -11,44 +11,57 @@ const DemographicsInputs = () => {
...
               fieldName='number_of_dependents'
               options={numberOfPeopleOptions}
               format={formUtils.formatNumber}
-              label='Number of Dependents' />
+              label='Number of Dependents'
+            />
           </FormGrid.Group>
         </FormGrid.Item>
-
+      </FormGrid.Row>
```
james-wills-sf, line 26: So I'm happy with the `Object.SpecificComponent` naming convention for something that's used less frequently, but it might be nice if we started just using these directly as `<Row />` `<Group />` etc. Since they're supposed to be very generic components I think it's okay that they have short, generic names.

Something to think about, not asking you to make changes here since it's the existing pattern.

james-wills-sf (approved):
@akegan lgtm!

PR #385, opened 2020-08-27
Add .vscode to .gitignore

akegan:
Forgot to make this change in #384 before merging


james-wills-sf:
@akegan lgtm

james-wills-sf (approved):

akegan:
@james-wills-sf So I'm actually thinking now that we should keep settings.json for this workspace in git after all? Especially since we now have recommended settings to get linting to work on save. Folks can override settings in user settings if they want? 

Another suggestion from [this stack overflow post](https://stackoverflow.com/questions/32964920/should-i-commit-the-vscode-folder-to-source-control) is to commit them as default like settings.json.default so folks can copy if they want to

james-wills-sf:
@akegan yeah I think keeping it in the repo is fine, if we run into issues with it we can reconsider later

PR #386, opened 2020-08-27
Features/update status unit tests

james-wills-sf:


spec/javascript/components/molecules/LeaseUpSidebarButtons.test.js
```diff
@@ -1,10 +1,10 @@
 import React from 'react'
-import renderer from 'react-test-renderer'
+import { shallow } from 'enzyme'
 import LeaseUpSidebarButtons from 'components/molecules/LeaseUpSidebarButtons'
 
 describe('LeaseUpSidebarButtons', () => {
   test('should render correctly with mobile styling', () => {
-    const wrapper = renderer.create(
+    const wrapper = shallow(
```
james-wills-sf, line 7: Instead of using `react-test-renderer`'s `render.create` to make wrappers, use [shallow](https://enzymejs.github.io/enzyme/docs/api/shallow.html) instead. 

The react-test-renderer create method deeply renders a component, meaning it renders the component's elements and all of its children's components elements. This makes the snapshots huge and hard to read, and means our unit tests for wrapper components are actually testing every child component they contain.

Shallow only renders one level deep, so say you're writing a test for MyComponent:

```
const MyChildComponent = (someProp) => (
<div>
  <div className="div3">
    <div className="div4" />
  </div>
  <div className="div5">
    <div className="div6" />
  </div>
</div>
)

const MyComponent = () => (
<div>
  <div className="div1">
    <div className="div2" />
  </div>
  <MyChildComponent someProp={true} />
  <MyChildComponent someProp={false} />
</div>
)
```

The shallow snapshot would look like:
```
<div>
  <div className="div1">
    <div className="div2" />
  </div>
  <MyChildComponent someProp={true} />
  <MyChildComponent someProp={false} />
</div>
```
and the renderer.create snapshot would look like:
```
<div>
  <div className="div1">
    <div className="div2" />
  </div>
  <div>
  <div className="div3">
    <div className="div4" />
  </div>
  <div className="div5">
    <div className="div6" />
  </div>
  </div>
  <div>
    <div className="div3">
      <div className="div4" />
    </div>
    <div className="div5">
      <div className="div6" />
    </div>
  </div>
</div>
```

spec/javascript/components/molecules/StatusItems.test.js
```diff
@@ -1,43 +1,139 @@
...
   })
 
-  test('should render multiple status items correctly', () => {
-    const wrapper = getWrapper(mockStatusItems())
-    expect(wrapper).toMatchSnapshot()
+  describe('with a single item in the status items list', () => {
+    const statusItems = [mockStatusItems()]
+
+    test('should render one status item when no limit is specified', () => {
+      expect(getWrapper(statusItems).find(StatusItem)).toHaveLength(1)
```
james-wills-sf, line 74: `shallow` wrappers also let us do queries using components, instead of complicated selectors with only root elements and classnames.

spec/javascript/components/molecules/StatusItems.test.js
```diff
@@ -1,43 +1,139 @@
...
 )
 
+const TIMESTAMP_AUGUST_25_2020 = 1598338800
+const TIMESTAMP_AUGUST_26_2020 = 1598400000
+
 describe('StatusItems', () => {
-  test('should render empty status items correctly', () => {
-    const wrapper = getWrapper([])
-    expect(wrapper).toMatchSnapshot()
+  describe('snapshot tests', () => {
```
james-wills-sf, line 18: Snapshot tests can still sometimes be useful, but our main tests should be specific unit tests. This file has some snapshot and some unit tests

app/javascript/components/molecules/StatusItems.js
```diff
@@ -1,46 +1,9 @@
...
-
-const StatusDate = ({ timeStamp }) => (
-  <div className='status-item-date'>
-    {moment.unix(timeStamp).format('MMMM Do YYYY')}
-  </div>
-)
-
-StatusDate.propTypes = { timeStamp: PropTypes.number }
-
-const StatusItem = ({ statusItem }) => {
```
james-wills-sf, line 18: One issue with shallow rendering is if we declare multiple components in a single file and only export one of them, the shallow rendering won't actually be testing the child components. If you want the children to be tested, you should either move the child into its own file or add an `export` to the child component definition that's only used by the spec files.

spec/javascript/components/atoms/StatusPill.test.js
```diff
@@ -0,0 +1,37 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import StatusPill from 'components/atoms/StatusPill'
+
+const getWrapper = (status) => shallow(<StatusPill status={status} />)
+
+describe('StatusPill', () => {
+  test('should render properly with approved status', () => {
```
james-wills-sf, line 8: With sufficiently simple components I think snapshot tests do the job fine.

for example this snapshot just looks like this:
```
<div
  className="status-pill is-appealed"
>
  Appealed
</div>
```

That's pretty readable and not much different from a test that specifically tests if the `is-appealed` class is applied

akegan (approved):
good improvements!

PR #387, opened 2020-08-28
Finalize layout of supp app DAH-36

akegan:
DAH-36

This PR:
- Rearranges fields to better match the design mock-up
- Fixes styling issues with the confirmed preferences table (changes overflow to scroll and stops sthe expanded panel from being so wide
- Refactors currency and percentage fields to be based on InputField
- Updates padding between fields and sections
- Updated h3 attributes to bold and form notes to tiny + not-bold. 

Current status at large screen:
wide_screen_sup_app

and at small screen:
small_screen_sup_app

Still TODO:
- Final copy changes
- Padding between status update and this update
- Follow-up tickets on styling of checkbox labels (make them not-bold and properly aligned)
- Follow-up ticket to stop the preference table from having overflow


app/javascript/components/molecules/InlineModal.js
```diff
@@ -0,0 +1,11 @@
+import React from 'react'
```
akegan, line 1: this is ExpandablePanel.js just renamed.

app/javascript/utils/form/final_form/Field.js
```diff
@@ -24,14 +24,16 @@ export const BlockNote = ({ value }) => (
   <span className='checkbox-block_note no-margin padding-left--half'>{value}</span>
 )
 
-export const Input = ({ input, id, meta, type, maxLength, placeholder, ariaLabelledby, fieldName }) => (
+export const Input = ({ input, id, meta, type, maxLength, placeholder, ariaLabelledby, fieldName, disabled }) => (
```
james-wills-sf: nit: with this many items it might be good to split this out onto multiple lines

app/javascript/utils/form/final_form/Field.js
```diff
@@ -65,74 +92,91 @@ export const InputField = ({ fieldName, label, blockNote, validation, placeholde
...
 
 /**
  * Field that allows only whole integer percent values.
  */
-export const PercentField = ({ fieldName, validation, id, label, placeholder, maxLength, disabled, isDirty = true }) => (
-  <Field
-    name={fieldName}
-    validate={validation}
-    component='input'
+export const PercentField = ({ fieldName,
```
james-wills-sf: nit: move fieldname to the next line

app/javascript/utils/form/final_form/Field.js
```diff
@@ -220,7 +265,7 @@ export const TextCheckboxField = ({ fieldName, label, blockNote, validation, id,
             meta={meta}
             aria-labelledby={ariaLabelledby}
             id={id || `form-base-${fieldName}`} />
-          <input id={id || `form-ctrl-${fieldName}`} fieldName={fieldName} type='checkbox' checked={input.value === trueValue} onClick={onChange} />
```
james-wills-sf, line 223: Why remove the fieldName prop here?

james-wills-sf (commented):
@akegan did a first pass, I might need to take another look on monday. I also looked at the review app and all of the spacing seems good

app/javascript/utils/form/final_form/Field.js
akegan: it was actually causing console log errors! not sure where it came from or when they started showing up

app/javascript/components/atoms/TableWrapper.js
```diff
@@ -2,8 +2,8 @@ import React from 'react'
 
 const TableWrapper = ({ children }) => (
   <div className='form-grid row expand'>
-    <div className='small-12 column'>
-      <div className='scrollable-table-container-under-xlarge'>
+    <div className='small-12 medium-12 large-12 column'>
```
james-wills-sf: nit: isn't `small-12` the same as `small 12 medium-12 large-12`? I think the common practice is to stop at the smallest relevant size

app/javascript/components/molecules/InlineModal.js
```diff
@@ -0,0 +1,11 @@
...
+
+const InlineModal = ({ children }) => {
+  return (
```
james-wills-sf: since this is just a rename no need to change it, but how do you feel about defaulting to the expression syntax when there's no additional logic in the render body?

```
const InlineModal = ({ children }) => (
  <div className='app-editable padding-left--2x padding-right--2x'>
    {children}
  </div>
)
```

I think sometime I might migrate us off of standard so we can specify our own linting rules. We can start by using the standard eslint config which has all of the rules in `standard` but that way we can add or ignore as we want

app/javascript/utils/form/final_form/Field.js
```diff
@@ -24,33 +24,70 @@ export const BlockNote = ({ value }) => (
...
+    disabled={disabled}
+  />
 )
 
-export const Label = ({ label, fieldName, blockNote, id, labelId, className }) => {
+export const Label = ({ label, fieldName, blockNote, id, labelId }) => {
   return (
     label ? (
-      <label htmlFor={id || `form-${fieldName}`} id={(labelId || `label-${fieldName}`).replace('.', '-')} className={className || 'form-label'}>
+      <label htmlFor={id || `form-${fieldName}`} id={(labelId || `label-${fieldName}`).replace('.', '-')} className='form-label'>
```
james-wills-sf: nit: this is a pretty complicated line, can we move each prop onto its own line?

james-wills-sf (approved):
@akegan added a couple nit comments that you can take or leave, lgtm!

PR #388, opened 2020-08-31
Replace status buttons with rev3 sidebar [DAH-37]

james-wills-sf:
Ticket: DAH-37

- add Sidebar to Supp app page
- remoe old status history section from supp app
- delete now-unused status history components
- Update the sidebar slightly so the props match more closely to the names of the existing statushistory data
- Add loading states to the sidebar
- make the save button an actual save button by adding an id and type to the button element


app/javascript/components/organisms/AsymColumnLayout.js
```diff
@@ -18,7 +21,7 @@ AsymColumnLayout.Container = ({ children }) => {
 
 AsymColumnLayout.MainContent = ({ children }) => {
   const classes = classNames(
-    'small-12',
+    'medium-12',
```
james-wills-sf, line 24: So my sidebar switches to the mobile view at medium width instead of small because it gets really tight at medium, this is something I still need to talk to chris about

james-wills-sf:
Related style PR: https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/143

app/javascript/components/molecules/LeaseUpSidebar.js
```diff
@@ -6,30 +6,42 @@ import LeaseUpSidebarButtons from './LeaseUpSidebarButtons'
...
 const ShowOrHideStatusesButton = ({ showingAllStatuses, onClick }) => {
   const buttonClasses = classNames(
     'button-link',
     't-tiny',
     'show-all-updates-toggle'
   )
 
+  const onClickPreventDefault = (event) => {
+    onClick()
+    event.preventDefault()
```
akegan: what does prevent default do?

spec/javascript/components/molecules/__snapshots__/LeaseUpSidebar.test.js.snap
```diff
@@ -55,10 +57,11 @@ exports[`LeaseUpSidebar should render with empty status items correctly 1`] = `
     className="show-large-up"
   >
     <LeaseUpSidebarButtons
+      isLoading={false}
       onAddCommentClicked={null}
       onChangeStatus={null}
       onSaveClicked={null}
-      status="Approved"
+      status={null}
```
akegan: hm this is a bit worrisome

james-wills-sf:
@akegan I'm still tweaking the e2e tests but this should be good to review

akegan (commented):
couple of questions. 1st round of QA looked good too. Will QA now after merge to make sure things still look/work OK

app/javascript/components/molecules/LeaseUpSidebar.js
james-wills-sf: For some reason this button was behaving as if `type='submit'` was set, so the form was trying to save on click. Prevent default stops the event from bubbling up past that point, but it was actually the wrong way to handle this.

I updated it to explicitly set `type='button'`, now we don't need to stop any event propagation.

spec/javascript/components/molecules/__snapshots__/LeaseUpSidebar.test.js.snap
james-wills-sf: Ah this is because we aren't explicitly passing in currentStatus anymore, we infer it from the status items that are passed in.

I updated the test file to expect null in the cast where no statusItems are present (my understanding is this shouldn't happen in the real world) and to follow the testing standard we came up with. Now there is only one snapshot test, and otherwise we manually test the features for that component.

akegan (approved):
QA looks good. Found a few minor things that we can tweak in this PR or in our follow-up polish pass:
1. Date format is slightly off. Figma has `Aug 2, 2020`, this PR has `August 31st 2020`. 
1. "Status History" is sans serif in the design. Can match it just by adding the t-sans class. 
1. Padding might be slightly off in the status history button, which may be leading to the 2-lines issue. I'll check in with Chris on it later but my guess is that the best solution is to just make the max width >74rem as we discussed

akegan:
one other small note is that if you have a long comment then you have a scroll even before you expand to all the status comments. not sure if that's the expected behavior. I'll ask chris about that too.

james-wills-sf:
@akegan to your second comment, yes that is the expected behavior, Basically if the sidebar ever extends past 100% of view height we have to add a scroll bar, otherwise the bottom content will be inaccessible. We could add a check to change the limit dynamically based on how many comments fit in the viewport, but that would require hacky js measurements

akegan:
@james-wills-sf FYI I made a couple of updates after doing a close check with figma and talking to Chris:

1. I updated the date formatting and added a test to check for the date format. Had to make another tweak to how you were generating sample timestamps to get the tests to work
1. I updated the style for "Status History" (see PR https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/144)
1. I updated the max width to 80rem with the 👍  from Chris

Otherwise Chris said things look good! I'm going to branch off of this branch to get started on the final copy updates.

app/javascript/components/organisms/AsymColumnLayout.js
akegan: chris was OK with this

akegan:
I'm working on figuring out why e2e tests are failing now

spec/javascript/components/molecules/StatusItem.test.js
```diff
@@ -50,4 +50,10 @@ describe('StatusItem', () => {
     const wrapper = getWrapper(mockStatusItem({ status: 'Lease Signed', substatus: null }))
     expect(wrapper).toMatchSnapshot()
   })
+
+  test('should format timestamps as expected', () => {
+    const wrapper = mount(<StatusItem statusItem={mockStatusItem({ status: 'Lease Signed', substatus: null })} />)
+    const timestamp = wrapper.find('StatusDate').text()
```
james-wills-sf, line 56: I'm surprised that you can search for a component name using mount! I guess I thought it would all be div and class selectors

james-wills-sf (commented):
@akegan reviewed your 3 commits and your changes lgtm

PR #393, opened 2020-09-02
Fix status header styling DAH-37

akegan:
DAH-37

For large screens: 
image

For small screens:
image


app/javascript/components/molecules/LeaseUpSidebar.js
```diff
@@ -54,10 +55,20 @@ const LeaseUpSidebar = ({ isLoading, statusItems, onSaveClicked, onChangeStatus,
 
   const numberOfStatusesToDisplay = showingAllStatuses ? statusItems.length : MAX_UPDATES_TO_SHOW_DEFAULT
 
+  const header = (title) => (
+    <>
+      <div className='padding-top--3x show-large-up'>
```
akegan, line 60: used padding here instead of margin just because we don't currently margins only go up to 3x. Can add it though if this should be a margin instead of padding

akegan (commented):
I fiddled with this a bit while in the 8cloud data meeting and this is what I came up with. Let me know if you have any other ideas/suggestions for switching between Header and SubHeader for small + large. 

This would allow us to get rid of the status-history-label class in partners totally too.

app/javascript/components/molecules/LeaseUpSidebar.js
james-wills-sf: not a big deal to me for margin vs padding.

james-wills-sf (approved):
@akegan lgtm!

PR #394, opened 2020-09-03
Update lease and rental assistance styling DAH-92

akegan:
DAH-92


akegan:
Current Status:
- Lease fields moved into inline modal and rental assistance moved into 2nd level of inline modal
- Lease + rental assistance fields moved around
- Content for lease updated

Left to do:

- [x] Add save + delete button for lease
- [x] Add state to show/hide the lease section + the description at the top of the lease section
- [x] Add create lease button
- [ ] Remove borders on preference + rental assistance tables - will follow up in DAH-137
- [ ] More yes/no selectors need to be moved to checkboxes -> might be a separate ticket, especially for parking where the parking cost needs to pop up correctly. - will follow up in DAH-138


Lease section: 
image
With new rental assistance panel open: 
image
With rental assistance edit: 
image

app/javascript/components/atoms/Button.js
```diff
@@ -1,15 +1,30 @@
 import React from 'react'
 import classNames from 'classnames'
+import PropTypes from 'prop-types'
 
-const Button = ({ text, small, tiny, tertiary, classes, ...rest }) => {
+const Button = ({ type, text, small, tiny, tertiary, classes, ...rest }) => {
```
akegan: I'd like to add primary, secondary, alert-fill, etc in here too and drop the classes attribute but this PR was already big enough.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -175,6 +176,15 @@ class SupplementalApplicationPage extends React.Component {
     }
   }
 
+  handleClickCreateLease = () => {
+    this.setState({ showLeaseSection: true })
+  }
+
+  handleClickCancelLease = () => {
```
akegan: Fine with renaming these too. wanted to make it clear for now that all that's going on is that we're clicking the button. Going to save pretty much all functionality for when it's wired up.

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -40,6 +40,7 @@ class SupplementalApplicationPage extends React.Component {
     leaveConfirmationModal: {
       isOpen: false
     },
+    showLeaseSection: true,
```
akegan, line 44: updated to default to showing the lease section to make unit tests work for now. I want to update the unit tests to test for the interactive elements on the lease once we get the interactions hammered out

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -0,0 +1,168 @@
+import React from 'react'
```
akegan, line 1: This was in fact renamed from LeaseInformationInputs, but I changed the file so much that even using git mv didn't work ☹️.

app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
```diff
@@ -8,7 +8,7 @@ const monthlyRentFieldName = 'lease.monthly_parking_rent'
 const Yes = 'Yes'
 const No = 'No'
 
-const ParkingInformationInputs = ({ form: { change }, values: { lease }, visited }) => {
+const ParkingInformationInputs = ({ form: { change, getFieldState }, values: { lease } }) => {
```
akegan, line 11: I removed "visited" in a few different places where we were already passing the form to reduce the number of fields that we were passing around. It seems like the visited object might be nice to pass when we're not passing form, but otherwise is redundant because you can get whether or not a field was visited from form.getFieldState(fieldName)

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -165,85 +169,94 @@ const RentalAssistanceForm = ({ values, onSave, loading, onClose, applicationMem
...
+              <FormGrid.Item>
+                <InputField
+                  label='Other Assistance Name'
+                  fieldName={`rental_assistances.${index}.other_assistance_name`} />
+              </FormGrid.Item>
+            </FormGrid.Row>
+          )}
+          <FormGrid.Row>
+            <div className='form-grid_item column'>
+              <Button
```
akegan: Updating to Button here didn't change functionality, I just want to be more consistent about using our pattern lib components.

akegan (commented):
Couple of explanatory comments on the current state of this PR. Looks like e2e tests are failing so I'll fix those then I think this'll be ready for review

akegan:
@cpalmatier can you take a look at these changes in a design pass to see if things look OK? The functionality of actually creating and saving a lease isn't there, but the look should be.

Review app: https://dahlia-lap-full-pr-394.herokuapp.com/
You can see the different states of the lease showing + hiding by clicking "cancel" on the lease on load, then "create lease" to go back and forth.

Known design tweaks that we're not going to make in this PR:
1. A couple more cases of wanting to switch yes/no select field or radio buttons to checkboxes for recurring assistance and BMR parking DAH-138
1. Border on the bottom of preference and rental assistance tables. DAH-137

akegan:
alright @james-wills-sf I think this is ready for review. See above for comments on functionality and known design issues. Designs are in [this figma file](https://www.figma.com/file/qjUryiHN6RGwMMretWbExa/explicit-lease-creation-164886190?node-id=1%3A964)

app/javascript/components/atoms/Button.js
```diff
@@ -1,15 +1,30 @@
...
+      type={type}
+      {...rest}
+    >
+      {text}
+    </button>
   )
 }
 
+Button.propTypes = {
+  type: PropTypes.oneOf(['button', 'submit'])
```
james-wills-sf: 👍

app/javascript/components/atoms/Button.js
```diff
@@ -1,15 +1,30 @@
...
+}
+
+Button.defaultProps = {
```
james-wills-sf, line 41: so I used defaultProps in my other pr but it seems like the new way to do it is directly in the definition?

```
 const Button = ({ type = 'button', text, small, tiny, tertiary, classes, ...rest }) => { 
```

Whichever way is fine by me, just letting you know I think they do the same thing

app/javascript/components/atoms/Button.js
```diff
@@ -1,15 +1,30 @@
...
     'small': small,
     'tertiary': tertiary,
     'tiny': tiny
   })
   return (
-    <button className={btnClassNames} {...rest}>{text}</button>
+    <button
+      className={btnClassNames}
+      type={type}
+      {...rest}
```
james-wills-sf, line 29: Definitely don't change it in this PR but just letting you know this is generally a react anti-pattern to allow any additional props to be passed through a component.

app/javascript/components/molecules/ExpandableTable.js
```diff
@@ -39,7 +39,7 @@ const ExpandableTableRow = ({ row, rowKeyIndex, numColumns, expanderRenderer, ex
 }
 
 const ExpandableTable = ({ columns, rows, rowKeyIndex, expanderRenderer, expandedRowRenderer, originals, classes, closeAllRows }) => (
-  <table className={classNames('td-light td-plain th-plain', classes)} role='grid'>
+  <table className={classNames('td-light td-plain th-plain', classes)} style={{ background: 'transparent' }} role='grid'>
```
james-wills-sf: nit: can you move these props onto separate lines?

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -70,26 +69,18 @@ const Income = ({ listingAmiCharts, visited, form }) => (
...
-    >
-      <ParkingInformationInputs form={form} values={values} visited={visited} />
-    </ContentSection.Sub>
-    <ContentSection.Sub
-      title='Rental Assistance Information'
-      description='Rental Assistance includes recurring vouchers and subsidies, as well as one-time grants and other assistance.'
-    >
-      <RentalAssistance form={form} submitting={submitting} />
-    </ContentSection.Sub>
+    { showLeaseSection ? <Lease
```
james-wills-sf: nit: I'd move Lease onto it's own line and wrap the whole thing in parentheses

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
james-wills-sf: can you add a TODO comment here to remove this line?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
james-wills-sf: nit: Remember when I said the new version of Standard has weird rules around `handle*` and `on*`naming? This reminded me I didn't really understand it so I googled it and found this: https://jaketrent.com/post/naming-event-handlers-react/

Basically the function should be named `handle*` and the props should be `on*`. I think that kind of makes sense and reads nicely in the component:
```
<button onClick={handleButtonClick} />
```

So in this case we'd keep the function name and the context variable `handleClickCancelLease`/`handleClickCreateLease`, but the prop names would change.

It looks like we're already using `onCancel` for the cancel case so that's good, but maybe we can change the `LeaseSection` prop to `onClickCreate` or something?

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -0,0 +1,168 @@
...
+  })
+  const confirmedPreferenceOptions = formUtils.toOptions(
+    map(
+      [{ id: null, preference_name: 'None' }, ...confirmedPreferences],
+      pluck('id', 'preference_name')
+    )
+  )
+
+  const getVisited = (fieldName) => (
+    form.getFieldState(fieldName)?.visited
```
james-wills-sf, line 80: I totally thought `?.` was only a typescript thing, that's great

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -165,85 +169,94 @@ const RentalAssistanceForm = ({ values, onSave, loading, onClose, applicationMem
...
+            </FormGrid.Item>
+          </FormGrid.Row>
+          <FormGrid.Row>
             <FormGrid.Item>
-              <InputField
-                label='Other Assistance Name'
-                fieldName={`rental_assistances.${index}.other_assistance_name`} />
+              <SelectField
+                label='Type of Assistance'
+                fieldName={`rental_assistances.${index}.type_of_assistance`}
```
james-wills-sf: I'm seeing `rental_assistances.${index}.*` a lot. I wonder if this would be better pulled out as a simple function like:

```
const getField = (assistanceField) => `rental_assistances.${index}.${assistanceField}`
```

and then this line could be `fieldName={getField('type_of_assistance')}`

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -165,85 +169,94 @@ const RentalAssistanceForm = ({ values, onSave, loading, onClose, applicationMem
...
+                isDirty={visited && visited[`rental_assistances.${index}.assistance_amount`]}
+              />
+            </FormGrid.Item>
+          </FormGrid.Row>
+          <FormGrid.Row>
+            <FormGrid.Item>
+              <YesNoRadioGroup
+                label='Recurring Assistance'
+                fieldName={`rental_assistances.${index}.recurring_assistance`}
+                uniqId={(!isEmpty(values) && values.id) || 'new'}
```
james-wills-sf: could this be simplified to `values?.id || 'new'`?

Can you tell I'm excited that I just learned about the `?.` thing?

app/javascript/components/supplemental_application/sections/RentalAssistance.js
```diff
@@ -165,85 +169,94 @@ const RentalAssistanceForm = ({ values, onSave, loading, onClose, applicationMem
...
+                trueValue='Yes'
```
james-wills-sf: I think these are the defaults. You're welcome to keep them if you'd prefer to be explicit though

james-wills-sf (commented):
@akegan added a few nit comments

I tested in the review app and it looks great!

couple design things: I think the rental assistance inline modal is supposed to go all the way across, in the review app it's pretty small:

### review app


### designs
[figma](https://www.figma.com/file/qjUryiHN6RGwMMretWbExa/explicit-lease-creation-164886190?node-id=0%3A1)


Couple of small things:
- It looks like there's an extra 16px of vertical padding around the lease modal (and the rental assistances modal)  due to the button padding on each
- left margin on rental assistance panel + table is smaller than in the design
- margin above "tenant contribution" header is a little smaller than design when rental assistance panel is opened.

app/javascript/components/molecules/ExpandableTable.js
akegan: do we want to have a standard # of props at which we move to their own lines? in theory our linter would complain about line length.

app/javascript/components/supplemental_application/sections/Lease.js
akegan: I feel like I try in every environment I'm in and am always thrilled when it exists.

app/javascript/components/supplemental_application/sections/RentalAssistance.js
akegan: OK So I googled it because you made me nervous that it's maybe not a thing

It looks like it "has not yet reached cross-browser stability" 😭  and until v80 of chrome even you had to enable it with a flag. Look at [implementation progress on this page](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)

Not really sure what that means for us since everything is being minified anyways? Not sure if we have to test on browserstack or something

app/javascript/components/supplemental_application/sections/RentalAssistance.js
akegan: actually looks like true/falseLabel are yes/no but not true/falseValue. Snapshots caught something helpful for once!

app/javascript/components/atoms/Button.js
akegan: I made this change then type="button" was removed from snapshots 🤔. I reverted the change but maybe should investigate more

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: I just went to update then got confused. What happens when it's a function that you're passing as a prop down many levels? Am I just overthinking it?

james-wills-sf: @akegan yeah, I thought the same thing. I think only the top level would be handle-, all of the props are on-

Basically, if it's a function, it should be called handle-, if it's a prop it should be on-

(at least according to what that one blog post said that we can totally ignore if you want to)

app/javascript/components/atoms/Button.js
james-wills-sf: okay then don't worry about it for now. I'll try it next time I write a component and get back to you

app/javascript/components/molecules/ExpandableTable.js
james-wills-sf: Airbnb styleguide says more than 1 prop always wraps to a new line which I think is a little aggressive.

I think line length is is a good way to do it but it doesn't seem like standard.js complains about long lines, so that's something we can add when we get around to extending the standard linter.

If you want a hard and fast rule so you don't have to think about it maybe we say 3 props? Otherwise I just look at how complex the line is and decide that way 🤷

app/javascript/components/supplemental_application/sections/RentalAssistance.js
james-wills-sf: Looks like we have `@babel/plugin-proposal-optional-chaining` in our `yarn.lock` so it should be compiled to something compatible with all browsers.

We may want to explicitly add plugin-proposal-optional-chaining to our package.json file so that we know we'll always have it

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
akegan: OK just pushed changes based on the article. Also renamed to put click at the end because i liked that suggestion too. Let me know if that's what you were thinking! Still feels kinda weird and inconsistent.

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -155,7 +155,7 @@ const Lease = ({ form, submitting, values, store }) => {
         {/* TODO: Wire up actions for buttons, set to disabled when loading */}
         <LeaseActions
           onSave={handleSaveLease}
-          onCancel={handleClickCancelLease}
+          onCancelLeaseClick={handleCancelLeaseClick}
```
james-wills-sf, line 159: So I didn't mean to say you had to change this one but it's good either way!

james-wills-sf (approved):
@akegan lgtm

app/javascript/components/supplemental_application/sections/RentalAssistance.js
akegan: looks like it's ultimately coming from webpack so we should be all good. Thanks for checking!

as a bonus it looks like we get [https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_operator]("nullish coalescing") too, which is like a null-aware ternary alternative. `const foo = null ?? 'default string';`  gives you 'default string'

akegan:
@james-wills-sf I just pushed up some changes that should fix the tests if you want to take a minute to review my latest changes before merge.

spec/javascript/e2e/SupplementalApplicationPage/rental_assistance.e2e.js
```diff
@@ -177,9 +148,37 @@ describe('SupplementalApplicationPage Rental Assistance Information section', ()
...
+  // This test requires a rental assistance to be already present in the rental
+  // assistances table. If the prior test for creating a rental assistance has
+  // succeeded, then there will be at least one rental assistance present.
+  test('should allow a rental assistance to be deleted', async () => {
+    let { page } = await SetupBrowserAndPage(testBrowser)
+
+    await sharedSteps.goto(page, `/applications/${LEASE_UP_LISTING_APPLICATION_ID}/supplementals`)
+
+    // Record how many rental assistances are in the table before we attempt our delete
+    const prevTableSize = await page.$$eval('.rental-assistances > tbody > .tr-expand', elems => elems.length)
```
james-wills-sf, line 163: nit: is the double dollarsign important in `$$eval`?

james-wills-sf (approved):
@akegan lgtm!

PR #395, opened 2020-09-03
Clean up supplemental application actions [DAH-93]

james-wills-sf:
Ticket: [DAH-93]

Clean up some of the lease and application logic to make it easier to modify once we add the new lease interactions.

### proposed guidelines for frontend API code
1. Move as much logic as possible out of components into actions
  a. like any kind of data manipulation should happen in the actions
1. Component API logic should be limited to triggering actions and handling success/error responses
1. Components should never call apiService methods directly 
1. Use promise .then().catch().finally() handlers where possible instead of `await`.
1. mark a function as `async` when it returns a promise
  a. This means almost all actions or apiservice functions should be `async`
  b. `async` will also automatically wrap any return value that isn't a promise. But it won't double wrap a promise.
1. Bubble up errors instead of returning response = false or something


app/javascript/apiService.js
```diff
@@ -107,20 +107,40 @@ const deleteRentalAssistance = async (rentalAssistanceId) => {
...
-  if (!leaseToCreateOrUpdate['lease_start_date']) {
-    leaseToCreateOrUpdate['lease_start_date'] = {}
+const getLeaseRequestData = (rawLeaseObject, primaryApplicantContact) => {
+  return {
+    lease: {
+      ...rawLeaseObject,
+      // TODO: We should consider setting the Tenant on a Lease more explicitly
+      // either via a non-interactable form element or using Salesforce
+      primary_applicant_contact: primaryApplicantContact,
+      // TODO: determine why lease_start_date gets removed
```
akegan: I think we had an issue where with the way that we were processing the dates with multiple fields that when leasing agents cleared all of the lease start date fields we weren't actually deleting it on Salesforce

app/javascript/apiService.js
```diff
@@ -107,20 +107,40 @@ const deleteRentalAssistance = async (rentalAssistanceId) => {
...
+
+  const data = getLeaseRequestData(leaseToUpdate, primaryApplicantContact)
+
+  const leaseId = leaseToUpdate['id']
+  return request.put(`/applications/${applicationId}/leases/${leaseId}`, data)
+    .then(response => response.lease)
+}
+
+export const createLease = async (leaseToCreate, primaryApplicantContact, applicationId) => {
+  if (isLeaseAlreadyCreated(leaseToCreate)) {
```
akegan, line 124: this seems like a good update. We'll also have to do the check on the backend since it's more likely to have up-to-date info about whether or not there's a lease.

akegan (commented):
Do you know what the 12k lines are that were removed from the supp app snapshot? too many lines to load in github. 

Otherwise looks good, just need to do some QA

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -378,7 +402,6 @@ describe('SupplementalApplicationPage', () => {
       // of how the dropdown button and dropdown menu should render
       // when the dropdown menu is open
       expect(wrapper.find('.sidebar-content')).toHaveLength(1)
-      expect(wrapper).toMatchSnapshot()
```
james-wills-sf, line 381: @akegan this is what caused the 12k deleted lines.

This snapshot is so huge it's essentially unreadable, and after I fixed these tests to all be independent it actually added 15k lines. Since this test is really only supposed to test the existence of the sidebar i figured we can just keep the explicit test and get rid of the huge snapshot.. We have enough other snapshot tests in this file I'm confident we don't need it.

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -81,12 +101,12 @@ describe('SupplementalApplicationPage', () => {
   })
 
   test('it doesnt send field that are not changed', async () => {
-    const payload = supplementalApplication
```
james-wills-sf, line 84: Instead of re-using the same object, we make sure to clone it each time. This combined with the response object being manipulated was causing the tests to not be independent -- the first tests were modifying the object, and the later tests were using the modified object instead of the original fixture.

akegan:
I tried updating via the status dropdown and got this error: 
image

Seems like there may be a related e2e test problem too

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
akegan: ok great. Yeah I guess the snapshots are essentially useless when github doesn't even show the diff!

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
akegan: OK great! We were doing it in come cases but looks like we weren't being consistent.

james-wills-sf:
@akegan looks like I was just missing a return. Status message dropdown should work now. I'm about to add tests

app/javascript/apiService.js
james-wills-sf: Do you mind helping me come up with a better way to phrase this comment?

james-wills-sf (commented):
@akegan this should be ready for review now!

app/javascript/apiService.js
akegan: How about:
> TODO: Move removing empty lease start dates to the date component.

akegan:
@james-wills-sf couple of frontend tests are failing. Doing QA now

james-wills-sf:
@akegan oh that was because I put `async` on a function that just returned an object. That will definitely break saving leases, so make sure you QA on the latest release app

Edit: this is fixed now

akegan:
OK I've confirmed:
1. That a new lease is still created successfully on save if there wasn't a lease before
1. That if you go to a new application without a lease and have a rental assistance panel open with some stuff then hit save on the full application that you only get one lease. 
1. That least start date can still be removed
1. That if you create and save a rental assistance, then save the full application that you still just get one lease. 
1. Other application updates work
1. Also tested that updating a failed application with the status modal does cause the nice red error message to pop up

Anything else I should test @james-wills-sf ?

james-wills-sf:
@akegan that all sounds good, wrote this in slack but the only additional things I can think of are:
- submitting with an empty comment
- submitting a new status and verify it changes
- submitting with lease/rental assistance changed/unchanged
- submitting from the applications table
- If you know of any other pages that use these actions let me know so I can check those too

If all of those are covered then you're good!

akegan (approved):
did those last steps of QA and it looks good. Thanks for moving us in a good direction!

PR #397, opened 2020-09-04
[rev3] Fix empty status items behavior

james-wills-sf:
This PR makes it to that applications with no status history don't show the "status history" header at all.


akegan (approved):
Confirmed that it looks better without any status items present: 
image

and looks good in cases with status history too 
image

e2e tests are failing though

PR #398, opened 2020-09-04
Split out sidebar components

james-wills-sf:
While updating the LeaseUpSidebar component for https://github.com/SFDigitalServices/sf-dahlia-lap/pull/397, I realized the LeaseUpSidebar container was getting too complicated, and that our `const renderX()` pattern isn't very good.

This change looks like a lot of lines but it's really just shuffling things around

- move sidebar related components into lease_up_sidebar folder
- instead of LeaseUpSidebarButtons handling swapping the save and status buttons, do this in the container and change teh component to be LeaseUpStatusButtons
- Split out header, show/hide entire status item list behavior into the StatusHistoryContainer component
- added some additional tests to the StatusHistoryContainer component


james-wills-sf:
Accidentally closed this after updating base branch

akegan (approved):
looks reasonable and cleaner

PR #400, opened 2020-09-09
Add view lease state [DAH-93]

james-wills-sf:
- Convert the lease section state from a boolean into a string with 3 values: 'show_lease', 'no_lease', and 'edit_lease
- Set the initial state as either no_lease or show_lease depending on whether a lease exists or not
- Add a button to edit the lease when in show_lease state
- When in show lease state, disable all of the inputs

create edit delete

### DIFF
--- a/app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
+++ b/app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
@@ -18,6 +18,7 @@ import validate, { touchAllFields } from '~/utils/form/validations'
 import { convertPercentAndCurrency } from '../../utils/form/validations'
 import AsymColumnLayout from '../organisms/AsymColumnLayout'
 import LeaseUpSidebar from '../molecules/lease_up_sidebar/LeaseUpSidebar'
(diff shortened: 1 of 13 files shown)


akegan:
Couple of small QA notes:
1. Without create lease button or table, the label is squished 
image

Rental assistance spacing label looks better when there is a rental assistance table there but still may need to be bumped down: 
image

2. If there's no units assigned looks like disabled state shows "no units available" even when there are units available when you click edit lease
image

Update: It shows no unit available even when there is a unit assigned to the lease 
image
Which is this application: https://dahlia-lap-full-pr-400.herokuapp.com/applications/a0o0P00000Isnf4QAB/supplementals

james-wills-sf:
@akegan the two qa issues should be fixed now

- 2020-09-10 at 8 56 52 AM
- 2020-09-10 at 8 56 37 AM

- 2020-09-10 at 9 12 48 AM

akegan (approved):
Looks better!

Form note is a bit separated from the button now when the button is present, but I think it's an OK solution 
image

PR #401, opened 2020-09-10
Add unit test utils and examples for forms

james-wills-sf:
Allow testing via shallow rendering in components that use form or context.

For components that take a form object, render it using the new `withForm` util. The util will wrap it in a form object, and handle all of the complicated shallow rendering `dive()`s so you only have to deal with the component itself, shallow rendered

This change
- adds a file `wrapperUtil` that adds some utilities for dealing with shallow wrappers
- changes the RentalAssistance test file to follow our new testing rules
  - shallow rendering by default (only one test needed mount rendering)
  - assert on specific things instead of using snapshot tests

TODO in other pr(s):
- Add a `withContext()` function to wrapperUtil to test components that use context but don't use form
- Add mocks/factories for context and form
- Decide how we want to use Context moving forward
  - strict separation between connected and non-connected components?


spec/javascript/components/supplemental_application/sections/RentalAssistance.test.js
```diff
@@ -25,119 +31,216 @@ const rentalAssistance = {
...
+    expect(wrapper.find(RentalAssistanceTable)).toHaveLength(1)
+    expect(wrapper.find(RentalAssistanceForm)).toHaveLength(1)
   })
+})
 
-  test('should show the Other Assistance Name in the table when Type of Assistance is Other', () => {
+describe('RentalAssistanceForm', () => {
+  const getWrapper = ({
+    assistance,
+    isNew = false,
```
james-wills-sf, line 73: So doing this for default parameters does appear to work. I assume it would work the same in react component props

spec/javascript/components/supplemental_application/sections/RentalAssistance.test.js
```diff
@@ -25,119 +31,216 @@ const rentalAssistance = {
...
+    test('matches snapshot', () => {
+      expect(wrapper).toMatchSnapshot()
+    })
+
+    test('should render the correct modal ID', () => {
+      expect(wrapper.find('#rental-assistance-edit-form-0')).toHaveLength(1)
+    })
+
+    test('should render the delete button', () => {
+      expect(findByNameAndProps(wrapper, 'Button', { text: 'Delete' })).toHaveLength(1)
```
james-wills-sf, line 209: I think findByNameAndProps makes this super readable, and now we don't have to put any selector ids in the code just for testing.

akegan: I had to add the id selector for puppeteer e2e tests, not sure if this would work for puppeteer selecting but maybe it would

akegan:
From our convo earlier: please also add some documentation to the readme about how to use these test utiils and our unit test best practices.

spec/javascript/components/supplemental_application/sections/RentalAssistance.test.js
james-wills-sf: @akegan I don't think it would unfortunately. I think we'll probably need those ids for puppeteer tests for now

spec/javascript/testUtils/wrapperUtil.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
+import { shallow, mount } from 'enzyme'
+import Context from '~/components/supplemental_application/context'
+import { Form } from 'react-final-form'
+
+const formNode = (application, formToChildrenFunc) => (
+  <Form
+    onSubmit={() => null}
```
akegan, line 8: should we make this a mock? or be able to pass a mock? I suppose we can add when we find that it's needed

spec/javascript/testUtils/wrapperUtil.js
```diff
@@ -0,0 +1,78 @@
...
+      </form>
+    )}
+  />
+)
+
+const diveThroughFormWrappers = (wrapper) =>
+  wrapper
+    .find('form')
+    .children()
+    .dive()
```
akegan, line 22: this is really great to be able to "dive" through a couple of layers to be able to keep using shallow render 🎉

spec/javascript/testUtils/wrapperUtil.js
```diff
@@ -0,0 +1,78 @@
+import React from 'react'
+import { shallow, mount } from 'enzyme'
+import Context from '~/components/supplemental_application/context'
+import { Form } from 'react-final-form'
+
+const formNode = (application, formToChildrenFunc) => (
```
akegan, line 6: formToChildrenFunc isn't super intuitive for me. What about something like componentWithForm?

spec/javascript/testUtils/wrapperUtil.js
```diff
@@ -0,0 +1,78 @@
...
+const mountWithForm = (application, formToChildrenFunc) =>
+  mount(formNode(application, formToChildrenFunc))
+
+/**
+ * Return a component wrapper for any component that uses react-final-form
+ *
+ * @param application the application object to seed the form with
+ * @param formToChildrenFunc a function that takes a form object and returns a node.
+ * @param shouldMount true if component should be rendered with enzyme mount.
+ *  Only use shouldMount=true if you know you absolutely need mount functionality.
```
akegan: could be helpful to clarify here or in the readme docs what we expect could be a good reason for using mount. I'm still not clear on it myself.

akegan (approved):
Great to see the size of the snapshots go down by so much, and glad we've got the easy wrapper for adding form and context now. Thanks for adding these utils.

Couple of questions but should be good to merge once you add some documentation around using the new test utils 😄

spec/javascript/testUtils/wrapperUtil.js
james-wills-sf: I think theres a lot that can and should be added but I figure we can add to these utils as we need new features.

spec/javascript/testUtils/wrapperUtil.js
james-wills-sf: @akegan how about renderWithForm? That way it kind of matches with react-final-form's `render` prop

README.md
```diff
@@ -108,6 +108,84 @@ To run an individual test:
...
+  const wrapper = withForm(application, (form) => <ChildThatNeedsForm form={form} />)
+
+  expect(wrapper.find(ComponentA)).toHaveLength(1)
+})
+```
+
+##### Example: shallow rendering a component that uses form and context
+Say you want to test a component that looks like this:
+```
+const ComponentThatUsesFormAndContext = ({ form }) => {
```
akegan: Could be helpful to illustrate using the context here too by adding form to the props + then using something from the store similar to how you have the example of using something from the form
```suggestion
const ComponentThatUsesFormAndContext = ({ form, store }) => {
```

spec/javascript/testUtils/wrapperUtil.js
akegan: oh I see. I had missed before that we have to pass in (form) => <Component form={form} /> instead of just the component. Maybe formToChildrenFunc does actually make the most sense then.

akegan: would it be possible to just pass the component and then in the formNode function pass in the form to the component?

spec/javascript/testUtils/wrapperUtil.js
james-wills-sf: It is!

It can get confusing if used too much though. I think putting it in named functions like this or adding comments that explain what you're diving into is important. At my last job people joked that if your test didn't work just keep adding dive() until it does. That's bad lol

akegan:
documentation looks great!

spec/javascript/testUtils/wrapperUtil.js
james-wills-sf: @akegan it would (I think, never added a prop to a component after it's been created) but i'd rather we didn't do that.

It would totally work for what we're doing here, but if a component ever has a prop with a different name, or a component just uses a single field of the form this wont work.

Another thing I'm thinking we'll need eventually is to add support for the other final-form callback params (I think `values` is one?), and it would be easiest to just add it to the render func instead of adding them all as props to the child

spec/javascript/testUtils/wrapperUtil.js
james-wills-sf: I added a short explanation here and a long explanation in the readme, want to take a look and see if it makes sense to you?

I'm heading offline for the night btw

akegan: yes I think it's good!

spec/javascript/testUtils/wrapperUtil.js
akegan: ok those seem like good reasons for keeping it as is. 👍

PR #402, opened 2020-09-10
Fix parking space bug [DAH-147]

james-wills-sf:
[DAH-147]

- 2020-09-10 at 3 59 01 PM

There was a bug in the parking space select where the value would always show up as "No" on reload, even if the backend value was "Yes". This was because we were using the presence of a related dollar value field to set this field on initial load, even though we should have been using the parking toggle itself.

I also removed the useState code since this doesn't actually need to be part of component state, we can just calculate the value in render.

[DAH-147]: https://sfgovdt.jira.com/browse/DAH-147

### DIFF
--- a/app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
+++ b/app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
@@ -10,8 +10,8 @@ const No = 'No'
 
 const ParkingInformationInputs = ({ form: { change, getFieldState }, values: { lease }, disabled = false }) => {
   const onChangeHasParkingSpace = ({ target: { value } }) => {
-    if (value === No) {
-      change(monthlyRentFieldName, '')
+    if (value !== Yes) {
+      change(monthlyRentFieldName, null)
     }
   }
 
@@ -25,12 +25,11 @@ const ParkingInformationInputs = ({ form: { change, getFieldState }, values: { l
         <FormGrid.Item>
           <SelectField
             label='BMR Parking Space Assigned?'
-            selectValue={hasParkingSpace ? Yes : No}
             onChange={onChangeHasParkingSpace}
             fieldName='lease.bmr_parking_space_assigned'
             options={[Yes, No]}
             disabled={disabled}
-            noPlaceholder />
+          />
         </FormGrid.Item>
         <FormGrid.Item>
           <CurrencyField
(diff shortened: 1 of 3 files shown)


app/javascript/components/supplemental_application/sections/ParkingInformationInputs.js
```diff
@@ -9,23 +9,24 @@ const Yes = 'Yes'
...
+
+  const hasParkingSpace = lease?.bmr_parking_space_assigned === Yes
+
   return (
     <>
       <FormGrid.Row>
         <FormGrid.Item>
           <SelectField
             label='BMR Parking Space Assigned?'
-            selectValue={parkingSpaceAssigned}
```
akegan, line 27: I think we probably don't need a selectValue at all actually. None of our other select fields have values set this way. They get it from react-final-form. 

If we remove it here then I think we can remove it from SelectField entirely? If we have a case where the value doesn't match what's displayed exactly we can set that when setting the available options.

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
```diff
@@ -1434,7 +1434,7 @@ Array [
                         onBlur={[Function]}
                         onChange={[Function]}
                         onFocus={[Function]}
-                        value="Yes"
```
akegan, line 1437: Curious why this changed.

akegan (changes requested):
On this application, parking space assigned is empty on the lease, but "No" on the application s
a0o0P00000IvWgcQAF, which led me to investigate if we were saving "No" correctly on applications if leasing agents selected "No"

Looks like with both on Full and on this PR we're not saving a "No" as "No" on salesforce, unless you start with "Yes", hit save, and then switch to "No" 🤔. Makes me wonder if we should add a "Select One" option for null?

There's another issue where If you switch to "No" the monthly parking rent is showing up as $0 instead of being empty:
image
That's also an issue on full too 🤦‍♀️

spec/javascript/components/supplemental_application/__snapshots__/SupplementalApplicationPage.test.js.snap
james-wills-sf: The initial value used to be based on the monthly rent, so if you started with:
monthly_rent = $100
bmr_parking_space_assigned = "No" (or null)

The old version would initialize to "Yes" and the new version inits to "No"

In this case the monthly rent is set to 790.49

james-wills-sf:
@akegan I switched this to have a "Select one..." option for null and made it so that if you switch to "No" the monthly rent is saved as null, not $0.00.

james-wills-sf:
@akegan e2e tests should be fixed now

akegan (approved):
Did QA and it looks good! Don't forget to add Jira ticket number to PR title/commit on merge

PR #403, opened 2020-09-12
Connect new lease api [DAH-93]

james-wills-sf:
[DAH-93]

This change:
- hooks up the save button in the lease section to save just the lease and assistances
- Changes the behavior of the top-level save button to only save the lease and assistances when the user is in edit mode
- changes the behavior of the lease, assistances and application actions to only trigger an api call if something has changed.
- Updates the tests to no longer put expects in .then(). Errors were going unnoticed because if you throw in a then it still marks the test as passing for some reason!
 

### DIFF
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -111,12 +111,12 @@ const updateUnsavedRentalAssistances = async (application, prevApplication) => {
   )
 }
 
-const createOrUpdateLease = async (lease, prevLease, applicantId, applicationId) => {
+const createOrUpdateLease = async (lease, prevLease, primaryApplicantContactId, applicationId) => {
   const isLeaseChanged = isChanged(prevLease, lease)
 
   const saveLeasePromise = isLeaseAlreadyCreated(lease)
-    ? () => apiService.updateLease(lease, applicantId, applicationId)
-    : () => apiService.createLease(lease, applicantId, applicationId)
+    ? () => apiService.updateLease(lease, primaryApplicantContactId, applicationId)
+    : () => apiService.createLease(lease, primaryApplicantContactId, applicationId)
 
   return performOrDefault(
     isLeaseChanged && !isEmpty(lease),
--- a/app/javascript/components/supplemental_application/actions.js
+++ b/app/javascript/components/supplemental_application/actions.js
@@ -19,6 +19,11 @@ const transformApplicationResponses = (lease, application, rentalAssistances) =>
   rental_assistances: rentalAssistances || []
 })
 
+const defaultLeaseResponse = (application) => ({
+  lease: application.lease,
+  rentalAssistances: application.rental_assistances
+})
+
 /**
  * Update any fields on the application that have been changed from
  * prevApplication. This action triggers multiple separate requests:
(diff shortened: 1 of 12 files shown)


app/javascript/components/supplemental_application/actions.js
```diff
@@ -76,26 +104,25 @@ const updateUnsavedRentalAssistances = async (application, prevApplication) => {
...
+  return performOrDefault(
+    !isEmpty(promises),
+    () => Promise.all(promises).then(() => apiService.getRentalAssistances(application.id)),
+    rentalAssistances
+  )
 }
 
-export const createLease = async (lease, primaryApplicantContact, applicationId) =>
-  apiService.createLease(lease, primaryApplicantContact, applicationId)
+const createOrUpdateLease = async (lease, prevLease, applicantId, applicationId) => {
```
akegan: maybe primaryApplicantContactId here? Not sure if the Id changed or just the label, but there's also a household member for the applicant that has an ID so this might be a bit ambiguous

akegan (changes requested):
Looks like rental assistances aren't coming through on lease save. They do show up when you save at the application level so maybe that's a place to start
missing-rental-assistance

It also looks like changes to the application outside of lease are wiped out by lease save rather than just not saved
app-changes-erased

- [ ] changes the behavior of the lease, assistances and application actions to only trigger an api call if something has changed. - > Save on the application does still seem to be triggering API calls even if there's no change but that's OK


Thanks for the unit test updates, I was thinking that we needed a ticket to clean up our unit tests warnings so that we could spot these errors!

Other than that looks good, I confirmed:

- [x] hooks up the save button in the lease section to save just the lease and assistances
- [x] Changes the behavior of the top-level save button to only save the lease and assistances when the user is in edit mode - (hard to test because can't make changes not in edit mode)
- Tried creating a lease + rental assistance at the same time and they were connected correctly

james-wills-sf:
@akegan okay, those issues should be fixed now.

I also fixed a thing where the rental assistances buttons weren't switching to the loading state when the lease or application were being loading. I think all of these buttons should sync their disabled states so there aren't any race conditions.

akegan:
Just found another issue where a rental assistance edit panel didn't collapse on lease save. 

Tried to then save at the application level and got this error at the top:
image

On this application https://dahlia-lap-full-pr-403.herokuapp.com/applications/a0o0P00000GZazsQAD/supplementals.

I don't see any network requests or console logs though, and it happens even after refreshing the page. 
Edit: It looks like it was because type of assistance was missing but the sequence of things I did made it not visible somehow or maybe I just missed it. nice-to-have would be if the rental assistance panel opened if there was an error on one of the fields. That application still has the issue if you want to check it out

james-wills-sf:
@akegan I updated with a fix for the rental assistances edit issue

There are a couple error handling related things I think we should push to a separate PR:
1. open rental assistance when there's a rental assistance error
2. validate lease on save (I don't think we're doing any of this now)
3. Decide how we want to handle other network errors. Probably just catch, log, and show an alert window?

2 is a big change which is why I'd rather it be in another PR

akegan:
I think that's reasonable to push those off into a separate PR or even a separate ticket

akegan (approved):
changes look good now!

PR #404, opened 2020-09-14
Add delete functionality [DAH-93]

james-wills-sf:
Ticket: [DAH-93]

- hook up delete button functionality
- create a new lease even if it's empty but you're in edit mode

### DIFF
--- a/app/javascript/apiService.js
+++ b/app/javascript/apiService.js
@@ -131,6 +131,16 @@ export const createLease = async (leaseToCreate, primaryApplicantContact, applic
     .then(response => response.lease)
 }
 
+/**
+ * Delete a lease associated with an application.
+ * Additionally, delete any rental assistances associated with that lease.
+ *
+ * @param {Number} applicationId the application ID the lease is attached to
+ * @param {Number} leaseId the lease ID to delete
+ */
+export const deleteLease = async (applicationId, leaseId) =>
+  request.destroy(`/applications/${applicationId}/leases/${leaseId}`, null, true)
+
 export default {
   updateApplication,
   updateFlaggedApplication,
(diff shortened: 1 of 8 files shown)


akegan:
@james-wills-sf I think review instructions are going to be a bit ridiculous on this ticket, but can you make sure that you add that we need to test all this behavior logged in as a leasing agent too? Especially lease delete because msol had to add the permission

akegan:
somehow got into a state where if I click edit lease on https://dahlia-lap-full-pr-404.herokuapp.com/applications/a0o0P00000IvWgcQAF/supplementals then The new rental assistance panel is opening up. 
Screen Shot 2020-09-15 at 10 12 44 AM

I'll leave this app in that state so you can check it out

Another note with this: If you have the new rental assistance panel and an edit panel open and you click cancel on the edit panel it closes the new panel too 🤔

james-wills-sf:
@akegan Okay I made some non-trivial changes to how the rental assistance code works:

- View state that only pertains to the RentalAssistance component now lives in that component.
  - This means the SupplementalApplicationPage component no longer sets showNewRentalAssistanceButton, etc
- Only the top-level RentalAssistance component is connected to the store, other components get everything through props
  - This makes reasoning about rental assistance state logic much simpler, it's all in the cluster of handler functions in the RentalAssistance render
- I deleted the Panel component
  - once I removed the store connection this component wasn't doing much
- I made sure we're clearing out rental assistances in the react state on delete
  - this was a bug that wasn't handled before.
- I made sure "cancel" deleted the modified rental assistances
  - this was a bug that wasn't handled before

Another possible behavior change: we're not restricting which panels you can have open at once. If you open an edit panel and then the "Add new assistance" panel, that's fine we keep both open.

## Screencast: panel open/close behavior
opening and closing assistances

## Screencast: cancel closes all panels and keeps them closed
cancel closes panels

akegan (approved):
Not sure I did all the same QA steps again but everything seems to be working as expected and the bug I found before is not showing up again so should be good

PR #405, opened 2020-09-15
Validate lease and rental assistances on save [DAH-93]

james-wills-sf:
This adds lease and rental assistance validation on lease save.

I think opening the rental assistance panel on error is going to be difficult and probably not worth the work, but we can open a ticket to add it later. Especially now that rental assistances are properly validated it shouldn't be possible to save an assistance without a type

save lease validation

### DIFF
--- a/app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
+++ b/app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
@@ -70,7 +70,7 @@ const Income = ({ listingAmiCharts, visited, form }) => (
   </ContentSection>
 )
 
-const LeaseSection = ({ form, submitting, values, onCreateLeaseClick, showLeaseSection }) => (
+const LeaseSection = ({ form, values, onCreateLeaseClick, showLeaseSection }) => (
   <ContentSection
     title='Lease'
     description={!showLeaseSection && 'Complete this section when a unit is chosen and the lease is signed. If the household receives recurring rental assistance, remember to subtract this from the unit’s rent when calculating Tenant Contribution.'}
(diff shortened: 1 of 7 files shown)


akegan (changes requested):
Found an issue:

Steps to reproduce:
1. Create a rental assistance without type of assistance
1. Try to submit lease, see error above
1. Fix error by adding a type of assistance
1. Try saving lease again

It doesn't save but doesn't show an error. Saving the rental assistance individually works, but the lease still thinks there's an error I think.

I'm also wondering if we can add an alert bar or focus on the errored field or something. It is likely out of the scope of this pr, but it seems doable [with this library](https://github.com/final-form/final-form-focus) and right now it's a bit confusing because for me the errored field wasn't on the screen when I hit save so it wasn't clear at all why the button wasn't working.

james-wills-sf:
@akegan I'm actually not able to reproduce with this application: /applications/a0o0P00000JQawYQAT/supplementals

is it possible the application you're on has a different rental assistance with an invalid type?

Here's a gif in case I'm reproing it wrong:
save working properly 2

akegan:
@james-wills-sf I really thought that I tried also starting a new rental assistance and having the type set on the same application and saved by lease save and that was fine. I'll look again.

Edit I'm getting the same thing on https://dahlia-lap-full-pr-405.herokuapp.com/applications/a0o0P00000GZazsQAD/supplementals

edit2:

Looks like I am getting a console error: 
image
Maybe it has to do with being the only rental assistance?

edit 3: Still get it with multiple rental assistances

james-wills-sf:
@akegan okay I fixed that specific error and fixed a small bug where clicking the top level page "save" button caused the edit forms to collapse when they shouldn't

Can you see if you're able to repro now?

akegan (approved):
looks good now! I had noticed that the rental assistance was collapsing kind of fast when I hit save so the 2nd fix with removing the submitting is great! All coming together 😄

PR #406, opened 2020-09-16
Add jsconfig.json for vscode path completion

james-wills-sf:
Vscode has trouble navigating to definitions when we use paths from an aliased home directory, like 
```
import { doesApplicationHaveLease } from '~/utils/leaseUtils'
```

With this new jsconfig.json file, vscode can resolve these references so you can easily jump to function or component definitions.


akegan (approved):
App still looks fine so this seems OK to me!

PR #408, opened 2020-09-17
Make styled dropdowns accessible DAH-124

akegan:
DAH-124

This PR:
- Switches our styled select component to use react-select
- Makes it so that you can use arrows to navigate our styled selects (currently just status and substatus dropdowns
- Updates the status modal wrapper to be a bit more inline with the rest of our form improvements

Still TODO:
- [x] Fix hover/focus/selected colors
- [x] Add tests
- [ ] Add example to the pattern lib


app/javascript/components/molecules/StatusDropdown.js
```diff
@@ -1,37 +1,85 @@
...
+    'text-align-left',
+    { 'expand': expand },
+    { 'tiny': size === 'tiny' },
+    { 'small': size === 'small' }
+  ]
+  const renderStatusToggle = ({ children, getValue, ...props }) => {
+    const val = getValue()[0]
+
+    return (
+      <button
```
akegan, line 40: if all of our styled dropdowns are going to be buttons then it would make sense to make more generalized. The differences between substatus and status dropdowns are just a couple of classes

app/javascript/components/atoms/colors.js
```diff
@@ -0,0 +1,35 @@
+export const COLORS = {
```
akegan, line 4: here's the colors constants I pulled in. figured we could generalize to styles if we need to at some point, but for now colors seem like enough

app/javascript/components/organisms/StatusModalWrapper.js
```diff
@@ -50,59 +53,72 @@ const StatusModalWrapper = ({
...
-          </Field>
+          <FormGrid.Row paddingBottom>
+            <FormGrid.Item fullWidth>
+              <Field
+                name='subStatus'
+                component={({ input: { onChange }, meta }) => {
+                  const hasError = !values.subStatus && meta.touched && meta.error
+                  return (
+                    <div className={classNames('form-group', hasError && 'error')}>
+                      <Label label='Status Detail' blockNote='(required)' />
```
akegan, line 91: could be nice to pull these into Field.js but I think they are maybe just slightly too complicated and specific for now.

akegan (commented):
Looks like just 2 items where I'm thinking I could refactor more.

app/javascript/components/molecules/SubstatusDropdown.js
```diff
@@ -0,0 +1,92 @@
...
+  const buttonClasses = [
+    'button',
+    'dropdown-button',
+    'substatus',
+    'has-icon--right',
+    'text-align-left',
+    { 'expand': expand },
+    { 'error': hasError }
+  ]
+  const renderStatusToggle = ({ children, getValue, ...props }) => {
```
james-wills-sf, line 42: what are the other props that can be passed in here? Is it just a couple that can be named instead?

spec/javascript/components/organisms/StatusModalWrapper.test.js
```diff
@@ -0,0 +1,127 @@
+import React from 'react'
+import { mount } from 'enzyme'
+
+import StatusModalWrapper from '~/components/organisms/StatusModalWrapper'
+import FormModal from '~/components/organisms/FormModal'
+import { LEASE_UP_SUBSTATUS_OPTIONS } from '~/utils/statusUtils'
+
+const statusUtils = require('~/utils/statusUtils')
```
james-wills-sf, line 14: I assume we need to import with `require` because of jest? Can you make a comment explaining that?

spec/javascript/components/organisms/StatusModalWrapper.test.js
```diff
@@ -0,0 +1,127 @@
...
+      expect(ON_SUBMIT).not.toHaveBeenCalled()
+    })
+
+    test('should stop showing error state and allow submission when required fields are filled out', () => {
+      wrapper = getWrapper({ status: 'Appealed', isOpen: true })
+      wrapper.update()
+      wrapper.find('button.primary').simulate('submit')
+
+      // Verify there are errors initially
+      expect(wrapper.find('SubstatusDropdown').props().hasError).toBe(true)
```
james-wills-sf: I think when it's a component that's already exported it's better to import and pass in the component itself instead of using the string name

```
import SubstatusDropdown from '~/components/molecules/SubstatusDropdown'
...
expect(wrapper.find(SubstatusDropdown).props().hasError).toBe(true) 
```

This comment also goes for `TextAreaField`

spec/javascript/e2e/LeaseUpPage.e2e.js
```diff
@@ -3,8 +3,9 @@ import { LEASE_UP_LISTING_ID, DEFAULT_E2E_TIME_OUT } from '../support/puppeteer/
 import supplementalApplicationSteps from '../support/puppeteer/steps/supplementalApplicationSteps'
 import SetupBrowserAndPage from '../utils/SetupBrowserAndPage'
 
+const firstRowDropdown = '.rt-tr-group:first-child .rt-td .dropdown .dropdown-button'
 describe('LeaseUpPage', () => {
-  test('should change "Lease Up Status" for specific application preference using dropdown in row', async () => {
+  test.only('should change "Lease Up Status" for specific application preference using dropdown in row', async () => {
```
james-wills-sf: be sure to remove the .only here!

james-wills-sf (approved):
@akegan QA'd and things look good!

One additional question I have is why is toolkit.scss modified?

app/javascript/components/molecules/SubstatusDropdown.js
akegan: I'm not sure, they're passed by Select and they don't necessarily enumerate them in the docs. I think there might be a lot though.

spec/javascript/components/molecules/StatusDropdown.test.js
```diff
@@ -0,0 +1,130 @@
...
+      const expectedButtonClasses = [
+        'button',
+        'dropdown-button',
+        'has-icon--right',
+        'text-align-left',
+        'is-appealed'
+      ]
+      const toggleButton = wrapper.find('Select').find('.status-dropdown__control').find('button')
+      expect(toggleButton.text()).toEqual('Appealed')
+      expect(toggleButton.hasClass(expectedButtonClasses.join(' '))).toBe(true)
```
james-wills-sf: I think this might be tied a little closely to the implementation, since I think this will check ordering too.

How about we just check that the is-appealed class is added here, since that seems most relevant to the test

spec/javascript/components/molecules/StatusDropdown.test.js
```diff
@@ -0,0 +1,130 @@
...
+    })
+
+    test('when the dropdown is disabled', () => {
+      wrapper = getWrapper({ disabled: true })
+      const toggleButton = wrapper.find('Select').find('.status-dropdown__control').find('button')
+      expect(toggleButton.prop('disabled')).toEqual(true)
+    })
+
+    test('when a placeholder is provided', () => {
```
james-wills-sf: I think semantically you're naming some of these tests more like how a describe block would be named.

I think `test` blocks should be more like "should do X" or "does X when Y"

So this could be test('renders a placeholder when provided')

spec/javascript/components/molecules/StatusDropdown.test.js
```diff
@@ -0,0 +1,130 @@
...
+    const testLabel = 'testLabel'
+    const sampleOption = {
+      value: testValue,
+      label: testLabel,
+      statusClassName: testStatusClassName
+    }
+
+    test('with a styled value', () => {
+      wrapper = getStatusOptionWrapper(sampleOption, 'notSelected')
+      const expectedLiClasses = `dropdown-menu_item ${testStatusClassName}`
```
james-wills-sf: Another case where I'm concerned that we're checking ordering where we don't really care what order the classes are in. Could this be two separate expects, one for each class we want to check?

james-wills-sf (commented):
@akegan thanks for writing all those tests, lgtm but I'll re-review after the merge conflicts are resolved

spec/javascript/components/molecules/StatusDropdown.test.js
akegan: updated in StatusDropdown and SubstatusDropdown

akegan:
@james-wills-sf OK merge conflicts resolved and comments addressed! Just needs a last quick review + QA then we should hopefully be able to merge

spec/javascript/components/molecules/Dropdown.test.js
```diff
@@ -3,6 +3,9 @@ import { shallow } from 'enzyme'
 import Dropdown from '~/components/molecules/Dropdown'
 import { components } from 'react-select'
 
+// Use a string selector in these tests to avoid confusion with the StateManager wrapper that exists in react-select
```
james-wills-sf, line 6: Thanks for adding a comment here!

james-wills-sf (approved):

PR #409, opened 2020-09-17
Remove extra GET request on delete confirm

james-wills-sf:
Turns out `<a href="#" />` triggers a GET request on click.

This only used a link when a button destination is provided, otherwise it renders as a regular button.

- 2020-09-17 at 4 18 38 PM


james-wills-sf:
@akegan could you take a look at this PR? I think this fix should be merged before we cut the release branch

akegan (approved):
looks good. Still looks and works as expected

PR #411, opened 2020-09-18
Chores/fix silent test errors [DAH-154]

james-wills-sf:
Ticket: [DAH-154]

## fail tests on proptype errors
- Add jest-prop-type-error package
- fix tests that fail as a result

This caught a few issues with our tests, like 

## Fix tests that fail silently due to async issues
- Add eslint jest rules to fail when a test isn't awaiting or returning a promise

## how to run linting
Same as before, `yarn lint` and `yarn lint:fix`, precommit hook works as well.

# Changes you'll need to make in your local config
run `yarn install`
If you're using vscode:
- disable prettier-standard and standardJS extensions
- download eslint extension
  - in your vscode settings, set:
```
    "editor.codeActionsOnSave": {
        // For ESLint
        "source.fixAll.eslint": true,
    },
    "eslint.format.enable": true,
    "eslint.lintTask.enable": true,
    "eslint.packageManager": "yarn"
```

[DAH-154]: https://sfgovdt.jira.com/browse/DAH-154


spec/javascript/components/applications/__snapshots__/ApplicationsPage.test.js.snap
```diff
@@ -507,12 +466,13 @@ Array [
                   }
                 }
               >
-                <span>
-                   
-                </span>
+                <div>
+                  08/31/2016
```
akegan, line 470: nice this looks like it fixed a bug in a test?

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable jest/no-conditional-expect */
```
akegan, line 1: how does this differ from expect-expect?

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
```diff
@@ -149,23 +150,23 @@ describe('LeaseUpApplicationsPage', () => {
...
+        expect(emptyStatus.toLowerCase()).toContain('select one...')
 
         wrapper.find(subStatusSelector).find('button').simulate('click')
         await tick()
         wrapper.find(`${subStatusSelector} .dropdown-menu li a`).first().simulate('click')
         await tick()
 
         const updatedStatus = wrapper.find(`${subStatusSelector} button`).html()
-        expect(updatedStatus.toLowerCase().includes('select one...')).toBe(false)
+        expect(updatedStatus.toLowerCase()).not.toContain('select one...')
```
akegan, line 169: these are going to have to be re-done on my PR since I moved the tests to another file. I can make the changes now

akegan (approved):
I can't comment on it, but I'm wondering why toolkit.scss has a diff. It looks like it's causing the inline modal for lease to disappear:
image


Other than that, it looks good!

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
akegan: actually I don't have these checks in there anymore so it should be fine

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
james-wills-sf: @akegan this rule doesn't let you do:
```
if (someCondition) {
  expect(athing).toEqual(anotherThing)
}
```

expect-expect doesn't let you do:

```
expect(athing)
```

(notice that there's no .toEqual(), toHaveLegth, etc. assertion on that so it's not actually testing anything

james-wills-sf: @akegan Oh  nevermind that's valid-expect.

expect-expect makes sure you have an expect somewhere in your test. I turned it off because some of our tests don't have expects since we assume if you make it to the end without failing the test should pass.

akegan: I actually meant to ask about the difference with valid-expect so you somehow managed to answer my question with your first response!

PR #413, opened 2020-09-22
Add household size and accessibility info to supp app [DAH-142]

software-project:
DAH-142


app/javascript/components/molecules/FormGrid.js
```diff
@@ -30,8 +30,12 @@ FormGrid.Row.defaultProps = {
 
 // Wrapper for each form field.
 // NOTE: Grid is hardcoded. This should not be generic element
-FormGrid.Item = ({ children }) => (
-  <div className='form-grid_item small-6 medium-6 large-6 column'>
+FormGrid.Item = ({ children, small = false }) => (
+  <div className={classNames(
+    'form-grid_item column',
+    { 'small-6 medium-6 large-6': !small },
```
james-wills-sf: nit: if small/med/large values are equal only include the smallest class (in this case small-6)

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -102,9 +102,13 @@ const Lease = ({ form, values, store }) => {
   const disabled = !isEditingMode
 
   const availableUnitsOptions = formUtils.toOptions(
-    map(availableUnits, pluck('id', 'unit_number'))
+    map(availableUnits, pluck('id', 'unit_number', 'priority_type'))
   )
 
+  const accessibilityUnits = availableUnits && availableUnits.filter((unit) => {
+    return unit.priority_type && unit.priority_type.match(/Mobility|Hearing|Vision/)
```
james-wills-sf: nit: can be `return unit.priority_type?.match(/Mobility|Hearing|Vision/)`

james-wills-sf (approved):
I think I already submitted this review so I apologize for the double comments if so, but this LGTM and it looks good onm QA

only thing is the titles are font weight 700 in the review app and the designs have them at 600. BUt it seems to be what we default to for strong so that's something we can revisit later.

PR #414, opened 2020-09-23
Chores/new linting rules

james-wills-sf:
Close out some TODOs and add recommended linting rules

All of these rules are auto-fixable, so running yarn lint:fix should cover them all.

## How to review this PR
I would go commit by commit, they're all independent and each adds a new rule


.eslintrc.js
```diff
@@ -31,6 +31,10 @@ module.exports = {
     'react'
   ],
   rules: {
+    /*
+     * TODO: Add react-hooks library for the react-hooks/exhaustive-deps lint
```
akegan: should we make a ticket for this? Is it a significant amount of work?

akegan (approved):
All look like great changes! Appreciated looking at the commits one by one

james-wills-sf:
@akegan it was quick enough that I went ahead and added the exhaustive deps rule if you want to take another look.

I actually added react-hooks/recommended which adds two rules:

### 1. react-hooks/exhaustive-deps
This makes sure that if you provide an array of dependencies, that it actually includes all of the dependencies.

Note that this will trigger an error if you use an empty array of dependencies, which is what we do when we want the effect to run only once ever for a component. To get around this I made the custom hook useEffectOnMount(), which I think is more readable anyways.

If this gets too annoying we can turn the rule off, but I think we should be using useEffect sparingly anyways.

### 2. react-hooks/rules-of-hooks
See [rules of hooks](https://reactjs.org/docs/hooks-rules.html)
This makes sure you're never calling a hook conditionally, like:
```
[myVar1, setMyVar1] = useState(12)
if (someCondition) {
  [myVar2, setMyVar2] = useState(false)
}
[myVar3, setMyVar3] = useState({ anobject: 12 })
```

The reason this is bad is that react uses the order of state hooks to assign the variables, so if there are conditionals and the order changes your myVar1, myVar2, and myVar3 variables will actually switch values silently.

akegan (approved):
looks good still, thanks for adding those changes. Not sure why the deploy failed. Said that rails app didn't compile. Let's make sure we can get a successful deploy before merge. 

I just kicked off another deploy to see if it'll work

james-wills-sf:
@akegan I think it was that I was doing `import useEffectOnMount` directly instead of like `import { useEffectOnMount }`

PR #415, opened 2020-09-24
Chores/prettier linting

james-wills-sf:
Here's the new formatting change with prettier.

To set it up in vscode:
- run yarn install
- install the `prettier` extension in vscode
- install the `eslint` extension in vscode
- you should be good to go, some other settings that I have that you may or may not need:
```
    "editor.codeActionsOnSave": {
        // For ESLint
        "source.fixAll.eslint": true,
    },
    "eslint.format.enable": true,
    "eslint.lintTask.enable": true,
```


james-wills-sf:
A note on the updated snapshots: They don't actually correspond to a new line when the page renders:

- 2020-09-24 at 2 52 34 PM

akegan (approved):
looks good. can we add a section to the readme about linting with the recommended extensions and settings?

PR #416, opened 2020-09-26
Remove bootstrapped data for listings and applications pages [DAH-42]

james-wills-sf:
DAH-42

Switch the listings and leasupapplications pages to load via javascript instead of using data bootstrapped from rails.

This is a first step for us to be able to route using react-router instead of rails.

This PR causes the behavior to be a little weird on the applications page, since we don't have the listing title on first load. I think this behavior can be polished with future PRs for this ticket.

TODO:
- [x] update rspec tests. I was having issues getting VCRs to work so I'll revisit on monday
- [x] Add tests to the listings page javascript file


app/javascript/routes/LeaseUpApp.js
```diff
@@ -0,0 +1,35 @@
...
+    <div>
+      <Switch>
+        <Route path='/lease-ups/listings/:listing_id'>
+          <LeaseUpApplicationsPage />
+        </Route>
+        <Route path='/lease-ups/listings'>
+          <LeaseUpListingsPage />
+        </Route>
+        <Route path='/lease-ups/applications/:application_id/supplemental'>
+          <LeaseUpListingsPage />
```
akegan: Should this be the SupplementalApplicationPage?

also it looks like links from the applications table are still going to: https://dahlia-lap-full-pr-416.herokuapp.com/applications/a0o0P00000JlYj8QAF/supplementals, with no /lease-ups. Is that coming in a separate PR?

Actually looks like these routes don't work yet. Should I just totally ignore this part for now?

akegan (commented):
Do you have some review instructions for what I should QA for this PR? Not totally clear on what it does/doesn't cover.

james-wills-sf:
@akegan Ah yeah let me write up review instructions, I'll let you know when they're done.

This PR is pretty small in scope, just moving the listings and applications pages to load without bootstrapped data

james-wills-sf:
@akegan review instructions for this PR:

1. open the listings page
   - verify it loads properly
   - verify pagination works (not sure if you can do this with full listings since there aren't that many)
   - verify clicking any listing works
2. Applications page
   - Verify page loads when going directly to the URL and when clicking from the listing page
   - Verify listing title loads properly (this now loads after initial page load)
   - Verify switching pages works
     - Verify that the listing title remains even while the next page is loading
   - Verify filtering works
   - verify updating status works
   - verify clicking into supp app page works
3. supplemental/short form changes
   - Verify nothing changes with the behavior of these pages.

akegan:
looks like something with eagerpagination kinda broke. Now have to wait a couple beats to load between every page instead of only every 5 pages. 

also this may have still been a bug from before, but when we filter it doesn't reset the page # to 1 
image

james-wills-sf:
@akegan okay eager pagination should be working again.

About the page numbers not resetting on filter, I'm pretty sure it was a problem before this PR and that it's going to be a pain to fix. We'll have to refactor how the table is used, see https://stackoverflow.com/questions/48630384/use-buttons-to-trigger-filter-function-on-react-table-in-react so I think it's out of scope for this PR

james-wills-sf:
@akegan all tests are passing, do you mind re-reviewing and approving if it looks good to you?

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
```diff
@@ -203,8 +203,8 @@ describe('LeaseUpApplicationsPage (shallow)', () => {
           expect(mockFetchLeaseUpApplications.mock.calls[1][0]).toEqual(mockListing.id)
 
           // It's confusing but this page param is actually the server page, so even though
-          // the client, "eager" page is 1 the server page is still 0.
-          expect(mockFetchLeaseUpApplications.mock.calls[1][1]).toEqual(0)
+          // the client, "eager" page is 6 the server page is only 1.
```
akegan, line 206: I thought page 6 should be page 2 on the server? I thought we switched to  new server page every 100 records/5 pages

akegan (commented):
One question about deleting unused stuff, otherwise looks good and eager pagination seems to be working again

akegan (approved):
whoops I meant for that last comment to be an approval

spec/javascript/components/lease_ups/LeaseUpApplicationsPage.test.js
james-wills-sf: @akegan yeah the page is zero indexed so this is page 2

PR #418, opened 2020-09-30
Update shortform page to load via javascript [DAH-42]

james-wills-sf:
- Move rails find_application code from view controller to shortform api
- remove unnecessary `fields` param from bootstrapped response
- remove application and file_base_url from bootstrapped response
- load application and file_base_url via javascript

TODO (in future routing PR)
- remove the three bootstrapped fields that are left, `application_id`, `is_lease_up` and `show_add_btn`. All of these can be  grabbed via the url in javascript with react-router.


app/controllers/api/v1/short_form_controller.rb
```diff
@@ -27,6 +35,85 @@ def submit
 
   private
 
+  def find_application(id)
```
james-wills-sf, line 38: All of this added code is cut directly from application_controller.rb

app/controllers/applications/supplementals_controller.rb
```diff
@@ -6,6 +6,7 @@ class SupplementalsController < ApplicationController
     before_action :authenticate_user!
 
     def index
+      @application_id = params[:application_id]
```
james-wills-sf: Oops, I need to remove this!

app/controllers/applications_controller.rb
```diff
@@ -31,10 +32,6 @@ def application_listing
     @listing = soql_listing_service.listing(params[:listing_id])
   end
 
-  def custom_api_application_service
```
james-wills-sf, line 34: all changes below this line were moved to API short_form_controller

app/javascript/components/lease_ups/shortFormActions.js
```diff
@@ -0,0 +1,4 @@
+import apiService from '~/apiService'
+
+export const getShortFormApplication = async (applicationId) =>
```
james-wills-sf, line 3: I think after the dust settles on the routing/state work we should think about how to split up our actions and apiService request functions in a more reasonable way

spec/fixtures/controllers/applications/lease_up_application_domain.json
```diff
@@ -1,7 +1,92 @@
 {
-  "total_monthly_rent": null,
-  "sub_status": "Written withdrawal",
-  "status": "Submitted",
+  "alternate_contact": null,
```
james-wills-sf, line 2: so some of the data changed when the vcr was updated, and the new data was all in a different order.

I sorted this all by alphabetical order so that next time the data is updated hopefully the diff won't be so bad as long as we keep it sorted

akegan: oh interesting! I thought we hadn't gotten fixtures hooked up to vcr cassettes for this repo 🤔

james-wills-sf: @akegan well the vcr cassettes get data from real responses, these fixtures are just what we're comparing them with, so when the response changed, I had to change the expected result fixtures as well

akegan: ohh I see. I thought you were saying that this automatically updated, which is something that we've wanted to setup for a while 😅

app/javascript/components/applications/ApplicationPage.js
```diff
@@ -36,8 +39,20 @@ const buildActionLinkIfNecessary = (app, showAddBtn) => {
...
+  const [fileBaseUrl, setFileBaseUrl] = useState(null)
+  const [loading, setLoading] = useState(true)
+
+  useEffectOnMount(() => {
+    getShortFormApplication(applicationId)
+      .then((response) => {
+        setApplication(response.application)
+        setFileBaseUrl(response.fileBaseUrl)
+      })
+      .catch(() => window.alert('The application you requested could not be found.'))
```
akegan: this doesn't seem to be working, see comment above

akegan (approved):
QA'd a normal application: https://dahlia-lap-full-pr-418.herokuapp.com/applications/a0o1F000002NghqQAC and confirmed that attachments and hh members and everything show up as expected

Snapshots that are present look good too.

In the case where a snapshot isn't present, it just shows up as blank instead of showing an error, while still throwing a 503 error. Not sure if that's in the scope of this PR to fix, but could be in the scope of this ticket?
image

james-wills-sf:
@akegan thank you for catching this! I had forgotten to pass throwOnError = true to the request function. Updated this call and all of the other calls I've added in recent PRs. The behavior is fixed now.

PR #420, opened 2020-09-30
component unit test for AlertBox

aekong:


akegan:
@aekong e2e tests are failing because of a different issue, so no need to worry! I can re-run once we fix the other issue.

aekong:
> @aekong e2e tests are failing because of a different issue, so no need to worry! I can re-run once we fix the other issue.

ah cool yeah saw it in the circleci pipelines.  thanks for the heads up 👍

spec/javascript/components/molecules/AlertBox.test.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import AlertBox from '~/components/molecules/AlertBox'
+
+describe('AlertBox', () => {
+  test('it renders an alert box with a message', () => {
```
akegan, line 9: one small nit is that I might separate these out into 2 separate tests, so one that is 'it renders an alert box with a message', and one that is 'it calls onClose click when the close button is clicked' or something like that

akegan (approved):
This looks great!

spec/javascript/components/molecules/AlertBox.test.js
aekong: that makes sense! i'll separate them out 👍

akegan (approved):
looks great! merge away once tests pass 🚀  thank you for these tests!!

PR #422, opened 2020-10-05
Fix supp app saving on lease save clicked [DAH-550][DAH-549]

james-wills-sf:
[DAH-550][DAH-549]

## DAH-550
Just made it so that the lease and rental assistance save buttons say "Saving..." when a request is in progress.
- 2020-10-05 at 3 22 29 PM

Note: Even if you're just saving the lease or rental assistance, all save buttons will say "Saving...". I think this is okay

### The issue
Supplemental app was saving when changes were made to the lease and lease save is clicked

### The cause
During validation, we thought the lease had errors when it really didn't. This caused us to trigger `form.submit()` to show those non-existent errors, but since there were no real errors, the form actually tried to submit!

### The fix
Change validation to not change the errors object if no errors were found.

[DAH-550]: https://sfgovdt.jira.com/browse/DAH-550
[DAH-549]: https://sfgovdt.jira.com/browse/DAH-549


akegan (approved):
looks good

PR #423, opened 2020-10-06
[rev3] Fix preference used select [DAH-551]

james-wills-sf:
[DAH-551]

## The issue
Switching "Preference used" to "None" and clicking save would cause an error because we were passing the literal string "None" to the backend.

### The cause
The way SelectField works, if you have a value of `null`, it will treat the label as the value. Instead the value needs to be set as an empty string. This is really confusing

### The fix
we should always use FormUtils.toEmptyOption to create empty options to avoid this.

I verified that setting the form value to an empty string still saves the preference as null on the backend.

[DAH-551]: https://sfgovdt.jira.com/browse/DAH-551

### DIFF
--- a/app/javascript/components/supplemental_application/sections/Lease.js
+++ b/app/javascript/components/supplemental_application/sections/Lease.js
@@ -20,6 +20,8 @@ import { EDIT_LEASE_STATE } from '../SupplementalApplicationPage'
 import { doesApplicationHaveLease } from '~/utils/leaseUtils'
 import { areLeaseAndRentalAssistancesValid } from '~/utils/form/formSectionValidations'
 
+const NONE_PREFERENCE_LABEL = 'None'
+
 const toggleNoPreferenceUsed = (form, event) => {
   // lease.preference_used need to be reset, otherwise SF validation fails
   if (isEmpty(event.target.value)) {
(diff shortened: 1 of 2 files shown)


app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -217,6 +217,7 @@ const Lease = ({ form, values, store }) => {
           <FormGrid.Item>
             <SelectField
               label='Preference Used'
+              selectValue={values.lease.preference_used || NONE_PREFERENCE_LABEL}
```
james-wills-sf, line 220: This is a little weird, the handleCancelClicked handler wasn't setting this value back to none properly

to repro:
1. start with appication with lease that has "none" preference
2. change preference to something other than none
3. Click cancel
4. Note that the preference field hasn't reverted to "None" like it should have.


The ` || NONE_PREFERENCE_LABEL` here makes it so that when preference_used is undefined it sets the value to "None", since for some reason undefined doesn't revert to the "None" option

akegan (approved):
this looks good too. Wondering if there's a way to test this since it's been a tricky interaction in the past

james-wills-sf:
@akegan I think the Lease.js file needs a little bit of an overhaul to make it easier to test and then this could definitely be tested

james-wills-sf:
@akegan I think the Lease.js file needs a little bit of an overhaul to make it easier to test and then this could definitely be tested

akegan:
OK maybe we can create a follow-up ticket?

james-wills-sf:
@akegan follow up ticket here: DAH-558

PR #427, opened 2020-10-09
Add imputed assets field to confirmed income DAH-89

akegan:
DAH-89

Review instructions in the ticket.


app/models/force/application.rb
```diff
@@ -47,34 +44,39 @@ class Application < Force::ObjectBase
...
-      { custom_api: 'amiChartYear', domain: 'ami_chart_year', salesforce: 'AMI_Chart_Year' },
-      { custom_api: 'amiPercentage', domain: 'ami_percentage', salesforce: 'AMI_Percentage' },
       { custom_api: 'shortFormPreferences', domain: 'preferences', salesforce: '?', object: Force::Preference },
       { custom_api: 'primaryApplicant', domain: 'applicant', salesforce: '?', object: Force::ApplicationMember },
       { custom_api: 'alternateContact', domain: 'alternate_contact', salesforce: 'Alternate_Contact', object: Force::ApplicationMember },
       { custom_api: 'householdMembers', domain: 'household_members', salesforce: 'Household_Members', object: Force::ApplicationMember },
       { custom_api: '', domain: 'demographics', salesforce: '?', object: Force::Demographics },
       { custom_api: '', domain: 'applicant', salesforce: 'Applicant' },
       { custom_api: '', domain: 'listing_preference', salesforce: 'Listing_Preference_ID' },
+      # Supplemental Application Fields
```
akegan, line 65: I pulled out the supp app fields and put them in the bottom here. No functionality change. 

While looking at the fixtures we have in tests, I realized that we have a big issue of these supp app -only fields bleeding into the paper app and big object snapshot requests. 

This isn't a particularly helpful start, but hopefully this is at least the beginning of a start to creating a clearer definition of the difference between supp app and shortForm apps. Next steps are actually treating them differently during mapping and separating the salesforce endpoints and cleaning up the supp app fields from the big object snapshots and paper app short form endpoint

james-wills-sf (approved):
@akegan code LGTM and I just tested in the review app and the requests are working properly + saving to salesforce

PR #428, opened 2020-10-09
Load SupplementalApplicationPage via javascript [DAH-42]

james-wills-sf:
[[DAH-42]]
TODO:
- [x] add rspec tests for new controller, remove rspec tests from old controller
- [x] thoroughly test the page, preferences, etc
- [x] fix e2e tests


james-wills-sf:
@akegan okay the only thing I haven't done is fix the e2e tests, but otherwise this pr should be good to review

spec/controllers/api/v1/supplementals_controller_spec.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::SupplementalsController, type: :controller do
+  render_views
+  login_admin
+
+  describe '#show' do
+    let(:expected_supplemental_lease_up_app) { fixture('controllers/supplementals/index.json') }
```
akegan, line 10: can we rename this fixture to go under controllers/api/v1/supplementals?

spec/fixtures/controllers/supplementals/index.json
```diff
@@ -284,107 +300,111 @@
...
       "lease": "a130P000007U6tkQAC",
       "other_assistance_name": null,
       "recipient": "a0n0P00000D3qIZQAZ",
       "recurring_assistance": "Yes",
       "type_of_assistance": "Catholic Charities"
     },
     {
-      "assistance_amount": 1100.0,
-      "id": "a151F000001B3whQAC",
+      "assistance_amount": 1100,
```
akegan: weird, the tests complained to me when I didn't have the .0 in there. Only complained when something else in the test failed, but still made it harder to find the issues. Can we keep the .0? Do the tests complain for you if you have it in there? or is it auto-formatting?

spec/controllers/api/v1/supplementals_controller_spec.rb
akegan: can we rename this fixture to go under controllers/api/v1/supplementals?

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -126,7 +127,15 @@ describe('SupplementalApplicationPage', () => {
...
-    expect(await getWrapper()).toMatchSnapshot()
+    let component
+    await renderer.act(async () => {
+      component = renderer.create(
+        <SupplementalApplicationPage applicationId={getMockApplication().id} />
+      )
+    })
+
+    const tree = component.toJSON()
+    expect(tree).toMatchSnapshot()
```
akegan, line 138: nice that this looks like it got rid of 27k lines in this file

akegan (approved):
I think this looks pretty good. just a couple of questions/comments. QA looks good too. Takes a bit longer to load, but hopefully once we do async load it'll seem faster overall.

Looks like e2e tests are having trouble with waiting for things to load, not sure what the best solution for that is. Maybe wait for one of the selectors to indicate that then the page has loaded?

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
james-wills-sf: @akegan not exactly... I switched it to enzyme mount rendering which added 27k lines and decided that was too many lines so I reverted it in another commit which removed 27k lines

akegan: ah I see. I was just filtered on the most recent commits.

spec/fixtures/controllers/supplementals/index.json
james-wills-sf: @akegan I don't think any tests complained to me either way, but I can add the .0.

Generating these huge fixtures is something I'd like us to come up with a process for, when I update the vcr I just copy the application in the response and paste it into this file. Not ideal when we have to follow specific rules like adding .0 to numbers, maybe we can lint json?

PR #429, opened 2020-10-09
[release PR] Add error messages on request failure [DAH-562]

james-wills-sf:
DAH-562
At least show an error message when a request fails instead of failing silently.

Moving forward we may want to automatically return the user to the home page if it was an auth error, but that's out of scope for this PR

### DIFF
--- a/app/javascript/components/supplemental_application/SupplementalApplicationPage.js
+++ b/app/javascript/components/supplemental_application/SupplementalApplicationPage.js
@@ -251,13 +251,13 @@ class SupplementalApplicationPage extends React.Component {
           },
           leaseSectionState: SHOW_LEASE_STATE
         }))
-      }).finally(() => this.setState({ loading: false }))
-      // TODO: catch and handle errors
+      })
+      .catch(() => Alerts.error())
+      .finally(() => this.setState({ loading: false }))
   }
 
   handleDeleteLease = () => {
     const { application } = this.state
-
     this.setState({ loading: true })
 
     deleteLease(application)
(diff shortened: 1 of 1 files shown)


akegan:
Since this is a functionality change, should we add a ticket to track? We've been lax about what needs tickets and what doesn't. I think I'd lean towards:
- Eng chore-type improvements that add tests, update packages, or make the dev environment easier don't need to be tied to tickets (although they can/should be if they're planned and will take significant work)
- Bug fixes and feature changes should generally have tickets, even if they're made at the same time as the PR. That way it's easier to find the full motivation for what the change was. This also helps make sure that we include testing the tickets in releases and follow our guidelines around having review instructions, etc. 

What are your thoughts?

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -251,13 +251,13 @@ class SupplementalApplicationPage extends React.Component {
           },
           leaseSectionState: SHOW_LEASE_STATE
         }))
-      }).finally(() => this.setState({ loading: false }))
-      // TODO: catch and handle errors
+      })
+      .catch(() => Alerts.error())
+      .finally(() => this.setState({ loading: false }))
```
akegan, line 256: would it be feasible to test this with a couple of unit tests?

akegan (approved):
Was able to test failed lease update by using 9999 as the year (we should probably fix that bug at some point, but not a big deal)

image

Not sure how to test an error in deleting a lease, do you have steps to reproduce?

james-wills-sf:
@akegan Made a ticket: DAH-562

I think unit tests are a good idea, but I think there are enough changes coming to this and the unit test file that I'd rather not write them now when they should be easier to write later

PR #430, opened 2020-10-15
Add icon button popover to status history column DAH-157

akegan:
DAH-157

This PR:
1. Updates styling for the popover to fix arrow issues
1. Adds the new status history button to the status history column
1. Loads status history data on first click
1. Add route to get field update comments by application id, and moved routes around to make them more rest-y


app/assets/stylesheets/popover.scss
```diff
@@ -10,26 +10,38 @@ $arrow-offset: (-$arrow-size / 2);
   border-radius: $popover-border-radius;
   z-index: 1;
   width: 22rem;
+  white-space: normal;
 
-  #arrow {
+  #arrow,
+  #arrow::before {
```
james-wills-sf, line 16: What is the before selector on this block doing?

app/javascript/components/atoms/Icon.js
```diff
@@ -17,4 +16,15 @@ const Icon = ({ icon, size = 'small', alert, success }) => {
   )
 }
 
+Icon.propTypes = {
+  icon: PropTypes.string.isRequired,
+  size: PropTypes.oneOf(['small', 'medium', 'tiny']),
```
james-wills-sf, line 21: 👍

app/javascript/components/atoms/Icon.js
```diff
@@ -17,4 +16,15 @@ const Icon = ({ icon, size = 'small', alert, success }) => {
...
+  alert: PropTypes.bool,
+  success: PropTypes.bool
+}
+
+Icon.defaultProps = {
+  size: 'small'
```
james-wills-sf: nit: let's just use the props destructuring on line 5 to define defaults instead of specifying Icon.defaultProps, that's what I've been trying to stick to for functional components

app/javascript/components/atoms/Icon.js
```diff
@@ -1,13 +1,12 @@
 import React from 'react'
-
 import classNames from 'classnames'
+import PropTypes from 'prop-types'
 
 const Icon = ({ icon, size = 'small', alert, success }) => {
```
james-wills-sf: nit: I know it's not technically your change but do you mind adding `= false` for `alert` and `success`? I think general best practice is for all non-required props to have defaults (even if you're just setting the default to`null`)

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -99,32 +104,33 @@ const LeaseUpApplicationsTable = ({
...
     },
     {
       Header: 'Lease Up Status',
       accessor: 'lease_up_status',
       headerClassName: 'td-min-wide tr-fixed-right',
+      className: 'td-min-wide td-status td-fixed-right',
       Cell: (cell) => <LeaseUpStatusCell cell={cell} onChange={onLeaseUpStatusChange} />
     }
+    // {
+    //   Header: '',
```
james-wills-sf: nit: remove commented-out code

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -99,32 +104,33 @@ const LeaseUpApplicationsTable = ({
...
+    //   headerClassName: 'td-min-wide tr-fixed-right',
+    //   className: 'td-min-wide td-fixed-right',
+    //   Cell: (cell) => <LeaseUpStatusCell cell={cell} onChange={onLeaseUpStatusChange} />
+    // }
   ]
 
   const getTdProps = (state, rowInfo, column, instance) => {
     const attrs = {}
 
     // classes and onClick actions vary depending on the type of column
```
james-wills-sf: nit: can remove "classes and" from the beginning of this comment since we're only changing onclick now.

app/javascript/components/molecules/Popover.js
```diff
@@ -17,23 +18,21 @@ const useClickOutside = (ref, callback) => {
   })
 }
 
-const Popover = ({ buttonElement, children }) => {
+const Popover = ({ buttonElement, children, onButtonClick }) => {
   const [showPopper, setShowPopper] = useState(false)
+  const [referenceElement, setReferenceElement] = useState(null)
```
james-wills-sf: I know you already have `buttonElement` as the prop but `buttonElementRef` might be more descriptive, since popperElement and arrowElement are also references to elements, right?

app/javascript/components/molecules/Popover.js
```diff
@@ -47,20 +46,32 @@ const Popover = ({ buttonElement, children }) => {
...
           {children}
         </div>
       )}
     </>
   )
 }
 
+Popover.propTypes = {
+  buttonElement: PropTypes.func.isRequired,
+  children: PropTypes.element,
```
james-wills-sf: I think you want this to be PropTypes.node

I think the difference is node is anything that can be rendered (string, number, element) but element is just react components.

app/javascript/components/molecules/Popover.js
```diff
@@ -47,20 +46,32 @@ const Popover = ({ buttonElement, children }) => {
...
         <div
           className='popper-container'
-          ref={popperRef}
+          ref={setPopperElement}
           style={styles.popper}
           {...attributes.popper}
         >
-          <div ref={setArrowRef} id='arrow' />
+          <div ref={setArrowElement} style={styles.arrow} id='arrow' />
           {children}
```
james-wills-sf, line 64: Do you know what this looks like when children is null? If it shouldn't be null can you mark children as required in the props?

app/javascript/components/molecules/Popover.js
```diff
@@ -17,23 +18,21 @@ const useClickOutside = (ref, callback) => {
   })
 }
 
-const Popover = ({ buttonElement, children }) => {
+const Popover = ({ buttonElement, children, onButtonClick }) => {
```
james-wills-sf: nit: can you set defaults for non-required props here? Even if it's just onButtonClick = null

app/javascript/components/organisms/StatusHistoryPopover.js
```diff
@@ -0,0 +1,70 @@
...
+
+  const StatusIconButton = ({ onClick, ref }) => (
+    <button ref={ref} onClick={onClick} className='tiny icon margin-left--half'>
+      <Icon icon='list-unordered' size='medium' />
+    </button>
+  )
+  const [data, setData] = useState({ statusItems: [], loading: false })
+
+  const fetchStatusHistory = async () => {
+    try {
```
james-wills-sf: can this be done with promises?

```
setData({ statusItems: data.statusItems, loading: true })
apiService.getFieldUpdateComments(applicationId)
  .then((items) => setData({ statusItems: items, loading: false }))
  .catch(() => setData({ statusItems: data.statusItems, loading: false }))
```

if you still want this to be awaitable you can just return it:
```
// Keeping the async just to label this as a function that returns a promise -- not strictly necessary
const fetchStatusHistory = async () => {
  setData({ statusItems: data.statusItems, loading: true })
  return apiService.getFieldUpdateComments(applicationId)
    .then((items) => setData({ statusItems: items, loading: false }))
    .catch(() => setData({ statusItems: data.statusItems, loading: false }))
}
```

app/javascript/components/organisms/StatusHistoryPopover.js
```diff
@@ -0,0 +1,70 @@
...
+      setData({ statusItems: data.statusItems, loading: true })
+      const items = await apiService.getFieldUpdateComments(applicationId)
+      setData({ statusItems: items, loading: false })
+    } catch (e) {
+      setData({ statusItems: data.statusItems, loading: false })
+    }
+  }
+
+  const onClick = async () => {
+    data.statusItems.length === 0 && (await fetchStatusHistory())
```
james-wills-sf: Do we actually need to `await` the fetchStatusHistory() call? Seems like it's okay for it to complete on its own time?

app/javascript/components/organisms/StatusHistoryPopover.js
```diff
@@ -0,0 +1,70 @@
...
+const StatusHistoryPopover = ({ applicationId }) => {
+  const [showingAllStatuses, setShowingAllStatuses] = useState(false)
+  const onShowHideStatusesToggled = () => setShowingAllStatuses(!showingAllStatuses)
+
+  const StatusIconButton = ({ onClick, ref }) => (
+    <button ref={ref} onClick={onClick} className='tiny icon margin-left--half'>
+      <Icon icon='list-unordered' size='medium' />
+    </button>
+  )
+  const [data, setData] = useState({ statusItems: [], loading: false })
```
james-wills-sf: I think it's common to just call it `[state, setState]` when you're storing everything in one state object.

Only thing to be aware of is it's a little more fragile when stored as a big object, since you need to remember to set all state fields. If you ever just did setData({ loading: true }) it would wipe out your status items state.

app/javascript/components/organisms/StatusHistoryPopover.js
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  const numberOfStatusesToDisplay = showingAllStatuses
+    ? data.statusItems.length
+    : MAX_UPDATES_TO_SHOW_DEFAULT
+
+  return (
+    <Popover buttonElement={StatusIconButton} onButtonClick={onClick}>
+      {data.loading ? (
+        <Loading isLoading />
```
james-wills-sf: I think with loading you can just wrap the whole thing in a loading component, and set isLoading conditionally
```
<Loading isLoading={data.isLoading}>
 <StatusItems ... />
</Loading>
```

app/javascript/components/organisms/StatusHistoryPopover.js
```diff
@@ -0,0 +1,70 @@
...
+      ) : (
+        <>
+          <StatusItems
+            statusItems={data.statusItems}
+            limit={numberOfStatusesToDisplay}
+            height='20rem'
```
james-wills-sf: I'm a little worried about styles getting hidden in render functions. How do you feel about pulling all styling into constants at the top of the file it's used in? That way it's all in one place and would be easy to convert to styledComponents or whatever if we move to something like that.

So this would just be:
```
...imports...

// could also be a function if we're calculating style from props or state
const StatusItemStyle = {
  height: '20rem'
}

...

<StatusItems style={StatusItemStyle}>
```

james-wills-sf: Edit: actually I just realized `height` is a specific prop on StatusItems, so ignore this comment in this case, but are you cool with that pattern for other times we use inline styles?

spec/javascript/components/organisms/StatusHistoryPopover.test.js
```diff
@@ -0,0 +1,62 @@
...
+    await act(async () => await tick)
+    wrapper.update()
+    expect(mockGetFieldUpdateComments).toHaveBeenCalledTimes(1)
+  })
+  test('it only fetches status history on first click', async () => {
+    const wrapper = getWrapper()
+    act(() => {
+      wrapper.find('button').simulate('click')
+    })
+    await act(async () => await tick)
```
james-wills-sf: Do you mind removing these ticks and seeing if tests still pass? I found (I forget which file it was for) that all I needed was act() and update, no tick needed.

james-wills-sf (commented):
@akegan Some little comments, but there are enough of them that I'd like to take another pass!

james-wills-sf:
Just QA'd looks good + updates properly when a status is added but I have two questions:

## 1: Can we have the popover have a max height instead of a set height?
That way single items don't show with so much whitespace
- 2020-10-20 at 11 38 49 AM

## 2: Looks like the dates are cut off in the table + not scrollable
Maybe this is something we address after table styling is in.
- 2020-10-20 at 11 39 35 AM

app/javascript/components/molecules/Popover.js
akegan: it works it just looks empty 
image

app/javascript/components/organisms/StatusHistoryPopover.js
akegan: just figured in this case that the loading and the status items are pretty intertwined where we always want to update 1 with the other. I can split it out if that makes more sense

app/javascript/components/organisms/StatusHistoryPopover.js
akegan: saw this in your PR this morning. will try it

app/javascript/components/organisms/StatusHistoryPopover.js
akegan: yep I think that's reasonable

app/javascript/components/organisms/StatusHistoryPopover.js
james-wills-sf: That makes sense to me, fine to keep it as-is!

spec/javascript/components/organisms/StatusHistoryPopover.test.js
akegan: just tried removing it even after removing the awaits and asyncs in the StatusHistoryPopover file. Needs the extra tick to resolve the promise and get the status items back to be included in the snapshot

app/assets/stylesheets/popover.scss
akegan: mimicking the documentation for the popover here, but it has something to do with the placement and rotation. If I remove or replace all of the arrow::befores with just #arrow it doesn't work

akegan:
@james-wills-sf this is ready for re-review

james-wills-sf:
@akegan code LGTM and I just tried it out and it looks so good!

james-wills-sf (approved):

PR #431, opened 2020-10-15
Load pages via react-router [DAH-42]

james-wills-sf:
[DAH-42]

- Remove bootstrapped data, gather params from paths
- Add react-router package
- Set up LeaseUpApp router

todo: in follow-up PRs
- [ ] change paths to be formatted in the new way. This requires ruby changes so I want to do it in a separate PR
- [ ] connect single page app pages via react-router <Link /> components (currently it's not actually behaving like a single page app)

### DIFF
--- a/app/javascript/routes/LeaseUpRoutes.js
+++ b/app/javascript/routes/LeaseUpRoutes.js
@@ -1,0 +1,33 @@
+import React from 'react'
+// import ReactDOM from 'react-dom'
+import { Switch, Route } from 'react-router-dom'
+
+import LeaseUpApplicationsPage from '~/components/lease_ups/LeaseUpApplicationsPage'
+import SupplementalApplicationPage from '~/components/supplemental_application/SupplementalApplicationPage'
+import LeaseUpListingsPage from '~/components/lease_ups/LeaseUpListingsPage'
+import ApplicationPage from '~/components/applications/ApplicationPage'
+
+const LeaseUpRoutes = () => (
+  <Switch>
+    <Route exact path='/listings/lease-ups/:listingId/applications'>
+      <LeaseUpApplicationsPage />
+    </Route>
+    <Route exact path='/listings/lease-ups'>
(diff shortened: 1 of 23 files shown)


app/javascript/routes/LeaseUpRoutes.js
```diff
@@ -0,0 +1,33 @@
...
+    <Route
+      exact
+      path='/applications/:applicationId/supplementals'
+      render={({ match }) => (
+        // TODO: Convert SupplementalApplicationPage to functional components and use
+        // hooks to get path params.
+        <SupplementalApplicationPage applicationId={match.params.applicationId} />
+      )}
+    />
+    <Route exact path='/applications/:applicationId'>
```
akegan, line 27: hm so this is replacing the routing for viewing paper applications in the pre-lottery section too. I thought that we were re-working urls so that it would be different for viewing apps pre-lottery vs viewing the short form big object snapshots during leaseup.

maybe that's one of your checkboxes for follow-up PRs though

akegan (approved):
QA looks good

app/javascript/routes/LeaseUpRoutes.js
james-wills-sf: @akegan I haven't thought much about it, so thanks for bringing it up.

I think my next PR should be updating the paths, and I think that's a good time to split the pre-lottery section to use its own path.

How do you feel about still loading the pre-lottery apps via LeaseUpRoutes? It would be a separate route and path but would still load the ApplicationPage component. That would definitely be easiest from a technical perspective.

Maybe we re-name LeaseUpApp and LeaseUpRoutes to be more generic in that case

james-wills-sf: I'm going to go ahead and merge, I'll make sure to handle this in the path/url update pr

akegan: sure yeah that sounds reasonable.

PR #432, opened 2020-10-16
atom tests

aekong:
adds unit tests for the following atom components:
* Button
* Column
* FormGroup
* FormGroupCheckboxGroup


spec/javascript/components/atoms/Button.test.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import Button from '~/components/atoms/Button'
+
+describe('Button', () => {
+  test('it renders a button with the correct default properties', () => {
+    const wrapper = shallow(<Button text='This is a button' />)
+    const selector = 'button'
+    expect(wrapper.find(selector).exists()).toBeTruthy()
+    expect(wrapper.find({ type: 'button' })).toHaveLength(1)
```
aekong: @james-wills-sf @akegan i wanted to use `toHaveProperty` instead of the object property selector for `find` but was unsuccessful.  tried
```
expect(wrapper.find(selector)).toHaveProperty('type', 'button')
```
any experience with this method?

spec/javascript/components/atoms/Button.test.js
```diff
@@ -0,0 +1,39 @@
...
+    expect(wrapper.find(selector).hasClass('margin-bottom-none')).toBeFalsy()
+  })
+
+  test('it renders a button with the correct properties', () => {
+    const selector = 'button[type="submit"]'
```
james-wills-sf: In general, prefer to keep selectors really simple and lean on enzyme's utilities to select specific children.

also, instead of storing the selector and doing wrapper.find(selector) in each `expect`, you can store the child wrapper like:

```
const wrapper = shallow(<Button ... />)

const buttonWrapper = findByNameAndProps(wrapper, 'button', { type: 'submit' })

expect(buttonWrapper)...
expect(buttonWrapper)...
expect(buttonWrapper)...
```

That example also uses the findByNameAndProps util I added, but you can just as easily leave out the type in the selector and just add an extra expect to check the type

spec/javascript/components/atoms/Button.test.js
```diff
@@ -0,0 +1,39 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import Button from '~/components/atoms/Button'
+
+describe('Button', () => {
+  test('it renders a button with the correct default properties', () => {
+    const wrapper = shallow(<Button text='This is a button' />)
+    const selector = 'button'
+    expect(wrapper.find(selector).exists()).toBeTruthy()
```
james-wills-sf: nit: prefer `expect(wrapper).toHaveLength(1)` over `expect(wrapper.exists()).toBeTruthy()`

Mostly for consistency but also so that it fails if there are 2 or more.

james-wills-sf (commented):
@aekong I know you said this is verbose but I actually think it's good!

I think you have the right intuition around what to test for a simple component like this:
- Test defaults
- Test how changing props changes the component

I think it's good that you checked every prop, even though it seems redundant, hopefully they'll catch if any of this behavior changes accidentally

spec/javascript/components/atoms/Button.test.js
james-wills-sf: @aekong so here's a general rule to use when writing expects:

1. Anything chained on a wrapper is enzyme + react specific
2. Anything chained after expect() is jest _and therefore not react specific_
   - Won't know about props, state, how to work with wrapper objects

So in your example:
```
    expect(wrapper.find(selector)).toHaveProperty('type', 'button')
```
**wrapper.find(selector)** is the react/enzyme specific piece
**.toHaveProperty('type', 'button')** is the jest piece

So `toHaveProperty` isn't checking for react props, it's probably just looking for regular JSObject properties.

The best way to check for props is like this:
```
expect(wrapper.find(selector).props().type).toEqual('button')
```

Breaking this line down:
**wrapper.find(selector).props().type** is all enzyme/react stuff that resolves to a simple string
 **toEqual('button')** is jest stuff that's just comparing two strings!

james-wills-sf: I should clarify that some jest stuff works fine with wrappers, mostly because enzyme and react are pretty good about using regular javascript arrays and objects where possible

Like `.toHaveLength(x)` is a good method of  checking to see if a wrapper exists, because a single wrapper has a length property = 1, and `wrapper.find()` will return a list of wrappers if multiple are found, which obviously also has a length property

aekong: super helpful, thanks!

spec/javascript/components/atoms/Button.test.js
aekong: also helpful, thanks!

spec/javascript/components/atoms/Button.test.js
aekong: good idea, thanks!

aekong:
awesome, thanks for the tips and review!

spec/javascript/components/atoms/Column.test.js
```diff
@@ -0,0 +1,19 @@
+import React from 'react'
+import { shallow } from 'enzyme'
+import Column from '~/components/atoms/Column'
+
+describe('Column', () => {
+  test('it should render a column with the correct properties', () => {
+    const wrapper = shallow(
+      <Column span='span-test' end form>
+        <div className='some-div'></div>
```
james-wills-sf: super minor nit: components with no children should use the closing shorthand like `<div className='some-div' />`

I should probably add a linting rule for this

james-wills-sf (approved):
@aekong lgtm! One super minor nit.

Thanks for making all of those changes

PR #433, opened 2020-10-19
Add unit dropdown component DAH-139

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAH-139


app/javascript/components/molecules/Dropdown.js
```diff
@@ -20,7 +20,8 @@ const Dropdown = ({
       ...provided,
       padding: 0,
       backgroundColor: (state.isSelected && COLORS.primaryTint) || (state.isFocused && COLORS.vapor)
-    })
+    }),
+    control: () => ({})
```
james-wills-sf: what does this do?

app/javascript/components/molecules/UnitDropdown.js
```diff
@@ -0,0 +1,104 @@
...
+  availableUnits: PropTypes.arrayOf(PropTypes.object),
+  disabled: PropTypes.bool,
+  onChange: PropTypes.func.isRequired,
+  placeholder: PropTypes.string,
+  unit: PropTypes.string,
+  id: PropTypes.string
+}
+
+UnitDropdown.defaultProps = {
+  availableUnits: null,
```
james-wills-sf: I think instead of null you want to set this to `[]` by default since you're calling `.length` and `.map` on it

app/javascript/components/molecules/UnitDropdown.js
```diff
@@ -0,0 +1,104 @@
...
+    'expand'
+  ]
+  const renderUnitToggle = ({ children, getValue, ...props }) => {
+    const val = getValue()[0]
+
+    const callToAction = availableUnits.length > 0 ? placeholder : 'No Units Available'
+    return (
+      <button className={classNames(buttonClasses)} type='button' disabled={disabled} id={id}>
+        <Icon icon='arrow-down' />
+        {val?.unit_number ? val.unit_number : callToAction}
```
james-wills-sf: This should probably be `val?.unit_number ?? callToAction`

With the current version, if a unit number happens to be `0`, this will revert to `callToAction`, which I don't think is the behavior we want.

With `val?.unit_number ?? callToAction` it will only revert to callToAction if `val?.unit_number` is null/undefined.

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -201,13 +203,23 @@ const Lease = ({ form, values, store }) => {
...
-            <SelectField
-              id='lease_assigned_unit'
-              label='Assigned Unit Number'
-              fieldName='lease.unit'
-              options={availableUnitsOptions}
-              disabled={disabled || areNoUnitsAvailable}
-              disabledOptions={areNoUnitsAvailable && noUnitsOptions}
+            <Field
+              name='lease.unit'
+              component={({ input: { onChange }, meta, ...rest }) => {
```
james-wills-sf: nit: it doesn't look like you're using rest here, I think you can leave it out of the function params

spec/javascript/components/molecules/UnitDropdown.test.js
```diff
@@ -0,0 +1,96 @@
...
+import { mount, shallow } from 'enzyme'
+import Select from 'react-select'
+
+import UnitDropdown, { renderUnitOption } from '~/components/molecules/UnitDropdown'
+
+const ON_CHANGE = jest.fn()
+const units = [
+  {
+    id: '1',
+    unit_number: '123',
```
james-wills-sf, line 11: Can we add an example with unit_number = '0'? It's unlikely but doesn't seem impossible, just want to make sure this doesn't break

james-wills-sf (commented):
@software-project changes look good! Just a few comments for a first pass, I haven't QA'd it yet either

app/javascript/components/molecules/UnitDropdown.js
software-project: Good catch

app/javascript/components/molecules/Dropdown.js
software-project: It resets styles for control element.

app/javascript/components/molecules/UnitDropdown.js
software-project: `unit_number` is actually a string, but I'll change it just in case.

james-wills-sf: You're right, and '0' resolves to `true` so should be okay either way

akegan:
@software-project heads up looks like there are merge conflicts now. My bad. Also, can you update the PR title to match our PR name convention (`Something that would make a good summary commit message DAH-[ticket #]`

james-wills-sf (approved):
LGTM except for this one issue I found in qa:

doublescroll

It looks like the element width is greater than the dropdown width, so it creates this scroll. Can you fix that before you merge?

software-project:
@akegan @james-wills-sf should be good now.

akegan:
@software-project speaking of pattern lib changes, I think that the current toolkit.scss file you have doesn't have my changes in it to make the status history column wider. 
image

also, checkboxes are missing 
image

akegan:
otherwise looks good
fixed_unit_dropdown

akegan:
Status dropdown column updates look good, but looks like checkboxes are still missing the checks: 
image

I think that the difference of npr run-script build vs npm start is probably what's going on.

Make sure that you have npm start running in another terminal tab when you run grunt

akegan (approved):
Looks good now with latest changes. I'm going to go ahead and merge it

PR #434, opened 2020-10-20
Update paths and route via JS for lease up [DAH-42]

james-wills-sf:
[DAH-42]

### This is the last PR for this ticket!
I accidentally combined both of the following todos into this ticket:
- update paths
- switch links to use react router components

## Routing changes [[Sheet]](https://docs.google.com/spreadsheets/d/1_L3JU4ibqwWdejojT3qQb2ZzEkLUV98xGVn7Vyswbh8/edit#gid=0)

Page | Old path | New Path | Route (unchanged)
-- | -- | -- | --
Supplemental Application | /applications/{application_id}/supplementals | /lease-ups/applications/{application_id}/supplemental | applications/supplementals#index
Short form application | /applications/{application_id}?lease_up=true | /lease-ups/applications/{application_id} | applications#show
Applications list | /listings/lease-ups/{listing_id}/applications | /lease-ups/listings/{listing_id} | listings/lease_ups/applications#index
Listings list | /listings/lease-ups | /lease-ups/listings | listings/lease_ups#index
Non-lease-up application | /applications/{application_id} | /applications/{application_id} | applications#show

In this PR (unchecked are todos):

- [x] Update routes (in js and ruby) to match the above chart
- [x] Update links to use react router [\<Link />](https://reactrouter.com/web/api/Link) and [\<NavLink />](https://reactrouter.com/web/api/NavLink)  components
- [x] Update `<Breadcrumbs />` and `<TabMenu />` components to support routed links
- [x] Hook up the Application page to use the Single page app version whether in lease up or not
- [x] Remove old routes  
- [x] Fix e2e tests


akegan (commented):
Looks good. One minor note is that the swap of the order of the loading on the application page makes it look a bit wonky 
image
image

james-wills-sf:
@akegan if you're still around this is ready for a re-review

app/controllers/applications/supplementals_controller.rb
```diff
@@ -8,5 +8,8 @@ class SupplementalsController < ApplicationController
     def update
       application_service.update
     end
+
+    def show
```
akegan: wondering why this needs to be here or what it's doing? Maybe a comment could help if it's intentional?

app/javascript/components/molecules/Loading.js
```diff
@@ -1,11 +1,22 @@
...
+}) => {
+  const loaderHeightWrapperStyle = { ...(loaderViewHeight && { height: loaderViewHeight }) }
+  const renderChildren = !isLoading || renderChildrenWhileLoading
   return (
     <div className={'loading-panel' + (isLoading ? ' loading' : '')}>
-      {isLoading ? <Spinner /> : null}
-      {children}
+      {isLoading && (
+        <div style={loaderHeightWrapperStyle}>
+          <Spinner />
```
akegan, line 16: could we use the `<Loading/>` pattern here too? might have same issue with height but would match what we do elsewhere

package.json
```diff
@@ -32,7 +32,7 @@
     "moment-timezone": "^0.5.25",
     "postcss-import": "^12.0.1",
     "prop-types": "^15.7.2",
-    "puppeteer": "^1.14.0",
+    "puppeteer": "^5.3.1",
```
akegan: wow you weren't kidding about being far behind 👀

spec/javascript/components/applications/__snapshots__/ListingApplicationsPage.test.js.snap
```diff
@@ -70,44 +70,48 @@ Array [
...
-        <div
-          className="sk-circle10 sk-circle"
-        />
-        <div
-          className="sk-circle11 sk-circle"
-        />
-        <div
-          className="sk-circle12 sk-circle"
-        />
+          className="spinner sk-fading-circle"
```
akegan, line 76: weird that we have the spinner in the snapshot, is it supposed to be loading? Probably not your fault if it is

akegan (approved):
throwing an approval on here actually since it all looks fine and QA looks good, just had a couple of questions

app/javascript/components/molecules/Loading.js
james-wills-sf: @akegan haha so look at the file name for this...

Turns out I was wrong about `<Loading />`, it still renders the child view even if loading is true, so I modified this file to accept a prop to change that behavior.

akegan: 😂 🤦‍♀️  must be time to sign off for the day

app/controllers/applications/supplementals_controller.rb
james-wills-sf: you're right, this shouldn't be here. I thought I needed it to fix the routes but that was a separate issue

spec/javascript/components/applications/__snapshots__/ListingApplicationsPage.test.js.snap
james-wills-sf: Looks like the spinner was there before this change, see line 73.

My guess is we're missing some combination of `await`, `update()`, or `tick` in the test. But I'm not going to change it right now...

akegan: yep that seems reasonable

PR #436, opened 2020-10-22
Add waitlist and assigned number to parking [DAH-146]

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAH-146

### DIFF
--- a/app/javascript/components/molecules/FormGrid.js
+++ b/app/javascript/components/molecules/FormGrid.js
@@ -22,12 +22,13 @@ FormGrid.Row.defaultProps = {
 
 // Wrapper for each form field.
 // NOTE: Grid is hardcoded. This should not be generic element
-FormGrid.Item = ({ children, small = false, fullWidth = false }) => (
+FormGrid.Item = ({ children, small = false, fullWidth = false, medium = false }) => (
   <div
     className={classNames(
       'form-grid_item column',
-      { 'small-6': !(small || fullWidth) },
+      { 'small-6': !(small || fullWidth || medium) },
       { 'small-3': small },
+      { 'small-4': medium },
       { 'small-12': fullWidth }
     )}
   >
--- a/app/javascript/components/molecules/FormGrid.js
+++ b/app/javascript/components/molecules/FormGrid.js
@@ -1,4 +1,5 @@
 import React from 'react'
+import PropTypes from 'prop-types'
 
 import classNames from 'classnames'
 
@@ -22,7 +23,7 @@ FormGrid.Row.defaultProps = {
 
 // Wrapper for each form field.
 // NOTE: Grid is hardcoded. This should not be generic element
-FormGrid.Item = ({ children, width }) => {
+FormGrid.Item = ({ children = null, width = '50%' }) => {
   let widthStyle
 
   switch (width) {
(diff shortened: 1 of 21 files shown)


app/javascript/components/molecules/FormGrid.js
```diff
@@ -22,12 +22,13 @@ FormGrid.Row.defaultProps = {
 
 // Wrapper for each form field.
 // NOTE: Grid is hardcoded. This should not be generic element
-FormGrid.Item = ({ children, small = false, fullWidth = false }) => (
+FormGrid.Item = ({ children, small = false, fullWidth = false, medium = false }) => (
```
james-wills-sf: We should convert these conflicting booleans into a single string enum

app/javascript/components/molecules/FormGrid.js
```diff
@@ -22,12 +22,13 @@ FormGrid.Row.defaultProps = {
...
   <div
     className={classNames(
       'form-grid_item column',
-      { 'small-6': !(small || fullWidth) },
+      { 'small-6': !(small || fullWidth || medium) },
```
akegan: I'm wondering if we can turn all 3 of these options into one option with proptypes setting the available options as ['small', 'medium', 'fullWidth'] or something like that. I also can't remember what the fullWidth option really does, but here you could set small + medium =true, then what would happen?

akegan (changes requested):
Couple of issues

Right now if you choose yes, then fill out values for cost + parking space number, then switch to "waitlist" or "no" the cost field clears out but the parking space field doesn't
image

I think Chris also had a slightly different vision for how the fields should be lined up:
image

If it's feasible, I think we also wanted to hide the fields for cost + parking space number unless they choose "yes" instead of just disabling the fields.

james-wills-sf (commented):
only the one comment, i see andrea has comments on qa as well

app/javascript/components/molecules/FormGrid.js
```diff
@@ -22,19 +22,26 @@ FormGrid.Row.defaultProps = {
...
-      { 'small-6': !(small || fullWidth || medium) },
-      { 'small-3': small },
-      { 'small-4': medium },
-      { 'small-12': fullWidth }
-    )}
-  >
-    {children}
-  </div>
-)
+FormGrid.Item = ({ children, width }) => {
```
akegan: can you add proptypes to this component that lists out what the options are? here's an example where we specify which values are supported https://github.com/SFDigitalServices/sf-dahlia-lap/blob/main/app/javascript/components/atoms/Button.js#L20

akegan (changes requested):
One minor comment and 2 more QA issues:

1. We need to update the label + placeholder text to match the mock-up that chris made:
image
1. If it's feasible, we should hide the parking cost and the parking space fields unless "BMR parking space assigned" is "Yes", similar to how we disable it now.

james-wills-sf:
@akegan @software-project I can go ahead and make these changes and get this merged today

james-wills-sf:
@akegan this should be ready for another review. I addressed the comments you had and added a test for the parking inputs file

## When parking = yes
- 2020-10-23 at 12 55 13 PM

## When parking = no
- 2020-10-23 at 12 55 18 PM

akegan (approved):

akegan:
looks good to merge! not sure why e2e tests have been so flaky lately. Just kicked them off again...

PR #437, opened 2020-10-22
Fix the way loading looks on the short form page [DAH-42]

james-wills-sf:
Followup on routing change because the loading view didn't look right on the short form page

[DAH-42]

## New loading look on non-lease up app
- 2020-10-22 at 11 33 24 AM

## New loading behavior on lease up snapshot
applicationpageloading


akegan (approved):
Looks good including QA. e2e tests seem to be failing a on a couple of different PRs, I'm wondering if something is up

PR #442, opened 2020-10-26
Show 'None' label when no accessibility requests made [DAH-142]

james-wills-sf:
Followup to [DAH-142] found while writing review instructions

Before when no a11y requests were made, we showed a blank row in the lease table, this didn't look very good, and even though this case isn't specified in the designs I went ahead and replaced it with "None", shown below. I plan on linking chris to this PR to get his sign off before merging.

## Update: Chris signed off
See [slack message](https://sfinnovation.slack.com/archives/CPHDCBBS6/p1603740522017700?thread_ts=1603740440.016900&cid=CPHDCBBS6)

## Old table with blank requests
- 2020-10-26 at 12 24 04 PM

## New table with "None"
- 2020-10-26 at 12 19 02 PM

[DAH-142]: https://sfgovdt.jira.com/browse/DAH-142


akegan:
looks a bit wonky that the rows aren't aligned 
image

akegan:
I guess that's true for the case when there are requests too 
image

akegan:
another minor note is that I think the section headers are supposed to be 14px and they're 13 px, not sure how that matches up to header size

james-wills-sf:
@akegan Okay both of those should be fixed now, and now the "Accessibility Requests" header stays on one line like in the designs

- 2020-10-26 at 1 22 29 PM

EDIT: replaced the screenshot, in original I forgot to change both header text sizes

akegan (approved):
Looks much better now, thanks!

PR #443, opened 2020-10-26
Add new number and input-clear icons DAH-159 DAH-563

akegan:
DAH-159 DAH-563

Pulls in latest icons from Pattern lib and latest pattern lib too.


james-wills-sf (approved):
@akegan lgtm

PR #444, opened 2020-10-26
Add MultiSelect component [DAH-565]

james-wills-sf:
[DAH-565]

Add the multiselect component. I am letting the component be completely controlled by its parents, you can see an example in MultiSelectWrapper. This means react-select won't be doing any magic, it's just controlled by its props.

Fingers crossed this will make state management easier in the parent.

[DAH-565]: https://sfgovdt.jira.com/browse/DAH-565


akegan (approved):
It looks pretty good. 

Only weird thing is that it's very tall? 
image
Maybe styling will work better inline with react-final-form integration and whatnot, just not sure why it's so tall here.

james-wills-sf:
@akegan styles on this have been updated, idk if you wanted to re-review

app/javascript/components/molecules/MultiSelect.js
```diff
@@ -2,21 +2,48 @@ import React from 'react'
 import Select from 'react-select'
 import PropTypes from 'prop-types'
 
+const HEIGHT_NORMAL = 'normal'
+const HEIGHT_SMALL = 'small'
+
+const getStyleOverrides = (controlHeight) => {
+  let controlHeightPx
+  switch (controlHeight) {
+    case HEIGHT_SMALL:
```
akegan, line 11: do we have a case where we have inputs with a height of 38px?

app/assets/stylesheets/multi-select.scss
```diff
@@ -0,0 +1,10 @@
+// Styles for the MultiSelect.js component
+
+.input-custom-height {
+  // Don't override input height and margin values
+  input {
+    height: initial;
```
akegan, line 6: how does this initial prop work?

app/views/pattern_library/index.slim
```diff
@@ -34,6 +34,7 @@
   == pl_component 'TableSimple', { header: 'Name', value: 'Solange', headerB: 'Other', valueB: 'Jay' }
   == pl_component 'TabsMenu', { items: [ { title: 'Short Form Application', url: '/pattern_library?url1=true' }, { title: 'Supplemental Information', url: '/pattern_library?url2=false' }, { title: 'Other Information', url: '/pattern_library?url3=false' }], currentUrl: request.original_fullpath }
   == pl_component 'MultiSelectWrapper', { items: [{value: 'item1', label: 'Item1'}, {value: 'item2', label: 'Item2'}]}
+  == pl_component 'MultiSelectWrapper', { height: 'small', items: [{value: 'item1', label: 'Item1'}, {value: 'item2', label: 'Item2'}]}
```
akegan: this kinda ends up just looking like you accidentally put the same thing twice, do we need to have both in the pattern lib? Or could you have the wrapper have 2 of them, one with each height and labels? I don't think we have a great pattern for showing variations on things unfortunately

akegan (commented):
Not blockers necessarily, just questions

app/javascript/components/molecules/MultiSelect.js
james-wills-sf: Only the inputs inside of the application table.

I think  this is good to keep not because the 38px version will be used, but because it'll be very easy to create inputs of different heights when we do need them

akegan: ok sounds good

app/assets/stylesheets/multi-select.scss
james-wills-sf: we spoke on slack about this, it just sets this value to whatever the default is, as if it had never been set.

app/views/pattern_library/index.slim
james-wills-sf: removed the small variation. I think it's okay to just ignore variations for now until we migrate to storyboard or w/e

PR #445, opened 2020-10-27
Add self-close linting rule for components with no children

aekong:
enforces rule for self-closing components with no children


aekong:
@james-wills-sf is this the correct linting rule you mentioned in https://github.com/SFDigitalServices/sf-dahlia-lap/pull/432#discussion_r512226902?

> components with no children should use the closing shorthand like `<div className='some-div' />`

cc @akegan

.eslintrc.js
```diff
@@ -75,6 +75,10 @@ module.exports = {
     'react/prop-types': 'off',
     'react/display-name': 'off',
     'react/state-in-constructor': 0,
+    'react/self-closing-comp': ['error', {
+      'component': true,
```
james-wills-sf, line 79: since component/html=true is the default, you can omit them and just make the above line: `"react/self-closing-comp": "error"`

james-wills-sf (commented):
@aekong thanks for adding this rule! 

2 questions:
1. why are there 9 commits on this PR, shouldn't there just be one? Will that change once your other PR is merged?
2. Did you run yarn lint:fix? I'd be surprised if this rule isn't broken anywhere in our codebase

akegan:
If there are linting errors that need to be fixed, shouldn't circleCI be failing?

james-wills-sf:
Also just making sure: do you have the pre-commit hook working where it doesn't let you commit if there are any linting errors?

Edit: if you don't just run `overcommit --install`

Again I'm just suspicious that no code in our codebase needed to be fixed after adding this rule

and @akegan , I actually can't remember if we added linting as a CI check, or did we just do it as a pre-commit hook?

.eslintrc.js
aekong: good to know!

aekong:
weird! i already merged that `atom-tests` branch.  i wonder if it's because i branched off my `atom-tests` branch to test the rule out.  i'll make a new branch off `main` and bring these changes in.

yeah, i have precommit set up (think it got set up for free when i set up the project initially).

i'll revisit this later today.  thanks for looking, all 👍

akegan:
@james-wills-sf yep I just double checked and it's in there. We had issues in the past with only having it as a pre-commit hook because it's easy to bypass it plus the pre-commit hook should in theory only check the changed files to speed it up

aekong:
closing in favor of #446

PR #446, opened 2020-10-27
Add linting rule to enforce self closing on components with no children

aekong:


aekong:
@james-wills-sf ran `yarn lint:fix` and there wasn't a report of fixed things.  also, no changed files.  surprise!

james-wills-sf (approved):
@aekong Thanks for double checking, lgtm!

To triple check I checked out this change, added a non-self-closing-brace, and hit save and the lint rule applied, so this is definitely working as expected.

PR #447, opened 2020-10-28
Add combined applicant search functionality DAH-159

akegan:
DAH-159
combined search


akegan:
@james-wills-sf I have a couple of [minor functionality questions](https://sfgovdt.jira.com/browse/DAH-159?focusedCommentId=63396) out to Chris but this should be ready to start reviewing as-is

app/javascript/utils/form/final_form/SearchField.js
```diff
@@ -0,0 +1,77 @@
...
+  >
+    {({ input, meta }) => {
+      const hasValue = !!input.value
+      return (
+        <>
+          <div
+            className={classNames(
+              'search',
+              label && 'form-group',
+              (meta.error && meta.touched && 'error') || ''
```
james-wills-sf: I think the preferred way to use classNames with conditional classes is like this:
```
classNames(
  'search',
  {
    'form-group': label,
    error: meta.error && meta.touched
  }
)
```

edit: preferred meaning I prefer it 😆

app/javascript/utils/form/final_form/SearchField.js
```diff
@@ -0,0 +1,77 @@
+import React from 'react'
+import { Field } from 'react-final-form'
+import classNames from 'classnames'
+import { Label, FieldError, HelpText } from './Field'
+
+export const SearchField = ({
```
james-wills-sf: can you add explicit proptypes for this?

Seems like fields are our most complicated components so any extra documentation helps

james-wills-sf: I think this component should use `export default` since the file is only exporting one component

app/javascript/utils/form/final_form/SearchField.js
```diff
@@ -0,0 +1,77 @@
...
+            />
+            <input
+              {...input}
+              id={id || `form-${fieldName}`}
+              className={
+                classNames(hasValue && 'has-value', meta.error && meta.touched && 'error') || ''
+              }
+              type='text'
+              maxLength={maxLength}
+              // aria-labelledby={ariaLabelledby}
```
james-wills-sf: nit: remove commented-out code

app/javascript/utils/form/final_form/SearchField.js
```diff
@@ -0,0 +1,77 @@
...
+              label={label}
+              id={id || `form-${fieldName}`}
+              fieldName={fieldName}
+              blockNote={blockNote}
+            />
+            <input
+              {...input}
+              id={id || `form-${fieldName}`}
+              className={
+                classNames(hasValue && 'has-value', meta.error && meta.touched && 'error') || ''
```
james-wills-sf: nit: can use classNames object syntax for conditional classes here too

app/services/force/soql/preference_service.rb
```diff
@@ -34,15 +34,31 @@ def buildAppPreferencesFilters(opts)
...
+      search_fields = [
+        "Application__r.Name",
+        "Application__r.Applicant__r.First_Name__c",
+        "Application__r.Applicant__r.Last_Name__c",
+        "Application__r.Applicant__r.Email__c",
+      ]
+      search = []
+      for search_term in search_terms
+        for search_field in search_fields
+          search.push("#{search_field} like '%#{search_term}%'")
```
james-wills-sf: do we have an idea of how much longer the query takes as the number of ORs increases?

Just wondering because if we want to eventually search all fields on the table we'll probably have to switch to sosl since this won't scale well, right?

app/services/force/soql/preference_service.rb
```diff
@@ -52,12 +68,14 @@ def app_preferences_for_listing_query(opts)
       # complexity of the query we are dynamically adding filters
       # within the WHERE clause with the helper function above. The
       # Preference_All_Name and Preference_All_Lottery_Rank fields
-      # were added to bring all prefernces into the same query.
+      # were added to bring all preferences into the same query.
       filters = buildAppPreferencesFilters(opts)
+      search = opts[:search] ? buildAppPreferencesSearch(opts[:search]) : nil
```
james-wills-sf, line 87: if `opts[:search]` is an empty string, this will set `search` to nil, right?

spec/services/force/soql/preference_service_spec.rb
```diff
@@ -23,4 +23,35 @@
...
+        ].join('')
+        expect(filters). to eq(expected_filters)
+      end
+    end
+    it 'creates the expected query string for multiple search terms' do
+      VCR.use_cassette('services/soql/preference_service') do
+        filters = subject.buildAppPreferencesSearch('string1,string2')
+        expected_filters = [
+          "Application__r.Name like '%string1%' OR ",
+          "Application__r.Applicant__r.First_Name__c like '%string1%' OR ",
```
james-wills-sf, line 65: the %'s here are wildcards right? So we're checking if the string is contained, not exact equality?

james-wills-sf (commented):
@akegan Overall interaction is fast and feels great!

Questions from QA:
- quotations break this search. Maybe we should ignore them? Would be nice to make quotations behave the way they do in other searches but that may be a can of worms
- I think if the search field is dirty, the "search" button should be active, even if it's empty.
  - ex: Enter a search query, hit search, delete the search query so the field is empty. It's not super intuitive to go over and click "Filter" to apply the search
- Would be nice if clicking the 'x' focused the search input so you could immediately start typing

app/services/force/soql/preference_service.rb
akegan: I looked into SOSL and it didn't really seem like it had the same ability to do where clauses and reference parents that SOQL does. 

I don't have a sense of how the performance scales except that I tried up to 26 different search phrases with the fields we have now and it didn't seem to impact performance.

spec/services/force/soql/preference_service_spec.rb
akegan: correct, in the actual query, not in the test though.

akegan:
Quotations are a good point. We should also make sure to sanitize the field. Only logged in users have access but this could def lead to sql/soql injection

app/javascript/utils/form/final_form/SearchField.js
akegan: thanks for flagging! I had just copypastad from Field.js. Updated there too

app/services/force/soql/preference_service.rb
akegan: yeah that was the idea

app/services/force/soql/preference_service.rb
```diff
@@ -34,15 +34,37 @@ def buildAppPreferencesFilters(opts)
...
+        "Application__r.Name",
+        "Application__r.Applicant__r.First_Name__c",
+        "Application__r.Applicant__r.Last_Name__c",
+      ]
+      search = []
+      for search_term in search_terms
+        # For each term we join the fields with OR to return a record if it matches
+        # across any of these fields.
+        search_term_clause = []
+        sanitized_term = ActiveRecord::Base.sanitize_sql_like search_term
```
akegan, line 66: tbh this doesn't seem to do much, but at least it escapes quotes and %'s so hopefully it helps

app/javascript/components/lease_ups/leaseUpActions.js
```diff
@@ -5,13 +5,30 @@ import { performInSequence } from '~/utils/promiseUtils'
...
-  })
+export const convertToCommaSeparatedList = (str) =>
+  // Take a series of phrases from the search box and turn it into a comma separated list.
+  str
+    ?.split(/ +/)
+    .filter((x) => x !== '')
+    .join(',')
+
+export const sanitizeAndFormatSearch = (str) => {
+  return convertToCommaSeparatedList(str.replace(/["']/g, ''))
```
akegan, line 16: this strips out `" `and `'`s. Let me know if there's anything else you think we should strip out

app/javascript/components/pattern_library/wrappers/MultiSelectWrapper.js
```diff
@@ -24,7 +24,7 @@ const SelectItemShape = PropTypes.shape({
 MultiSelectWrapper.propTypes = {
   items: PropTypes.arrayOf(SelectItemShape),
   disabled: PropTypes.bool,
-  height: PropTypes.oneOf('normal', 'small')
+  height: PropTypes.oneOf(['normal', 'small'])
```
akegan, line 27: not really in the scope of this PR, but I saw a warning in the console so I fixed this

app/javascript/utils/form/final_form/Field.js
```diff
@@ -208,7 +203,7 @@ export const SelectField = ({
         disabled && disabledOptions ? disabledOptions : labelize(options, {}, noPlaceholder)
       return (
         <>
-          <div className={classNames('form-group', (meta.error && meta.touched && 'error') || '')}>
+          <div className={classNames('form-group', { error: meta.error && meta.touched })}>
```
akegan, line 206: went ahead an fixed these across all of the fields while I was at it

akegan (commented):
@james-wills-sf this should (finally) be ready for re-review

app/javascript/components/pattern_library/wrappers/MultiSelectWrapper.js
james-wills-sf: good catch, thanks

app/javascript/components/lease_ups/leaseUpActions.js
james-wills-sf: I don't know much about sanitizing inputs, so I did some googling and found a few opinions like this: https://kevinsmith.io/sanitize-your-inputs

Which suggest validation instead of sanitization, like our search wouldn't allow you to type quotes. The idea is that this way the user knows what's happening to their input, whereas sanitization changes the query behind the scenes.

Just something to think about, if we're already sanitizing on the ruby side using a built in function to protect against bad soql queries I think we should be good to leave it like this

james-wills-sf (approved):
@akegan changes and behavior seem really good to me, lgtm

PR #448, opened 2020-10-28
Fix label for confirmed annual income field on supp app DAH-89

akegan:
DAH-89 fix for the release


akegan:
@james-wills-sf I updated this based on the label discussion after standup today. should be ready for review now.

james-wills-sf (approved):
@akegan lgtm!

PR #449, opened 2020-10-28
Fix release script for 8+ digit SHAs

james-wills-sf:
3 small changes:
- Script stopped working because our commit sha's are 8 digits, not 7. Changed the regex to be 7+ digits
- when no ticket or PR number in a commit title, we weren't getting or printing the title correctly
   - fixed a regex here
   - added a condition to print just the title for the release


akegan (approved):
looks fine!

PR #450, opened 2020-10-28
Add Household size column and filter [DAH-160]

james-wills-sf:
[DAH-160]
- 2020-10-28 at 4 29 59 PM

A few open questions/things to resolve once all filters are in their new location and we're working on styling:
- Currently the label on a selectfield is just the first option. I don't think it makes sense here for the first option to be "HH Members" it should be "Any Household members" to match the other fields.
- HH column is too wide right now

[DAH-160]: https://sfgovdt.jira.com/browse/DAH-160


james-wills-sf:
Update: I tried using `td-min-narrow` as the custom classname but it didn't change as it only sets min-width. I think this would probably need a custom class that sets the flex shrink or flex grow values.

I say unless we plan on releasing this change before the styling changes, we just hold off on worrying about styling until we're working on styling the entire table

akegan (approved):
This looks good except for some funny behavior when we get to page 6 (new server page)

I tested on main too and it looks like this was already a bug so I created DAH-598. Could have been a bug since the beginning, kinda need a big listing to have the filter results go over 5 pages so it's not totally surprising that it hasn't come up.

james-wills-sf:
@akegan good catch on the bug, definitely something we should fix as part of rev3. I'm going to go ahead and merge this PR

PR #455, opened 2020-11-03
[Linting] Fix jump-to-definition and auto importing

james-wills-sf:
This change does the following:
- fixes jump to definition for imports that start with `components/*`
- Linting errors on bad imports
- Linting errors on '~/*` imports
   - Why? Because you could do `~/components/*` or `components/*`, this keeps it consistent
- Adds defaults for autoimports
- Adds sorting for imports
   - Why? This + auto imports extension means you don't have to think about your imports anymore
   - Also groups imports between react, external, internal


akegan:
probably worth adding an extensions.json file (https://code.visualstudio.com/docs/editor/extension-gallery#_workspace-recommended-extensions) with recommended extensions in addition to adding it to the readme. 

Looking forward to having import sorting! and better auto-importing.

james-wills-sf:
@akegan added recommended extensions file and updated readme

akegan:
thanks! I should probably test this locally with my vscode before I approve. I can do that tomorrow morning

akegan (approved):
Confirmed that it recommends suggested extensions 
image

Confirmed that now my imports autosort and have linting errors if they aren't in the right order and also that autoimporting works much better now!!

This is great, thanks!

PR #457, opened 2020-11-10
Use status updated date from application instead of doing separate query DAH-591

akegan:
DAH-591

The code is done for this, but there is data missing from before the field was added (around 10/8). We probably shouldn't merge + deploy until we get that sorted so there isn't missing data on prod

[working with larry on slack](https://sfinnovation.slack.com/archives/CPZHGET3R/p1605050178007500) to figure out how to resolve the data issue.


james-wills-sf (approved):
@akegan code lgtm. Tried it out and the request feels a little faster but you're right, I'm seeing last updated = "none" for every row. Probably should QA again after the data is backfilled

akegan:
update on this: Larry is working on the data fixes. We're going to test in full before going to prod

akegan:
@james-wills-sf this is ready for QA again now that Larry has updated the data. He only updated 1400 MISSION UNIT 1100 so you'll have to QA with that listing

james-wills-sf (approved):
@akegan looks good!

PR #458, opened 2020-11-10
Show hide filters interaction [DAH-568]

james-wills-sf:
[DAH-568]
Also closes [DAH-161]

includes pattern library changes from https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/159

showhidefilters2

- 2020-11-10 at 3 45 41 PM

[DAH-568]: https://sfgovdt.jira.com/browse/DAH-568

[DAH-161]: https://sfgovdt.jira.com/browse/DAH-161


app/javascript/components/atoms/colors.js
```diff
@@ -34,5 +34,6 @@ export const COLORS = {
   success: '#2e8540',
   successTint: '#b4e5be',
   warn: '#fdb81e',
-  warnTint: '#fee8b6'
+  warnTint: '#fee8b6',
+  whiteSmoke: '#f5f5f5'
```
akegan, line 38: is whitesmoke one that I missed from pattern lib or a new color?

software-project (approved):
It looks good to me. Only case where filter state might be a little misleading is when you change the value of a select and change it back to a previous one. Apply button is blue but technically filters are not changed. But I don't think it's a big deal. All in all nice work 👍

akegan (changes requested):
Looks good, one thing is that the apply and clear are supposed to be aligned right I think.

Should we fix up styling for the search in this PR too or wait until we do bulk actions? Looks like the input needs to be a little taller unfortunately (or the buttons a bit smaller) and maybe the search button is supposed to have the blue/gray behavior instead of being disabled when there isn't a change? It also looks like the search button and the show/hide filters button have different heights.

Also wondering when we should add the labels for the filter fields, maybe when we switch to multi-selects? Just want to make sure that those changes don't fall through the cracks of the tickets.

app/javascript/components/atoms/colors.js
james-wills-sf: @akegan it existed in the pattern library, I think it wasn't carried over somehow

james-wills-sf:
@akegan I'll take a stab at the styling. Question about your labels comment, I thought I added labels for the filters? Or maybe I'm not sure what you're talking about

akegan:
I just totally missed them 🤦. I was just thinking about the "Any accessibility request" option that was still in there as being a substitute for the label, but we just still need that null label until we do the multi-select.

james-wills-sf:
@akegan okay styling issues should be fixed. I went ahead and switched "disabled" with tertiary-inverse on the search button because that's how it is in the designs.


ONly thing I didn't do anything about is: "Looks good, one thing is that the apply and clear are supposed to be aligned right I think."

I think they are right aligned?

Review app:
- 2020-11-12 at 4 39 29 PM

[Figma designs](https://www.figma.com/file/p3ir6cmCeZwqYNVCoqTpPb/applicant-table-DAH-104?node-id=286%3A3020):
- 2020-11-12 at 4 39 51 PM

akegan:
OK this one was real! They are left aligned for me still, maybe a firefox/chrome difference?
image

Other changes look good except now the clear icon isn't centered. It's absolutely positioned so the margin is hardcoded 
image

just need to change [this variable](https://github.com/SFDigitalServices/sf-dahlia-pattern-library/blob/main/public/toolkit/styles/molecules/_inputs.scss#L389) to 17px (input is now 50px high total, (50-16)/2 = 17px.)

james-wills-sf:
@akegan okay I addressed your comments

The alignment issue was because I used `justifyContent: 'end'` instead of `justifyContent: 'flex-end'`. I looked for a linting rule that would catch this in the future but couldn't find one.

I also fixed the search padding by adding to this pattern library pr: https://github.com/SFDigitalServices/sf-dahlia-pattern-library/pull/160

akegan (approved):
looks good now. thanks for the changes

PR #459, opened 2020-11-18
Add confirmed/invalid icon to lease-up app table [DAH-581]

james-wills-sf:
[DAH-581]

- Add icon to show if pref is confirmed, invalid, or unconfirmed (no icon)
- remove red row styling for invalid prefs

- 2020-11-18 at 3 51 19 PM

[DAH-581]: https://sfgovdt.jira.com/browse/DAH-581


app/javascript/components/lease_ups/application_page/preferenceValidationUtils.js
```diff
@@ -0,0 +1,9 @@
+export const VALIDATION_CONFIRMED = 'Confirmed'
```
akegan: nit. I'd be fine with putting this in the same file as preference rank call just since it doesn't look like we're using them elsewhere? I just get mildly annoyed by having lots of validation files. 

If there are other use cases out there then having this in its own file like this is probably worth it.

akegan (approved):
Looks good. 

One small thing is that I'm not sure what the alignment should be when the cell is this wide
image

Seems a bit hard to connect the checkbox to the preference in that case, but we also are going to make the cell smaller in the styling PR so it's probably fine

app/javascript/components/lease_ups/application_page/preferenceValidationUtils.js
james-wills-sf: yep I can move it back. At one point this was being used in two files but now it's not

james-wills-sf:
@akegan I agree about the justification looking a little weird. Let's keep it as-is for now and we'll mess around with it once the checkbox + row sizing changes are in

PR #460, opened 2020-11-19
Add checkbox column and state to lease up applications page [DAH-598][DAH-564]

akegan:
DAH-564

also fixed [DAH-598]

[DAH-598]: https://sfgovdt.jira.com/browse/DAH-598


app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -81,10 +81,26 @@ const LeaseUpApplicationsPage = () => {
...
+    const emptyCheckboxes = uniqAppIds.reduce((a, b) => {
+      a[b] = false
+      return a
+    }, {})
+    // set state to false for all of the initial applications
+    setBulkCheckboxesState(emptyCheckboxes)
+  }
+
+  const onBulkCheckboxClick = (appId) => {
+    setBulkCheckboxesState({ ...bulkCheckboxesState, [appId]: !bulkCheckboxesState[appId] })
```
akegan, line 101: kept the object method since it seemed like it would maybe take more time to check if value is in set and remove if it is or add it if it isn't. Maybe it would improve performance though?

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -51,6 +51,22 @@ const PreferenceRankCell = ({ cell }) => {
...
+const CheckboxCell = ({ appId, value = false, onClick }) => (
+  <div className='form-group'>
+    <input
+      id={`bulk-action-checkbox-${appId}`}
+      type='checkbox'
+      checked={value}
+      onChange={() => onClick(appId)}
+    />
+    <label
+      style={{ margin: '0px' }}
```
akegan: this is a bit hacky. The way we do checkboxes is that we hide the actual input and render it as a ::before to the label element instead. I kept that pattern but reduced the margin and have an empty label. Not sure if there's a cleaner way to do it 

 Maybe this is worth pulling out to create a pattern for checkbox w/o label?

akegan (commented):
Interestingly enough, the response time on the checkbox is way faster on the review app than it it is locally for me. I have also found that paging between pages on the lease up applications page is crazy slow for me locally but fine when deployed. I wonder if something is just funky for me locally.

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
james-wills-sf: I think this way is fine, probably wouldn't be a noticeable difference either way

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
james-wills-sf: I would love to do this, but currently the css is so entangled with the element types I don't think it'll be easy... I think this hack is okay for now?

james-wills-sf (approved):
@akegan lgtm. I added tests for the checkbox but if you want to go in and add tests for the applicationpage you can.

Btw a one line change fixed the server page bug (DAH-598)!

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -101,26 +101,6 @@ const LeaseUpApplicationsPage = () => {
     setBulkCheckboxesState({ ...bulkCheckboxesState, [appId]: !bulkCheckboxesState[appId] })
   }
 
-  const performAsyncRequest = ({
```
akegan, line 88: nice! glad you got rid of this function, was very unnecessary obfuscation. What was actually causing the bug @james-wills-sf ?

PR #461, opened 2020-11-20
Update application table styling [DAH-162]

james-wills-sf:
[DAH-162]

Page-level changes:
- [done] Bottom prev/next buttons should be on white background
- [done] Show/hide filters button down to status popover button should be aligned with the same padding 

All rows:
- [done] Height: 56px
- [done] Cell padding: 16px (text line-heights all 24px)
- [done] No borders except for header row (see below)
- [done] Row background colors (to indicate app status) stay the same

Header row:
- [done] Change the header titles to match figma
- [done] Height: 56px
- [done] Font: Open Sans semibold, title case (change from Lato all-caps)
- [done] BG color: #F9F9F9
- [done] Border bottom: 1px solid #CCCCCC

Row/cell details:
- [done] Application numbers show as links, not buttons
- [done] Status popover “buttons” to invoke status history popover should not have a border in their default state

Column details:
- [lets talk about this] Columns are still resizeable but the vertical borders are gone
   - I decided to make the tables not resizable because the resize functionality is kind of broken... You can't resize the last column because it goes behind the fixed column
- [done] Minimum widths are set for each column, then the columns can stretch wider to fill the screen (similar to behavior now) 
- [done] For smaller screens, the status column stays pinned and other columns overflow under it and you can scroll to see them (same as current behavior)

- 2020-11-19 at 4 28 29 PM

[DAH-162]: https://sfgovdt.jira.com/browse/DAH-162


cpalmatier:
Hey @james-wills-sf it looks pretty good to me, in general (even the slightly oddball color stuff). one thing i noticed was that at 1440, the cols are wide enough that it invokes the scroll-under behavior. i'm thinking maybe we can cut the min-width on the first name and last name columns such that at 1440 there's no horizontal scrolling. feels like there's enough room that we can cheat those a teensy bit smaller than it appeared in Figma.

also, i was wondering if it's possible to put a light gray border-left on the (fixed) status column when the viewport is narrow enough to trigger the scroll-under behavior, just to give some visual cue that there's info you have to scroll to see, since the scrollbar is below the fold, at least on 1440x900 screens (and is certainly invisible on screens shorter than 900px). thoughts?

also idk if this is in-scope for this PR, but it seems like the "Previous" button in pagination should be styled as a secondary when it's usable (i.e. pages other than 1 when there are 2+ pages) as opposed to tertiary. not a showstopper but helpful for making it more understandable IMO.

anyway thanks for pinging me w/ the review app—much easier to reason about when there's a built thing to play with as opposed to just fiddling in Figma!

james-wills-sf:
@cpalmatier okay almost all of your comments have been addressed.

After about an hour I decided to give up on getting the status column border to show/hide whether or not there's overflow.

Basically there are two solutions:
1. Add a media query and set it to a fixed value
   - Two issues with this: 1) the media query would have to go in the pattern library but all of the widths are in javascript, 2) this is fragile because if we ever change the widths or add a column you'll have to remember to update this value. If you don't the interaction will be very janky
2. Listen to HTML element width changes
   - This is probably the better solution but we'd need to use a third party library like https://github.com/exogen/react-overflow-indicator and I just didn't feel it was worth it just for this. If this ever comes up on another page we should consider changing this too, though!

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -7,33 +7,33 @@ import ReactTable from 'react-table'
...
 
-import StatusDropdown from '../molecules/StatusDropdown'
-import StatusHistoryPopover from '../organisms/StatusHistoryPopover'
+const CELL_PADDING_PX = 16
 
-const LeaseUpStatusCell = ({ cell, onChange }) => {
-  const applicationId = cell.original.application_id
-  const applicationPreferenceId = cell.original.application_preference_id
+const getCellWidth = (baseSizePx, isAtStartOrEnd = false) => {
+  const padding = isAtStartOrEnd ? 1.5 * CELL_PADDING_PX : CELL_PADDING_PX
```
akegan, line 23: can you maybe add an explanatory comment here? I'm not clear on why it's sometimes 1.5 x 16px

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -87,52 +87,79 @@ const LeaseUpApplicationsTable = ({
...
-      className: 'td-min-wide td-status td-fixed-right',
-      Cell: (cell) => <LeaseUpStatusCell cell={cell} onChange={onLeaseUpStatusChange} />
+      paddingLeft: '10rem',
+      className: 'border-left padding-left td-status td-fixed-right',
+      headerClassName: 'border-left padding-left non-resizable tr-fixed-right',
+      minWidth: getCellWidth(192, true),
+      Cell: (cell) => {
+        const {
+          application_id: applicationId,
+          application_preference_id: preferenceId
```
akegan, line 158: minor thing is that in my PR I remove the need for the application preference id, but I will deal with it after this is merged

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -87,52 +87,79 @@ const LeaseUpApplicationsTable = ({
...
     },
     {
-      Header: 'Lease Up Status',
+      Header: 'Status',
       accessor: 'lease_up_status',
-      headerClassName: 'td-min-wide tr-fixed-right',
-      className: 'td-min-wide td-status td-fixed-right',
-      Cell: (cell) => <LeaseUpStatusCell cell={cell} onChange={onLeaseUpStatusChange} />
+      paddingLeft: '10rem',
+      className: 'border-left padding-left td-status td-fixed-right',
```
akegan: can we combine these into one class instead of having so many different classes here?

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -87,52 +87,79 @@ const LeaseUpApplicationsTable = ({
...
+      Cell: textCell
     },
     {
-      Header: 'Lease Up Status',
+      Header: 'Status',
       accessor: 'lease_up_status',
-      headerClassName: 'td-min-wide tr-fixed-right',
-      className: 'td-min-wide td-status td-fixed-right',
-      Cell: (cell) => <LeaseUpStatusCell cell={cell} onChange={onLeaseUpStatusChange} />
+      paddingLeft: '10rem',
```
akegan: having trouble figuring out what this paddingLeft applies to, where is it adding padding?

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
```diff
@@ -87,52 +87,79 @@ const LeaseUpApplicationsTable = ({
...
     },
     {
-      Header: 'Substatus',
+      Header: 'Latest Substatus',
       accessor: 'sub_status',
-      headerClassName: 'td-offset-right',
-      className: 'td-offset-right text-right td-min-wide'
+      className: 'td-offset-right',
+      headerClassName: 'td-offset-right non-resizable',
+      minWidth: getCellWidth(345),
```
akegan: does this include the status column width? should we somehow show that addition or add a comment?

akegan (commented):
Some minor comments in comments and here:

One minor tweak is that the HH column should be left-aligned now:
image

General 100 works because we don't have to add the X or ✅  for confirming preferences 😂

james-wills-sf:
@akegan isn't the hh column left aligned in your screenshot?

akegan:
ohhh it is. It just looked right aligned because the last name column has so much extra space. my bad

app/javascript/components/lease_ups/LeaseUpApplicationsTable.js
james-wills-sf: That was an experiment that I forgot to remove. Deleted.

akegan:
this is also super minor but it seems like some columns stretch and some don't for big screens? Rank and Application number don't seem to grow at all, and HH grows a really small amount. Just seems to maybe have an overall effect of looking a bit uneven at big screen sizes, but maybe that's just me?
odd stretch behavior

image

james-wills-sf:
@akegan okay updated this to address your comments. The only-some-columns-stretch thing was on purpose, since I figured some columns like hh size/ application number, etc. never need to be bigger, but some fields, like name, a11y requests or substatus could benefit from stretching to show more content. 

@cpalmatier do you have opinions on this?

akegan (approved):
Going to go ahead and approve this because it looks good so you can get it merged. We can always tweak the widths and width behavior in a future PR

PR #462, opened 2020-11-23
Add bulk edit functionality to the lease up applications table DAH-564

akegan:
DAH-564

Remaining items:

- [x] Fix individual status change missing application id
- [x] Reset bulk status update dropdown on submit - Has minor issue, but close enough for now.
- [x] Add top bulk checkbox select behavior with indeterminate checkbox state (in https://github.com/SFDigitalServices/sf-dahlia-lap/pull/463)
- [x] Get latest main for pattern lib once all 3 current PRs are merged
- [x] Speed up field update comment requests - Spend 30 mins investigating if we can do bulk creates with restforce
   - I don't think speeding this up with a batch request is going to work for the current version of the restforce gem. We need support for [composite requests](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_create.htm) for example like the proposed solution [here](https://github.com/restforce/restforce/pull/360), but neither of those are implemented in restforce yet.
- [x] Add tests: 
  - [x] Unit test for StatusModalWrapper because of new Bulk update count subtitle message
  - [x] unit test for status dropdown changes
  - [x] unit test for Button changes
  - [x] Unit tests for LeaseUpApplicationsPage:
      - [x] Checking and unchecking checkboxes
      - [x] Using bulk change dropdown to open up modal
      - [x] Maybe testing that requests are fired off for all app ids on status modal submit from bulk update
  - [x] E2E test for bulk update functionality -> this could potentially cause collisions with e2e test that updates the status of the 1st row. Maybe just test the 2nd + 3rd row or something? 
- [x]  Solve the error case where some of the updates fail:
  - I think we're supposed to change the alert message to something like "x requests failed". Ideally we'd then update the state for the successful ones to reflect that they were successful. That's why I was originally chaining the state change to the request directly. Maybe now that I'm using the (prevState) => {..prevState, ...newState} model that method could work again 
- [x] Consider how we can move the [checkbox styling](https://github.com/SFDigitalServices/sf-dahlia-lap/pull/462/files#diff-415ea40fd76dfc98fa2237bc275e371b88ada7fde13c54cee32464746c7edf6cR81-R86) into the pattern lib or use utility classes to reproduce -> **Not going to bother**
- [x] Add extensive review instrux to the ticket, including making sure close functionality of modal + individual status update still works


akegan:
@james-wills-sf do you want to do a quick QA gut check on the review app of the speed of the bulk requests to see what you think?

james-wills-sf:
@akegan just qa'd and the speed seems reasonable to me for a bulk request. I'd guess a batched version would be super fast in comparison though so if that's easy (as in just a new ruby endpoint, no Salesforce changes) I would still prefer that

Only thing I'm a little concerned about is the load issues of sending X requests all at once. The max is 20 which seems okay, there's no way to increase the table size to 100 is there? if there is we might want to run them in batches of 10 or so

akegan:
No way to do more than 20 at a time, not sure what trying to do it in bulk in rails would take/if it's even possible without Larry adding a new API endpoint. Maybe I can spend a little bit of time looking into it before we merge and if it'd be a big lift maybe we can spin it off into another ticket.

akegan:
@james-wills-sf the other piece missing here is the error case where some of the updates fail. I think we're supposed to change the alert message to something like "x requests failed". Ideally we'd then update the state for the successful ones to reflect that they were successful. That's why I was originally chaining the state change to the request directly. Maybe now that I'm using the (prevState) => {..prevState, ...newState} model that method could work again 🤔

akegan:
@james-wills-sf I updated the PR description with the pieces left on this PR, including a rough outline of my plans for tests. No worries if you can't get it all sorted out tomorrow, we can wrap up on Monday

james-wills-sf:
@akegan sounds good. Have a good weekend!

james-wills-sf:
Okay I don't think speeding this up with a batch request is going to work for the current verison of the restforce gem. We need support for [composite requests](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/resources_composite_sobjects_collections_create.htm) for example like the proposed solution [here](https://github.com/restforce/restforce/pull/360), but neither of those are implemented in restforce yet.

akegan:
OK sounds like something we can make a follow-up ticket for with that info in case it's worth figuring out how to do it outside of the restforce gem somehow? Or at least outside of the restforce built-in REST API functionality. We can probably just generate an arbitrary request and send it the same way we send our custom API requests.

app/javascript/components/atoms/Button.js
```diff
@@ -89,13 +90,15 @@ const Button = ({
         {wrapWithStyle(text, textStyles(!!iconLeft, !!iconRight))}
         {wrapWithStyle(iconRight, iconWrapperStyles())}
       </div>
+      {children}
```
james-wills-sf: I think it makes more sense for {children} to go next to the text

app/javascript/components/atoms/UnlabeledCheckbox.js
```diff
@@ -0,0 +1,29 @@
+import React from 'react'
+
+import classNames from 'classnames'
+import { PropTypes } from 'prop-types'
+
+const UnlabeledCheckbox = ({ id, checked = false, indeterminate = false, onClick = () => {} }) => (
```
james-wills-sf: I think we should just call this component `Checkbox`. At the moment it's unlabeled but to me it makes sense for it to extendable with a "label" prop

app/javascript/components/lease_ups/LeaseUpApplicationsFilterContainer.js
```diff
@@ -46,13 +57,50 @@ const LeaseUpApplicationsFilterContainer = ({ onSubmit, preferences = [], loadin
...
   return (
     <Loading isLoading={loading}>
       <Form
         onSubmit={handleFormSubmit}
         render={({ form, handleSubmit }) => (
           <form style={styles.marginBottomZero} onSubmit={handleSubmit} noValidate>
             <div className='filter-row'>
+              <div className='filter-group filter-group--left'>
+                <div
+                  style={{
```
james-wills-sf: let's keep multi-line custom styles at the top of component files

app/javascript/components/lease_ups/LeaseUpApplicationsFilterContainer.js
```diff
@@ -1,13 +1,16 @@
 import React, { useState } from 'react'
 
 import classNames from 'classnames'
+import { PropTypes } from 'prop-types'
```
james-wills-sf: this should be `import PropTypes` not `import { PropTypes }`

```suggestion
import PropTypes from 'prop-types'
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -156,29 +157,42 @@ const LeaseUpApplicationsPage = () => {
...
-              moment().format(SALESFORCE_DATE_FORMAT)
-            )
-            updateResults(`[${index}]['sub_status']`, subStatus)
-          }
-        })
+  const handleStatusModalSubmit = async (submittedValues) => {
+    updateStatusModal({ loading: true })
+    const { applicationIds, isBulkChange } = statusModalState
+    const { status, subStatus } = submittedValues
+    var comment = submittedValues.comment && submittedValues.comment.trim()
```
james-wills-sf: nit: should be const
```suggestion
    const comment = submittedValues.comment && submittedValues.comment.trim()
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -156,29 +157,42 @@ const LeaseUpApplicationsPage = () => {
...
+    const data = {
+      applicationIds,
+      status,
+      comment,
+      ...(subStatus ? { subStatus } : {}),
```
james-wills-sf: nit: I think this can just be `...(subStatus && { substatus }),`, it's okay to do ...(undefined)
```suggestion
      ...(subStatus && { substatus }),
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -156,29 +157,42 @@ const LeaseUpApplicationsPage = () => {
...
+      isBulkChange: isBulkChange
+    }
 
+    createStatusUpdates(data)
+  }
```
james-wills-sf, line 173:
```suggestion
  }
  
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -156,29 +157,42 @@ const LeaseUpApplicationsPage = () => {
...
+    applicationIds,
+    comment,
+    status,
+    subStatus,
+    isBulkChange
+  }) => {
+    const statusUpdateRequests = applicationIds.map((appId) =>
+      createFieldUpdateComment(appId, status, comment, subStatus)
+    )
+    Promise.all(statusUpdateRequests)
```
james-wills-sf: if this is an async function you should return this promise
```suggestion
    return Promise.all(statusUpdateRequests)
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -190,36 +204,71 @@ const LeaseUpApplicationsPage = () => {
...
+      isOpen: false,
+      showAlert: false,
+      bulkUpdateCount: null,
+      applicationId: null
+    })
+  }
+  const handleLeaseUpStatusChange = (value, applicationId) => {
+    const appsToUpdate = applicationId
+      ? [applicationId]
+      : Object.keys(bulkCheckboxesState).filter((id) => bulkCheckboxesState[id])
```
james-wills-sf: can use entries instead

```suggestion
      : Object.entries(bulkCheckboxesState).filter(([id, checked]) => checked)
```

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -190,36 +204,71 @@ const LeaseUpApplicationsPage = () => {
...
-    setStateWithPrev((prevState) => ({
-      applications: set(clone(prevState.applications), path, value)
-    }))
+  // Updated the visible status, substatus, and status last updated for one or many applications
+  const updateApplicationState = (applicationIds, status, subStatus) => {
+    const updatedApplications = state.applications.map((app) =>
+      applicationIds.includes(app.application_id)
+        ? {
+            ...app,
+            ...{
```
james-wills-sf: nit: don't need the spread here, just do the assignment directly

app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
```diff
@@ -86,28 +57,38 @@ const LeaseUpTableContainer = ({
...
+        loading={statusModal.loading}
+        onClose={onCloseStatusModal}
+        onSubmit={onSubmitStatusModal}
+        showAlert={statusModal.showAlert}
+        status={statusModal.status}
         submitButton='Update'
-        onSubmit={createStatusUpdate}
-        onClose={closeStatusModal}
+        title='Update Status'
+        bulkUpdateCount={statusModal.isBulkChange ? statusModal.applicationIds.length : null}
```
james-wills-sf: nit: can this just be number of ids to change? And we decide which copy to show if it's 1 or not?

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -190,36 +204,71 @@ const LeaseUpApplicationsPage = () => {
...
-      statusModal: {
-        ...clone(prevState.statusModal),
-        ...values
-      }
-    }))
+  const handleCloseStatusModal = () => {
+    updateStatusModal({
+      isOpen: false,
+      showAlert: false,
+      bulkUpdateCount: null,
```
james-wills-sf: nit: we can remove this

app/javascript/components/molecules/StatusDropdown.js
```diff
@@ -18,7 +19,17 @@ export const renderStatusOption = ({ value, label, statusClassName }, { selectVa
...
+const StatusDropdown = ({
+  status,
+  onChange,
+  overrideValue = false,
+  disabled = false,
+  placeholder = 'Status',
+  size = null,
+  expand = false,
+  minWidthPx = null,
+  updateOnChange = true
```
james-wills-sf: it's best practice to have props default to false, let's rename this to `alwaysDisplayPlaceholderText`

app/javascript/components/molecules/StatusDropdown.js
```diff
@@ -29,14 +40,16 @@ const StatusDropdown = ({ status, onChange, disabled, placeholder, size, expand
...
     { small: size === 'small' }
   ]
+
   const renderStatusToggle = ({ children, getValue, ...props }) => {
-    const val = getValue()[0]
+    const val = updateOnChange ? getValue()[0] : null
 
     return (
-      <button
+      <Button
```
james-wills-sf, line 43: I think we can update this usage of button to use the button's icon utils.

app/javascript/packs/pattern_library.js
```diff
@@ -58,6 +60,7 @@ WebpackerReact.setup({ FormGroupRadioGroup })
 WebpackerReact.setup({ FormGroupTextArea })
 WebpackerReact.setup({ FormGroupTextInput })
 WebpackerReact.setup({ FormGroupTextValue })
+WebpackerReact.setup({ UnlabeledCheckbox })
```
james-wills-sf: I think the checkbox component can be removed here
```suggestion
WebpackerReact.setup({ FormGroupTextValue })
```

james-wills-sf (commented):
Added some review comments

app/javascript/components/lease_ups/LeaseUpApplicationsFilterContainer.js
```diff
@@ -46,13 +57,50 @@ const LeaseUpApplicationsFilterContainer = ({ onSubmit, preferences = [], loadin
...
+                  <UnlabeledCheckbox
+                    id='bulk-edit-controller'
+                    indeterminate={!allChecked && numChecked > 0}
+                    checked={allChecked}
+                    onClick={handleCheckboxClicked}
+                  />
+                </div>
+                <div className='filter-group_action'>
+                  <StatusDropdown
+                    onChange={(val) => onBulkLeaseUpStatusChange(val, null)}
```
james-wills-sf: I think it makes more sense for this to just pass the status, it shouldn't need to know to pass null. That can be moved to the parent LeaseUpApplicationsTableContainer

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -156,29 +157,43 @@ const LeaseUpApplicationsPage = () => {
...
+  const handleStatusModalSubmit = async (submittedValues) => {
+    updateStatusModal({ loading: true })
+    const { applicationIds, isBulkChange } = statusModalState
+    const { status, subStatus } = submittedValues
+    var comment = submittedValues.comment && submittedValues.comment.trim()
+    const data = {
+      applicationIds,
+      status,
+      comment,
+      ...(subStatus && { substatus }),
```
james-wills-sf: Edit: actually this can just be `subStatus,`, because we do a null check in the apiservice.

james-wills-sf:
@akegan Didn't get to everything, but I spent the day doing the following
- reviewed the PR and made the changes (I left all of my comments on this PR so you can go back and change whatever you disagree with if you want)
- Fixed some tests
- Added some tests
- Added error functionality. New behavior is where the alert says how many failed, updates successful bulk IDs, and clicking retry will only retry for failed ids
   - While typing this I realized there is one edge case I didn't handle. If you have some bulk checkboxes selected, then change the status for a single application (through the fixed sidebar), it will deselect the checkbox for that application on success. This is a nice-to-have but I think we probably want it to not affect the bulk checkboxes at all. Might need to re-add the `isBulkAction` boolean I removed to be able to tell the difference 😬

app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
akegan: you can use the checkbox to only update 1 status

app/javascript/components/atoms/UnlabeledCheckbox.js
akegan: OK! we do have labeled checkboxes already that we use throughout the forms, we just don't have a pattern lib component for it

spec/javascript/components/organisms/StatusModalWrapper.test.js
```diff
@@ -37,6 +37,26 @@ describe('StatusModalWrapper', () => {
...
 
+  test('should display a subtitle if numApplicationsToUpdate > 1', () => {
+    wrapper = getWrapper({ numApplicationsToUpdate: 2 })
+
+    expect(wrapper.find(FormModal).prop('subtitle')).toEqual(
+      'Update the status for 2 selected items'
+    )
+  })
+
+  test('should display a subtitle if numApplicationsToUpdate == 1', () => {
```
akegan: this should be "should not display..."
```suggestion
  test('should not display a subtitle if numApplicationsToUpdate == 1', () => {
```

spec/javascript/components/organisms/StatusModalWrapper.test.js
```diff
@@ -37,6 +37,26 @@ describe('StatusModalWrapper', () => {
...
+    wrapper = getWrapper({ numApplicationsToUpdate: 1 })
+
+    expect(wrapper.find(FormModal).prop('subtitle')).toBeFalsy()
+  })
+
+  test('should display a subtitle if numApplicationsToUpdate == 0', () => {
```
akegan: this one should also be "should not display.."``
```suggestion
  test('should not display a subtitle if numApplicationsToUpdate == 0', () => {
```

app/javascript/components/organisms/StatusModalWrapper.js
```diff
@@ -10,27 +11,35 @@ import { TextAreaField, Label, FieldError } from 'utils/form/final_form/Field'
...
-  header,
-  submitButton
+  numApplicationsToUpdate = 1,
+  title = null
 }) => (
   <FormModal
-    header={header}
+    title={title}
+    subtitle={
+      numApplicationsToUpdate > 1
```
akegan: I'm not sure if this is the right logic because you can check only 1 checkbox. That's why I had the bulk update flag in there

akegan (commented):
Status dropdown in the modal is now really wide and I think the bottom margins are bigger now 🤔 
image

app/javascript/components/organisms/StatusModalWrapper.js
james-wills-sf: Yeah sorry I realized that a little too late. Okay so adding back the boolean flag should let us fix this and the other thing I mentioned above, where we don't want to deselect any of the checkboxes if it's _not_ a bulk update operation.

james-wills-sf: I can add it.

spec/javascript/e2e/FlaggedApplicationsShowPage.e2e.js
```diff
@@ -38,7 +38,7 @@ describe('FlaggedApplicationsShowPage', () => {
       await saveButtonList[0].click()
 
       // Expect the row to be closed
-      await page.waitFor(() => !document.querySelector('.rt-expander.-open'))
+      await page.waitForFunction(() => !document.querySelector('.rt-expander.-open'))
```
akegan, line 41: waitFor -> waitForFunction and waitForTimeout turned out to be the only breaking-ish change from upgrading puppeteer

spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
```diff
@@ -0,0 +1,119 @@
...
+          await sharedSteps.getCheckboxVal(page, checkboxById(THIRD_ROW_LEASE_UP_APP_ID))
+        ).toBe(true)
+
+        // Select Processing as bulk status
+        await selectStatusDropdownValue(page, bulkStatusDropdown, statusMenuItem(PROCESSING))
+        // Fill out the status modal and submit
+        await fillOutAndSubmitStatusModal(page)
+
+        // Wait for field update comment requests to complete and fail if they were unsuccessful.
+        // The problem is that if one of these requests does not happen we get an ambiguous timeout error.
```
akegan, line 85: This is sub-optimal. Not sure if there's a way around it

spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
```diff
@@ -0,0 +1,119 @@
...
+        expect(thirdRowUpdatedStatus).toBe(PROCESSING)
+        expect(
+          await sharedSteps.getCheckboxVal(page, checkboxById(SECOND_ROW_LEASE_UP_APP_ID))
+        ).toBe(false)
+        expect(
+          await sharedSteps.getCheckboxVal(page, checkboxById(THIRD_ROW_LEASE_UP_APP_ID))
+        ).toBe(false)
+
+        // Repeat with status Appealed
+        await page.click(checkboxById(SECOND_ROW_LEASE_UP_APP_ID))
```
akegan, line 101: this ends up feeling very repetitive but is maybe worth it to make sure that we're always actually changing the status

spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
```diff
@@ -0,0 +1,119 @@
...
+
+const waitForLeaseUpAppTableToLoad = async (page) => {
+  await sharedSteps.goto(page, `/lease-ups/listings/${LEASE_UP_LISTING_ID}`)
+  await page.waitForSelector('#root')
+  await page.waitForSelector(firstRowStatusDropdown)
+}
+
+describe('LeaseUpApplicationsPage status update', () => {
+  describe('using the individual row status dropdown', () => {
+    test(
```
akegan, line 28: this was copied over from spec/javascript/e2e/LeaseUpPage.e2e.js then refactored significantly

spec/javascript/e2e/SupplementalApplicationPage/confirmation_modal.e2e.js
```diff
@@ -53,18 +53,13 @@ describe('SupplementalApplicationPage confirm modal', () => {
       expect(url).toBe(
         `http://localhost:3000/lease-ups/applications/${LEASE_UP_LISTING_APPLICATION_ID}`
       )
-
-      await testBrowser.close()
```
akegan, line 57: this was just a minor fix to use the functionality that we have to only setup + login to one browser session per file.

spec/javascript/support/puppeteer/steps/leaseUpStatusSteps.js
```diff
@@ -0,0 +1,59 @@
+import sharedSteps from './sharedSteps'
+
+export const openStatusModalSelector = '.form-modal_form_wrapper'
```
akegan, line 3: this ends up being a lot of selectors. Normally with strings I'd use CAPITAL_SNAKE_CASE but so far we haven't used that for selectors and there are SO MANY SELECTORS.

I am also inconsistent about using Selector in the variable honestly just because some of the variable names end up getting too long. 

If we can agree on some conventions I'd be happy to update though.

spec/javascript/support/puppeteer/steps/supplementalApplicationSteps.js
```diff
@@ -1,43 +1,36 @@
...
+  selectSubstatusIfRequired,
+  checkForStatusUpdateSuccess,
+  commentInputSelector,
+  openStatusModalSelector,
+  submitStatusModalSelector,
+  unselectedStatusMenuItem
+} from './leaseUpStatusSteps'
 import sharedSteps from './sharedSteps'
 
 const testStatusModalUpdate = async (page) => {
```
akegan, line 11: This seems weird to totally outsource the expectations and test to another file but 🤷. Instead of totally refactoring I just updated it to use the shared functions + selectors I created

akegan (commented):
E2E test added, with some other refactoring of e2e tests as part of it. 

Turns out we really only have status e2e tests for the lease up applications page so far, but I still made a directory because I'm hoping we might someday add some other tests. 

Edit: Looks like puppeteer update might have broken CircleCI. Going to try to fix by incrementing CACHE_VERSION and if that doesn't work I'll just revert those changes.

app/javascript/components/organisms/StatusModalWrapper.js
```diff
@@ -21,96 +21,75 @@ import FormModal from './FormModal'
...
-                )
-              }}
-            />
-          </FormGrid.Item>
-        </FormGrid.Row>
-        {values.status && LEASE_UP_SUBSTATUS_OPTIONS[values.status] && (
+}) => {
+  const bulkUpdateSubtitle =
+    numApplicationsToUpdate === 1
+      ? 'Update the status for 1 selected item'
```
akegan, line 39: nice catch!

akegan (commented):
added unit tests look good.

spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
james-wills-sf: yeah I think it's okay

spec/javascript/support/puppeteer/steps/leaseUpStatusSteps.js
```diff
@@ -0,0 +1,59 @@
...
+export const selectStatusDropdownValue = async (page, dropdownSelector, valueSelector) => {
+  await page.click(dropdownSelector)
+  await page.waitForSelector('li.dropdown-menu_item')
+  await page.click(valueSelector)
+  await page.waitForSelector(openStatusModalSelector)
+}
+
+export const selectSubstatusIfRequired = async (selectedStatus, page) => {
+  if (
+    selectedStatus.toLowerCase() !== 'processing' &&
```
james-wills-sf, line 19: instead of this are you able to just do a check if the substatus dropdown is present? I could see the business logic for this changing, like we add a new status without a substatus and we wouldn't remember to change this code

james-wills-sf: Oh I see now that this code was just ported over, it's not new code. We can leave it.

james-wills-sf (approved):
e2e tests look good if they pass!

PR #465, opened 2020-11-30
Features/add multiselect to filters DAH-566

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAH-566


software-project:
There is an issue with styling and I'm wondering if we should fix it in pattern library? Otherwise style section of `MultiSelect` would grow and there would be quite a few ifs.

app/javascript/components/lease_ups/LeaseUpApplicationsFilters.js
```diff
@@ -22,16 +22,16 @@ const LeaseUpApplicationsFilters = ({
   preferences = [],
   hasChangedFilters,
   onFilterChange = () => {},
-  onClearFilters = () => {}
+  onClearFilters = () => {},
+  form
```
james-wills-sf: is this prop used?

james-wills-sf:
Few design nits after QAing:
- selects shpould be 48px in height, currently 57px
- margins for multiple items seem different for first two items vs last item
- 2020-11-30 at 4 33 03 PM
- the "any*" options should be removed. There are strange cases where you can actually select and any* option if a different option is already selected
- 2020-11-30 at 4 37 38 PM
- 2020-11-30 at 4 37 45 PM
- it doesn't look like the "Apply filters" button is turning blue when values are modified like it should

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,56 @@
...
+
+import MultiSelect from 'components/molecules/MultiSelect'
+
+import { FieldError, HelpText, Label } from './Field'
+
+const format = (options) => {
+  return options && options.map((option) => (typeof option === 'string' ? option : option.value))
+}
+
+export const MultiSelectField = ({
```
james-wills-sf: prefer to use export default when exporting the main component of a file

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,56 @@
...
+  blockNote,
+  disabled,
+  fieldName,
+  height,
+  helpText,
+  id,
+  label,
+  onChange,
+  options,
```
james-wills-sf, line 22: can you add propTypes for this component? And defaults where they make sense

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,56 @@
...
+            <MultiSelect
+              {...input}
+              options={options}
+              height={height}
+              value={input.value}
+              onChange={(event) => {
+                input.onChange(event)
+                onChange && onChange(event)
+              }}
+              {...(disabled && { disabled })}
```
james-wills-sf: since disabled is a boolean this can just be `disabled={disabled}` right?

james-wills-sf (commented):
@software-project just did a pass on the javascript, I still have to look over the ruby code

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,76 @@
...
+  return options && options.map((option) => (typeof option === 'string' ? option : option.value))
+}
+const HEIGHT_NORMAL = 'normal'
+const HEIGHT_SMALL = 'small'
+
+const MultiSelectField = ({
+  blockNote,
+  disabled = false,
+  fieldName,
+  height = HEIGHT_NORMAL,
```
james-wills-sf: We;re not going to support HEIGHT_SMALL anymore, right? If so I would remove this prop (you can leave it in MultiSelect but just always hardcode the height to "normal" from MultiSelectField.

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,76 @@
...
+import { Field } from 'react-final-form'
+
+import MultiSelect from 'components/molecules/MultiSelect'
+
+import { FieldError, HelpText, Label } from './Field'
+
+const format = (options) => {
+  return options && options.map((option) => (typeof option === 'string' ? option : option.value))
+}
+const HEIGHT_NORMAL = 'normal'
```
james-wills-sf: This might be irrelevent depending on if you remove the height values, but can you export the consts from `MultiSelect.js` and use the imported versions here? since they need to be in sync.

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,76 @@
...
+              label={label}
+              id={id || `form-${fieldName}`}
+              fieldName={fieldName}
+              blockNote={blockNote}
+            />
+            <MultiSelect
+              {...input}
+              options={options}
+              height={height}
+              value={input.value}
```
james-wills-sf: I'm seeing this in the console during QA:
- 2020-12-02 at 1 35 13 PM

I think `input.value` needs to be converted to an array of SelectItemShape before it's passed into the MultiSelect component

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -0,0 +1,76 @@
...
+            <FieldError meta={meta} />
+            {helpText && <HelpText note={helpText} describeId={`describe-${id || fieldName}`} />}
+          </div>
+        </>
+      )
+    }}
+  </Field>
+)
+
+const MultiSelectFieldShape = PropTypes.shape({
```
james-wills-sf: Can we re-use the SelectItemShape from MultiSelect?

app/services/force/soql/preference_service.rb
```diff
@@ -27,22 +27,28 @@ def buildAppPreferencesFilters(opts)
         # This builds the syntax need for soql to do an
         # OR statement when passing the option of `Vision/Hearing`
```
james-wills-sf, line 28: Yeah this comment seems to explain the special Vision/Hearing case you need to handle

app/services/force/soql/preference_service.rb
```diff
@@ -27,22 +27,28 @@ def buildAppPreferencesFilters(opts)
...
+          preferences = opts[:preference].map { |preference| "Preference_All_Name__c like '%#{preference}'" }.join(' or ')
+          filters += "and (#{preferences})"
+        end
+        if opts[:status].present?
+          states = opts[:status].map do |status|
+            'Application__r.Processing_Status__c = ' + (status == 'No Status' ? 'NULL' : "'#{status}'")
+          end.join(' or ')
+          filters += "and (#{states})"
+        end
         filters += "and Application__r.Has_ADA_Priorities_Selected__c INCLUDES (#{accessibility_string}) " if opts[:accessibility].present?
```
james-wills-sf: can we move the accessibility string code closer (lines 27-32) to here just for readability?

spec/services/force/soql/preference_service_spec.rb
```diff
@@ -11,33 +11,60 @@
...
+
+    it 'creates the expected query string when householdmembers has two values' do
+      VCR.use_cassette('services/soql/preference_service') do
+        filters = subject.buildAppPreferencesFilters({ total_household_size: ['1', '5+'] })
+
+        expect(filters).to include('and (Application__r.Total_Household_Size__c = 1 or Application__r.Total_Household_Size__c >= 5)')
+      end
+    end
+
+    it 'creates the expected query string when status has two values' do
```
james-wills-sf, line 53: thanks for adding these. Can you add a test for the case that broke in the old version?

100811373-5d45fc00-33ef-11eb-93ae-4fa989878d7a

app/javascript/components/pattern_library/wrappers/MultiSelectWrapper.js
```diff
@@ -8,13 +8,15 @@ const MultiSelectWrapper = ({ items = [], disabled = false, height = undefined }
...
-      height={height}
-      selectedItems={selectedItems}
-      onChangeValues={(values) => setSelectedItems(values)}
-    />
+    <div className='form-group'>
+      <MultiSelect
+        options={items}
+        disabled={disabled}
+        height={height}
+        selectedItems={selectedItems}
```
james-wills-sf: I think this needs to be updated to use the new 'value' prop

app/javascript/components/molecules/MultiSelect.js
```diff
@@ -21,42 +21,57 @@ const getStyleOverrides = (controlHeight) => {
...
-      value={selectedItems}
-      options={items}
-    />
-  </div>
-)
+  onChange = () => {}
+}) => {
+  let selectedValues
+  if (value && value[0] && !value[0].value) {
+    selectedValues = options.filter((option) => value.includes(option.value))
```
james-wills-sf, line 28: I see now what you're doing to parse the string input. I think this is a fragile way of handling inputs because if your options list is like:
Children_Ages:
[
'1',
'2',
'Over 2'
]

And your `value` prop == `'1, Over 2'`, this code will think '1', '2', and 'Over 2' are selected.

I think it's better to either enforce a specific delimiter type in this component(e.g., `', '`), or do the conversion to SelectedItemShape before passing into this component. I slightly prefer the latter since I don't love one prop that can have multiple forms

james-wills-sf (commented):
@software-project It's looking pretty good! I have a couple code nits and some more QA finds:

Deselecting an item by clicking the "x" on a specific item or hitting backspace decrements the "filters applied" counter properly, but clicking the "x" on the right of the multiselect will remove the options but won't decrement the counter. See the image where no filters are applied but the counter still says 1:
- 2020-12-02 at 1 26 25 PM

Applying "Vision/Hearing" filter only results in an empty list, I don't think this should be the case, as there are many with vision or hearing separately. The rails backend should probably allow for vision or hearing if this option is selected
- 2020-12-02 at 1 26 40 PM

app/javascript/utils/form/final_form/MultiSelectField.js
```diff
@@ -56,21 +52,16 @@ const MultiSelectField = ({
...
-  value: PropTypes.string,
-  label: PropTypes.string
-})
-
 MultiSelectField.propTypes = {
   blockNote: PropTypes.string,
   disabled: PropTypes.bool,
   fieldName: PropTypes.string.isRequired,
-  height: PropTypes.oneOf([HEIGHT_NORMAL, HEIGHT_SMALL]),
+  form: PropTypes.object,
```
james-wills-sf: I don't think form needs a proptype since it's not a prop, right?

app/javascript/components/molecules/MultiSelect.js
software-project: not sure I follow. Component returns an array of strings, so it's formatted nicely for the API. But then react select expects {value, label} object. That functions searches options by value to find the correct one. I don't see a room for error here.

software-project: Value can be `'1'` or `['1', 'Over 2']`

james-wills-sf: Interesting, okay, so it's never a string if it's multiple values?

I still think the issue occurs if the value is `'Over 2'`. In that case `selectedValues` would be ['2', 'Over 2'] since both match with includes.

It should work with `value === option.value` as the check instead. The only thing is now this definitely doesn't work if multiple values are passed in a single comma-delimited string (which is fine if that's never happening

james-wills-sf (approved):
@software-project this lgtm! Can you add review instructions to the ticket?

software-project:
> @software-project this lgtm! Can you add review instructions to the ticket?

Will do

PR #467, opened 2020-12-03
Lease up shared state [DAH-537]

james-wills-sf:
[DAH-537]

This change adds global state to partners, and uses it to share listing and application names across pages as a proof of concept.

### Terms
- **Actions** are events. They're an object that contains a type and an optional data object
   - These are created in the `leaseUpActionCreators.js` file
- **ActionCreators** are functions that create and dispatch actions when something happens.
   - For example, the `supplementalPageLoadComplete` action creator dispatches an action with type `ACTION_TYPE_APPLICATION_LOADED`
   - Right now all of our action creators are pretty simple, but eventually they could fire requests or dispatch multiple actions
- **Reducers** are what actually listen for dispatched actions and make changes to the global state as a result.
   - ex: when the reducer sees `ACTION_TYPE_APPLICATION_LOADED` it will replace the application in the global state with the application packages inside the action data
- **Providers** are what we wrap the parent `<App />` object with so all of the children have access to the action creators and global state.

### Best practices
A lot of this is fuzzy, but here's what I'm thinking:

#### Naming
- **Action types are past-tense nouns.** Think of them how you would name an event, not how you'd name a function.
- **Action creator function names are nouns if they are simple, verbs if they're not.** If an action creator is just wrapping a single action, it's fine if it's named similarly to the action. If it's actively doing things like firing requests, name it with a verb (like `loadPage`). 
   - I read this on some blog, we don't have to stick with it.

#### How to separate state logic
- **Components call the action creators, that's it.**
   - They don't create the action objects themselves
   - They don't package objects into the correct form for the global state
     - Ex: they don't turn the first_name attr on applicant to firstName
   - We can even consider moving requests and loading states into the action creators and global state, respectively. That way the components mostly just have visual logic
- **Action creators do most of the logic**
  - They create the action objects. A single action creator function can even fire multiple actions. Ex: REQUEST_STARTED ... (do request) ... REQUEST_SUCCESSFUL
  - They (can) fire requests
  - They take various shapes of input and package them in a way that's consistent with the global state
    - Sounds like most people do this in the reducer, so we can definitely change this one
- **Reducers just apply state changes**
  - Don't do heavy business logic, they just swap state out for other state

#### When to attach a component to context
Using context in a component does make it harder to test, so it makes sense to have a separation between connected and unconnected components, where the connected component is simple but connects to the state, and passes that state as props to an unconnected component that does more logic. That way you can test all of the functionality of the unconnected comoonetnt without dealing with mocking context or whatever.

### How to actually use the state
Here's a simplified example that shows a component that has a single div that shows the time since last clicked, and when you click it the state updates:

```
// actionCreators.js
export const createActions = (dispatch) => ({
  updateTimeClicked: (application) =>
    dispatch({
      type: ACTION_TYPE_UPDATE_TIME_CLICKED,
      data: {
        currentTime: Time.now()
      }
    })
})


// Reducer.js
const Reducer = (state, action) => {
 switch (action.type) {
    case ACTION_TYPE_UPDATE_TIME_CLICKED:
      return { ...state, currentTime: action.data.currentTime }
  }
}

// Provider.js
export const AppContext = createContext()

const Provider = ({ children }) => {
  const [state, dispatch] = useReducer(Reducer, initialState)
  const actions = createActions(dispatch)
  return <AppContext.Provider value={[state, actions]}>{children}</AppContext.Provider>
}

// App.js
const App = () => {
  <Provider>
    <PageThatShowsTimeLastClicked />
  </Provider>
} 

// PageThatShowsTimeLastClicked.js
const PageThatShowsTimeLastClicked = () => {
  const [state, actions] = useContext(AppContext)
  return <div onClick={() => actions.updateTimeClicked()}>state.currentTime<div/>
}
```

TODO:
- [x] Fix tests
- [x] Check all of the possible weird interactions and states

[DAH-537]: https://sfgovdt.jira.com/browse/DAH-537


app/javascript/utils/customHooks.js
```diff
@@ -33,3 +34,40 @@ export const useQueryParamBoolean = (paramName, defaultValue = false) => {
...
+ *   state is the current state
+ *   setState is a function that either takes a function (prevState) => newStateOverrides or an object
+ *     with the new state overrides. This function will only override the values provided in the new
+ *     state object, it won't override the entire state object
+ *   overrideEntireState is an optional function that behaves the same as the second response param of
+ *     the original useState hook.
+ *
+ * @param {Object} initialStateObject the initial state to set the first time the component is mounted
+ */
+export const useStateObject = (initialStateObject) => {
```
akegan, line 101: handy!

james-wills-sf:
@akegan I still have to fix e2e tests and maybe write some new tests to check the shared state, but it's ready for you to take a look and see if the general structure makes sense to you.

It's basically a simplified version of how redux separates logic, but using react context

akegan:
Is the expectation that requests should cancel as you more between different pages? It might just be a function of single-threading but it seems like if I try to switch back to the lease up applications table before an individual supp app loads that the page takes longer to load because it's executing both requests.

james-wills-sf:
@akegan I brought up a symptom of the same problem on slack last week: https://sfinnovation.slack.com/archives/C015887535W/p1606956521004000 We'll probably need to start cancelling requests like [this blog post](https://juliangaramendy.dev/use-promise-subscription/) describes

The speed thing shouldn't be an issue once we bump threads/workers as it won't need to wait for other requests to finish, but we should be cancelling the requests anyways

Edit: Actually it doesn't look like that solution would cancel the requests, but I think it's okay if we at least stop listening to them

akegan:
OK thanks! Should we create a ticket to follow up?

akegan:
Minor QA note. 

If I:
1. Go to lease-up listings page
1. Click on listing
1. Hit back button
1. Hit forward button

The listing breadcrumbs don't show up (behaves as if hard refresh). Is that expected? Same thing happens with going from lease up applications table to supp app and back and forth using back and forward buttons.

akegan (approved):
I'll likely have to pair on making sure I get the distinctions right between actions and action creators and reducers, but this seems like a great start 😄

james-wills-sf:
@akegan to your QA note: that is supposed to happen, because we clear breadcrumb data for the deeper scope when you go up a level.

I guess the logic doesn't need to be this way though?

I also went ahead and fixed the subscription warnings if you want to take one more look?

akegan:
Seems to still be working fine, looks like a good idea to clean up requests. Not sure how to check to see if subscriptions are really cancelling in the network tab but I'll take your word (+ the code's word) for it

I think it's fine if it's the expected behavior! Just wasn't sure if it was or not.

james-wills-sf:
@akegan so they aren't actually cancelling now, my code changes just make sure that we're not trying to set state on unmounted components.

It looks like we could cancel the requests at the [axios leve](https://www.pluralsight.com/guides/all-need-to-know-about-axios) but my changes here don't do that

PR #469, opened 2020-12-07
Add available DTHP and NRHP counts to lease DAH-141

akegan:
DAH-141


app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -171,13 +199,33 @@ const Lease = ({ form, values, store }) => {
...
           <FormGrid.Item width='25%'>
             <strong className='form-note micro h-caps'>Accessibility</strong>
             <p className='margin-top'>
-              <span className='form-note'>{accessibilityUnits.length}</span>
+              <span className='form-note' id='priority-available-count'>
+                {remainingPriorityUnits.length}
+              </span>
+            </p>
+          </FormGrid.Item>
+          <FormGrid.Item width='25%'>
```
james-wills-sf: nit: It seems like these four could be put into a simple component in this file like:

```
const UnitCountDescription = ({ title, id, value }) => (
          <FormGrid.Item width='25%'>
            <strong className='form-note micro h-caps'>{title}</strong>
            <p className='margin-top'>
              <span className='form-note' id={id}>{value}</span>
            </p>
          </FormGrid.Item>
)
```

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -424,12 +446,89 @@ describe('SupplementalApplicationPage', () => {
...
+    test('it displays "No Units Available" and 0 units count when no units available', async () => {
       const wrapper = await getWrapper(ID_NO_AVAILABLE_UNITS)
       const unitSelect = wrapper.find('#form-lease_unit').first()
 
       expect(unitSelect.exists()).toBeTruthy()
       expect(unitSelect.html().includes('No Units Available')).toBeTruthy()
+      expect(wrapper.find('#unit-available-count').text()).toEqual('0')
+    })
+
+    describe('when no unit is selected', () => {
```
james-wills-sf, line 473: These tests are great!

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -126,18 +127,45 @@ const Lease = ({ form, values, store }) => {
...
+    const prefUsed = confirmedPreferences.find(
+      (pref) => pref.id === form.getState().values.lease?.preference_used
+    )
+    return prefUsed?.preference_name.includes(pref)
+  }
+
+  const unitIsUsedWithPrefElsewhere = (unit, pref) =>
+    unit.application_id &&
+    unit.application_id !== application.id &&
+    unit.preference_used_name?.includes(pref)
```
james-wills-sf: Can you explain the difference between preference_used_name on a unit and preference_used on a preference?

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -126,18 +127,45 @@ const Lease = ({ form, values, store }) => {
...
 
-  const handleDeleteLeaseConfirmed = () => {
-    setShowDeleteConfirmation(false)
-    handleDeleteLease()
-  }
+  const selectedUnit = form.getState().values.lease?.unit
 
   const areNoUnitsAvailable = !availableUnitsOptions.length
 
+  const prefUsedIs = (pref) => {
```
james-wills-sf: nit: prefUsedIs as a name is confusing to me. Is this saying "The preference selected on the current application is using `pref`"? If so I think an even simpler name like `isSelected(pref)` works.

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -126,18 +127,45 @@ const Lease = ({ form, values, store }) => {
...
+    const prefUsed = confirmedPreferences.find(
+      (pref) => pref.id === form.getState().values.lease?.preference_used
+    )
+    return prefUsed?.preference_name.includes(pref)
```
james-wills-sf: nit: `prefUsed?.preference_name.includes(pref)` makes it seem like you want `preference_name` to be nullable, but then the includes call will fail. Depending on if you think it should be nullable, make the whole chain use `?.` or `.`

Also: is there a reason for using includes instead of `===`? Is there ever a case where `pref` is expected to be a substring of `preference_name`?

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -126,18 +127,45 @@ const Lease = ({ form, values, store }) => {
...
+  }
+
+  const unitIsUsedWithPrefElsewhere = (unit, pref) =>
+    unit.application_id &&
+    unit.application_id !== application.id &&
+    unit.preference_used_name?.includes(pref)
+
+  const remainingSetAsidesForPref = (pref) =>
+    totalSetAsidesForPref(listing, pref) -
+    units.filter(
```
james-wills-sf, line 169: I think this could be more readable if this value was pulled out into a const named like `numUnitsUsingPref`. i was just a little confused when I first looked at this

james-wills-sf (commented):
@akegan reviewed most of this and only had a few nit comments. I'll review the merge code changes and QA tomorrow

app/javascript/components/supplemental_application/sections/Lease.js
akegan: preference_used_name is the name of the preference, vs preference_used on the lease is the application preference id, which is different for each application. I added the preference_used_name to the query instead of preference used because then we can figure out if it's DTHP or NRHP for any of the other applications.

Would a comment be helpful somewhere?

app/javascript/components/supplemental_application/sections/Lease.js
akegan: Mostly just laziness wanting to have the strings be "DTHP" and "NRHP" instead of "Displaced Tenant Housing Preference (DTHP)" but since I have a constant now maybe that makes more sense to just do the equality check

Edit: Updated to check against the full preference string

akegan:
Made updates based on review comments and got functionality working again.

Did not sort out a way to get the listing promises to behave nicely yet, which is why spec/javascript/components/supplemental_application/actions.test.js are still failing

app/javascript/apiService.js
```diff
@@ -22,17 +22,10 @@ const getSupplementalApplication = async (applicationId) =>
...
-      `/supplementals/units`,
-      { params: { application_id: applicationId, listing_id: listingId } },
-      true
-    )
-    .then(({ available_units, units }) => ({
-      availableUnits: available_units,
-      units
-    }))
+    .get(`/supplementals/units`, { params: { listing_id: listingId } }, true)
+    .then((r) => r.units)
```
akegan, line 28: We seem to have some cases where apiService returns `{thingName: thing}`, and some where we just return `thing`. I updated this one just return units straight because we're now only returning units and it seems pretty straightforward but @james-wills-sf let me know if you have a strong preference

akegan (commented):
Got things working again, including what I think works for the unit + listing promise. Not sure if there's an easy way to see if they're executing in parallel as expected, but otherwise this is ready for QA

This may be an error from before, but we get this error on the supp app page when adding a lease:
```
console.error node_modules/jest-prop-type-error/index.js:8
Warning: A component is changing an uncontrolled input of type text to be controlled. Input elements should not switch from uncontrolled to controlled (or vice versa). Decide between using a controlled or uncontrolled input element for the lifetime of the component. More info: https://fb.me/react-controlled-components```

Not sure if it makes sense to try to fix it in this PR or in a future one

File: app/javascript/apiService.js (hunk 7 above)
Comment (james-wills-sf) [reply app/javascript/apiService.js:28]: This is good to me. we can default to just returning the object itself unless there's more than one object in the response.

File: app/javascript/components/supplemental_application/sections/Lease.js (hunk 3 above)
Comment (james-wills-sf) [reply app/javascript/components/supplemental_application/sections/Lease.js]: No that's okay, not sure where the comment would go. Thanks for the explanation though!

File: app/controllers/api/v1/supplementals_controller.rb [hunk 8]
```diff
@@ -12,10 +12,15 @@ def show
         render json: {
           application: soql_application_service.application(id, { include_lease: false, include_rental_assistances: false }),
           file_base_url: current_user.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL'],
-          units: units_service.units_and_leases_for_listing(application[:listing_id]),
         }
       end
 
+      def units
+        units = units_service.units_and_leases_for_listing(params[:listing_id])
+        render json: {
```
james-wills-sf: nit: spacing on this block is a little wonky

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -84,7 +84,7 @@ const SupplementalApplicationPage = ({ applicationId }) => {
 
   const [{ breadcrumbData }, actions] = useContext(AppContext)
 
-  useAsyncOnMount(() => getSupplementalPageData(applicationId), {
+  useAsyncOnMount(() => getSupplementalPageData(applicationId, breadcrumbData?.listing.id), {
```
james-wills-sf: same nullable chaining here, I think these all need to be `?.`'s

Edit: looks like this was my change, but still should probably be updated!

app/javascript/components/supplemental_application/actions.js
```diff
@@ -23,10 +22,53 @@ const defaultLeaseResponse = (application) => ({
...
 
-export const getSupplementalPageData = async (applicationId) => {
-  const pageData = await apiService.getSupplementalPageData(applicationId)
-  const listing = await getListing(pageData.application.listing.id)
-  return { ...pageData, listing }
+export const getSupplementalPageData = async (applicationId, listingId = null) => {
+  const applicationPromise = async () => apiService.getSupplementalApplication(applicationId)
+  const unitsPromise = async (nonNullListingId) => apiService.getUnits(nonNullListingId)
+  const listingPromise = async (nonNullListingId) => apiService.getLeaseUpListing(nonNullListingId)
+  const listingAndUnitsPromise = async (nonNullListingId) =>
```
james-wills-sf, line 29: this looks good

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -126,18 +141,47 @@ const Lease = ({ form, values, store }) => {
...
+    ).length
+
+  const remainingSetAsidesForPref = (pref) =>
+    totalSetAsidesForPref(listing, pref) - numUnitsUsingPref(pref)
+
+  // Selected unit does not count in available a11y units.
+  const remainingPriorityUnits = availableUnits.filter(
+    (unit) =>
+      unit.id !== selectedUnit &&
+      unit.priority_type &&
```
james-wills-sf: nit: can be `unit.priority_type?.match(/Mobility|Hearing|Vision/)`

spec/javascript/apiService.test.js
```diff
@@ -22,12 +22,7 @@ const mockSuppAppGetFn = jest.fn(() =>
 )
 const mockGetLeaseFn = jest.fn(() => Promise.resolve({ lease: supplementalApplication.lease }))
 const mockGetStatusHistoryFn = jest.fn(() => Promise.resolve({ data: [] }))
-const mockGetUnitsFn = jest.fn(() =>
-  Promise.resolve({
-    available_units: [{ id: 'availableUnitId' }],
-    units: [{ id: 'unitId' }]
-  })
-)
+const mockGetUnitsFn = jest.fn(() => Promise.resolve({ units: [{ id: 'unitId' }] }))
```
james-wills-sf: this should just return `Promise.resolve([{ id: 'unitId' }])` directly, right? Since you removed the wrapper object

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -356,7 +376,11 @@ describe('SupplementalApplicationPage', () => {
 
       // Lease start date
       act(() => {
-        wrapper.find('#form-lease_unit').find('Select').instance().props.onChange({ value: 'id1' })
+        wrapper
+          .find('#form-lease_unit')
+          .find('Select')
+          .instance()
```
james-wills-sf: nit: I know you didn't actually change this part but lets use `.props()` directly instead of `.instance().props`

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -436,12 +461,89 @@ describe('SupplementalApplicationPage', () => {
...
+    })
+
+    describe('when unit without priority is selected', () => {
+      beforeEach(async () => {
+        wrapper.find('#edit-lease-button').first().simulate('click')
+        act(() => {
+          wrapper
+            .find('#form-lease_unit')
+            .find('Select')
+            .instance()
```
james-wills-sf: nit: `.props()` here too

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -436,12 +461,89 @@ describe('SupplementalApplicationPage', () => {
...
+    })
+
+    describe('when unit with priority is selected', () => {
+      beforeEach(async () => {
+        wrapper.find('#edit-lease-button').first().simulate('click')
+        act(() => {
+          wrapper
+            .find('#form-lease_unit')
+            .find('Select')
+            .instance()
```
james-wills-sf: nit: `.props()`

spec/javascript/fixtures/units.js
```diff
@@ -7,7 +7,7 @@ const units = [
     'property_type': null,
     'ami_chart_type': 'HUD Unadjusted',
     'ami_chart_year': 2017,
-    'reserved_type': null
+    'reserved_type': null,
```
james-wills-sf: I thought our linter didn't allow trailing commas? Idc either way, just surprised this was allowed

james-wills-sf (commented):
@akegan Still need to do more qa before approving but here are comments on the code so you don't have to wait

Also unrelated to this change I saw this application [a0o0P00000ItlpiQAB](https://dahlia-lap-full-pr-469.herokuapp.com/lease-ups/applications/a0o0P00000ItlpiQAB/supplemental) has a broken looking status history:
- 2020-12-10 at 8 39 25 AM

james-wills-sf (approved):
@akegan followed review instructions and QA lgtm

spec/javascript/apiService.test.js
akegan: I think it must be hitting the first promise or something weird because this works and then gives a response of [{id: 'unitId'}] that works in the test below. I was fixing it kinda late yesterday though so I'll look into it again

akegan: Oh I see. It's because the mockGetUnitsFn is mocking the actual http request response, which gives {units:  [units]}. Then we're testing that the apiService fn calls response.units to return just [units]

akegan: seems like all of these functions are actually poorly named and should be mockUnitsRequest or something like that. 

Edit: I went ahead and renamed them to make it clearer

PR #470, opened 2020-12-08
Default to lease-ups after login [DAH-534]

software-project:
Resolves https://sfgovdt.jira.com/browse/DAH-534


akegan (approved):
Looks good! Can you just add to test with both MOHCD admin + leasing agent login to the review instrux? I don't think it should make a difference (and it didn't in my QA) but it seems like a good thing to test

PR #471, opened 2020-12-09
Parallelize supp app requests [DAH-594]

james-wills-sf:
DAH-594

Split the single Supplemental application page data request into the following separate requests:
- application (without lease and assistances)
- lease
- rental assistances
- units
- status history

## Concurrency

This change also updates our puma configs to default to 2 workers and 3 threads each, for 6 concurrent processes total. I chose two workers because each worker will use about 300mb of memory, and our boxes have 1gb memory total so that should leave some room to grow. Our full/review app boxes are hobby boxes so we might need to set the env variables to 1 worker for those.

Also, you're only allowed to set threads > 1 if your rails app is thread safe. I'm not 100% sure that ours is, but I did a search in the codebase for class level fields (`@@fieldName`) or global fields (`$fieldName`) and we never use them, and most local fields (`@fieldName`) are on controllers which are created for each request, or on objects that the controllers create. So I _think_ we should be good. But we'll need to keep an eye out for exceptions or worse, weird race conditions.


app/controllers/api/v1/leases_controller.rb
```diff
@@ -4,6 +4,15 @@ module Api
   module V1
     # RESTful JSON API for lease actions
     class LeasesController < ApiController
+      def index
+        response = soql_lease_service.application_lease(params[:application_id])
+        if response
+          render json: { lease: response }
```
akegan: what's the response if there is no lease?

app/controllers/api/v1/supplementals_controller.rb
```diff
@@ -8,15 +8,20 @@ class SupplementalsController < ApiController
...
-          status_history: field_update_comment_service.status_history_by_application(id),
+          application: soql_application_service.application(id, { include_lease: false, include_rental_assistances: false }),
           file_base_url: current_user.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL'],
-          available_units: units_service.available_units_for_application(application[:listing_id], id),
-          units: soql_listing_service.units(application[:listing_id])
+        }
+      end
+
+      def units
+        application_id = params[:application_id]
```
akegan, line 19: we actually won't need application_id anymore which is nice and should simplify this

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -56,33 +60,38 @@ jest.mock('apiService', () => {
...
-              ]
+        application: { ...appWithPrefs, rental_assistances: null, lease: null },
+        fileBaseUrl: 'fileBaseUrl'
       }
     },
+    getLease: async (applicationId) => {
+      return { ...mockedApplication.lease }
+    },
+    getStatusHistory: async (applicationId) => [],
+    getUnits: async (applicationId, listingId) => ({
```
akegan, line 71: This will probably be a bit of a pain to merge but is OK!

akegan (commented):
Initial comments, looks great! Haven't tested yet. Are the threads updated on the review apps by default?

app/controllers/api/v1/leases_controller.rb
james-wills-sf: Same as before: `{}`

james-wills-sf:
@akegan yep review apps should have 1 worker 3 threads, compared to prod which will be 2 workers 3 threads. Reason for that is amount of ram on `hobby` vs `2x` dynos

akegan (approved):
QA looks good. Ready to merge once you work out what's going on with the lease service

james-wills-sf:
@akegan okay so `{}` was true because it wasn't literally `{}`, it was like `Hashie({})` or something. 

That also means that response would never be false, so I just removed the `if` entirely, we just always return success because like you said if something goes wrong rails will return an error anyways.

PR #472, opened 2020-12-10
Add filtering to lease-ups listings DAH-535

software-project:
Resolves: https://sfgovdt.jira.com/browse/DAH-535


james-wills-sf:
not sure why it's not showing up in the github ui but the review app is here: https://dahlia-lap-full-pr-472.herokuapp.com

app/javascript/utils/reactTableUtils.js
```diff
@@ -14,3 +14,10 @@ const date = (cell) => {
 export const cellFormat = {
   date
 }
+
+export const filterMethod = (filter, row) => {
+  return (
+    // do case insensitive RegExp match instead of the default "startsWith"
+    row[filter.id].match(new RegExp(filter.value, 'ig'))
```
james-wills-sf: I was a ,little confused at this code at first, I thought maybe `row` was actually a map of rowids to rows and we were getting the specific row by the row id, but I think what's actually happening is row is a map of column id to cells?

I think this would be clearer for me if we moved filter.id to a const like `columnId`

james-wills-sf (approved):
@software-project lgtm with one nit, qa looks good too

akegan:
@software-project looks like there was just one more small comment on this PR then we can merge

app/javascript/utils/reactTableUtils.js
james-wills-sf: went ahead and updated this

PR #474, opened 2020-12-11
Persist application table filters and page number [DAH-532]

james-wills-sf:
[DAH-532]

Save filters when jumping in and out of specific supplemental applications

[DAH-532]: https://sfgovdt.jira.com/browse/DAH-532


app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -89,11 +97,52 @@ const LeaseUpApplicationsPage = () => {
...
+      setListingState({
+        reportId: listing.report_id,
+        listingPreferences: getPreferences(listing)
+      })
+
+      actions.applicationsPageLoadComplete(listing)
+    }
+  })
+
+  useAsync(
```
james-wills-sf, line 114: This is basically the same as loadPage, except it's triggered by state change instead of a reactTable callback.

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -89,11 +97,52 @@ const LeaseUpApplicationsPage = () => {
     subStatus: null
   })
 
-  const [{ breadcrumbData }, actions] = useContext(AppContext)
   const isMountedRef = useIsMountedRef()
 
-  // grab the listing id from the url: /lease-ups/listings/:listingId
-  const { listingId } = useParams()
+  const [{ reportId, listingPreferences }, setListingState] = useStateObject({})
+  useAsyncOnMount(() => getListing(listingId), {
```
james-wills-sf, line 103: Moved listing load into it's own asyncOnMount because we shouldn't need to check to see if we need to load the listing every time the table page changes.

akegan:
One initial QA finding: Something a bit funny seemed to maybe be happening with the forward and back button

1. I wasn't able to reproduce it twice, but I clicked into "Marked duplicates" then after it loaded I hit the back button and the url changed to the lease up applications table, but the page didn't change or load.

2. I went from lease up applications table to marked duplicate, then used back button to go back to lease-up applications table. When I hit forward button, this is what I got. Url is correct, but it was trying to load an application? It looks like it's trying to load an application with id = 'flagged'

image
image

james-wills-sf:
@akegan I was able to repro, it's because the LeaseUpRouter thinks that /applications/flagged is an application with id="flagged". Looking into a fix now

Edit: Also looking if there are other urls that would cause similar issues

akegan:
Found another issue, looks like saving what page you're on works for page <5 (first server page), but doesn't really work for page >6 

Steps to reproduce:
1. Go to a big listing (e.g. 1601 Mariposa)
1. Filter (I did live/work preference)
1. Go to page 7 
1. Click into an application
1. Hit the back button

This is what you end up with at the bottom:
image

The results are right, starts at the same number, but then you can't go back or forth between pages. I hit back and it didn't change which applications were displayed and ended up at page 0 of 1 with forward and back disabled. 
image

james-wills-sf:
@akegan Okay I think I fixed both of the issues you brought up. The only issue I can't seem to figure out is:

go to page 6, add filters and apply, see that data is now showing page 0 as expected, but the react table pagination component still shows page = 6. This bug was around before this change, and I think it might be a bug with react-table since I logged the `page` state value and it's updating and passing to react-table correctly.

Edit: I'm guessing [this](https://github.com/tannerlinsley/react-table/commit/f138292c120e46783470eec03b2d5cd75987ac2b) would fix it, but we don't have it in our react-table dependency (I think it's only on the react-table v6 branch)

app/javascript/components/supplemental_application/SupplementalApplicationPage.js
```diff
@@ -58,7 +58,9 @@ const setApplicationsDefaults = (application) => {
   return applicationWithDefaults
 }
 
-const SupplementalApplicationPage = ({ applicationId }) => {
+const SupplementalApplicationPage = () => {
+  // grab the listing id from the url: /lease-ups/applications/:applicationId/supplemental
```
```suggestion
  // grab the application id from the url: /lease-ups/applications/:applicationId/supplemental
```

james-wills-sf:
@akegan I'm going to suggest _not_ fixing the last bug I mentioned in this PR, but I could make a ticket for it. I think the solution is either to switch to react-table-v6 with all of the latest fixes or make the jump to react-table 7+.

akegan:
OK that seems reasonable. If you're up for making a ticket that would be great. I just looked to see if we had one but I don't see one!

james-wills-sf:
ticket added: https://sfgovdt.jira.com/browse/DAH-660

akegan:
Still getting a bit of routing weirdness. Most likely it's from the previous routing PR. Maybe we should just make a ticket for it? 

It's if I go to marked duplicates and hit the back button, the url updates but the page does not.

back_button_weirdness

akegan (approved):
One mild comment typo but everything looks good otherwise! Up for moving the routing bug to a separate ticket.

james-wills-sf:
@akegan yeah I agree that behavior is a little weird, but it's actually default rails routing behavior, to keep you on the current page until the next page finishes loading. In your gif you can see the blue bar at the top is loading the whole time until the duplicate apps page finally appears.

I also just tried this from listings tab -> duplicate apps tab and I get the same thing. That means it's not the react router doing anything weird, it's just the duplicate apps page loads really slowly.

Also, about performance: I just tried to load the duplicate apps page on prod and got a timeout error, twice in a row.. that seems really bad.

Edit: Andrea and I talked about this, I totally missed the point of her gif. It is a real bug, but not caused by this PR. Andrea's going to make a ticket for it.

PR #475, opened 2020-12-12
Add bulk comment action DAH-588

software-project:
Resolves [DAH-588](https://sfgovdt.jira.com/browse/DAH-588)

### DIFF
--- a/spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
+++ b/spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
@@ -16,6 +16,7 @@ import SetupBrowserAndPage from '../../utils/SetupBrowserAndPage'
 let testBrowser
 
 const firstRowStatusDropdown = '.rt-tr-group:first-child .rt-td .dropdown .dropdown-button'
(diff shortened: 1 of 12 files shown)


akegan:
@software-project it looks like when you bulk add a status that the substatus isn't being preserved. Status stays the same but substatus disappears in the table + in the popover status history

software-project:
@akegan checking

akegan (approved):
QA looks good now. Only thing I'm wondering is if it could use another test or 2. Something to catch the issue where the substatus wasn't coming through and getting updated?

spec/javascript/e2e/LeaseUpApplicationsPage/status_update.e2e.js
```diff
@@ -113,6 +113,17 @@ describe('LeaseUpApplicationsPage status update', () => {
         thirdRowUpdatedStatus = await sharedSteps.getText(page, nthRowStatusDropdown(3))
         expect(thirdRowUpdatedStatus).toBe(APPEALED)
 
+        await testBrowser.close()
```
akegan: I think if you close the browser it'll cause issues in the next test

PR #477, opened 2020-12-22
Share state between shortform and supp app [DAH-531]

james-wills-sf:
[DAH-531]

- Replace SupplementalApplicationPage and ApplicationPage (with leaseup=true) with ApplicationDetailsPage
  - ApplicationPage still exists but is only for non-lease up pages 
- Move SupplementalApplicationPage context into AppContext
- Switch tab order (now supplemental comes first)
- Split reducers into separate, independent reducers
  - **Why:** This change grew the reducer by a lot, I find multiple, smaller reducers easier to manage. Let me know if you don't feel the same way, I could see how this could be confusing
  - note that multiple reducers can respond to the same actions
  - Use a single reducer (`Reducer.js`) to combine them all together into a single object
- Stop using fancy action creator pattern, action creators are now just simple functions that you have to pass `dispatch` to
  - **Why:** I found it frustrating to not be able to jump-to-definition for actions, also wanted to be able to split up the action creators files so they're not all in one giant file.
- convert actions from a bunch of consts to one big object
  - **Why:** I didn't like the huge list of imports
-  Remove separate shortform route from ruby and js
  - **Why:** they now share a url and always default to supp app tab.
- Convert error alert into text when shortform load fails
  - **Why:** shortform load fails for any listing that doesn't have a snapshot, and that seems really annoying to get that error on the supp app page
  
TODO:
- [x] Check forward/back button functionality
  - I didn't QA this much and am worried going back and forth won't work properly. 
- [x] Fix supplementalApplicationPage tests
  - Should this be converted to SupplementalContainer tests instead? 
- [ ] Add new tests
  - [x] ApplicationDetailsPage
    - [ ] test switching tabs
    - [ ] Test   
  - [x] for each new action creator
  - [ ] Reducer.test.js for each new action
    - Or should we test each action at the sub reducer level?
  - [ ] For jumping between tabs in   

[DAH-531]: https://sfgovdt.jira.com/browse/DAH-531


app/javascript/components/applications/ApplicationPage.js
```diff
@@ -45,18 +44,21 @@ const buildActionLinkIfNecessary = (app, showAddBtn) => {
...
+ * Non-lease up application page.
+ */
+const ApplicationPage = () => {
   const [application, setApplication] = useState(null)
   const [fileBaseUrl, setFileBaseUrl] = useState(null)
   const [loading, setLoading] = useState(true)
 
   const { applicationId } = useParams()
-  const [{ breadcrumbData }, actions] = useContext(AppContext)
+  const [, dispatch] = useAppContext()
```
james-wills-sf, line 56: Note that now instead of passing actions, we're just passing the dispatch function.

So before you could call `action.doSomething()`, now you need to call `doSomething(dispatch)` and pass it the dispatch. The nice thing about this is you can jump to definition, and these are just regular functions (reduces the amount of magic)

app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -142,99 +147,81 @@ const SupplementalApplicationContainer = ({ store }) => {
...
+            dispatch,
+            submittedValues,
+            convertPercentAndCurrency(form.getState().values),
+            prevApplication,
+            leaseSectionState
+          )
+        }}
+        showAlert={state.statusModal.showAlert}
+        status={state.statusModal.status}
+        submitButton={state.statusModal.isInAddCommentMode ? 'Save' : 'Update'}
```
james-wills-sf, line 218: Instead of setting this text in the statusModal state object, I switched this so the state object has the more "state-y" field isInAddCommentMode, and the render function can decide what text to show.

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -162,34 +168,43 @@ const Lease = ({ form, values, store }) => {
...
   // Selected unit does not count in available a11y units.
   const remainingPriorityUnits = availableUnits.filter(
     (unit) => unit.id !== selectedUnit && unit.priority_type?.match(/Mobility|Hearing|Vision/)
   )
 
+  const handleCancelLeaseClick = (form) => {
+    leaseCanceled(dispatch, state.application)
+
+    // Reset lease and assistances on the form
+    form.change('lease', state.application.lease)
```
james-wills-sf, line 192: I couldn't figure out a better way to integrate final form with the app context. It would be nice if all of this form data lived in the context as well, but it's not really feasible with the final form version we're using now.

app/javascript/context/Reducer.js
```diff
@@ -1,97 +1,44 @@
...
-  }
-})
-
-const overrideApplicationPageData = (state, { appliedFilters, page }) => ({
-  ...state,
-  applicationsListData: {
-    ...state.applicationsListData,
-    ...(appliedFilters && { appliedFilters }),
-    ...(!(page === undefined || page === null) && { page })
+const Reducer = (state, action) => {
```
james-wills-sf, line 26: Basically what this does is:
1. When an action comes in, see what subreducers can handle the action type (can be multiple)
2. Call the applicable subreducers and set their result state to the data key associated
3. return the entire state object, with each top-level key (breadcrumbData, applicationDetailsData, etc.) referring to a subreducer's state

app/javascript/context/actionCreators/actionCreators.js
```diff
@@ -0,0 +1,32 @@
...
+  getApplicationRowClickedBreadcrumbData
+} from 'context/actionCreators/breadcrumbActionHelpers'
+import ACTIONS from 'context/actions'
+
+const getListingChangedAction = (listing) => ({
+  type: ACTIONS.SELECTED_LISTING_CHANGED,
+  data: formatListingStateData(listing)
+})
+
+export const applicationsTableFiltersApplied = (dispatch, filters) =>
```
james-wills-sf: Action creators are now just simple functions that dispatch actions

app/javascript/context/subreducers/ApplicationDetailsSubreducer.js
```diff
@@ -0,0 +1,157 @@
...
+  submitButton: null,
+  isInAddCommentMode: false
+})
+
+export const getEmptyApplicationDetailsState = () => ({
+  shortform: {
+    application: null,
+    fileBaseUrl: null
+  },
+  supplemental: {
```
james-wills-sf: Empty state should enumerate all of the fields that will get populated, this just makes it a little easier to see what the shape of the state object looks like

I think navigating reducers will be way clearer in typescript since the shape of objects (and in action payloads!) is defined in the type

app/javascript/context/subreducers/ApplicationDetailsSubreducer.js
```diff
@@ -0,0 +1,157 @@
...
+    ].filter((v) => !!v),
+    statusModal: {
+      ...state.supplemental.statusModal,
+      ...overrides.statusModal
+    }
+  }
+})
+
+const APPLICATION_DETAILS_ACTIONS = {
+  [ACTIONS.LEFT_APPLICATION_SCOPE]: (_, __) => getEmptyApplicationDetailsState(),
```
james-wills-sf: If you have thoughts on how to make reducers easier to read I'm open to suggestions!

app/javascript/utils/customHooks.js
```diff
@@ -132,3 +134,10 @@ export const useStateObject = (initialStateObject) => {
 
   return [state, setState, overrideEntireState]
 }
+
+/**
+ * Use this instead of useContext(AppContext) directly. The only reason this is better is it
+ * provides a default that will destructure nicely when useContext returns false, like when
+ * shallow rendering a component.
+ */
+export const useAppContext = () => useContext(AppContext) ?? [undefined, () => {}]
```
james-wills-sf, line 143: One thing that's annoying is any component that uses context state can't be shallow rendered. However, if it doesn't use the state and only uses the `dispatch` function like this:
```
const [, dispatch] = useAppContext()
```
you can shallow render it (but dispatch won't do anything/be spyable unless you mount render it)

app/javascript/context/subreducers/ApplicationDetailsSubreducer.js
akegan: this is a great change

app/javascript/components/supplemental_application/ApplicationDetailsPage.js
```diff
@@ -0,0 +1,169 @@
+import React, { useState } from 'react'
```
akegan: I know that the context that this is in the suppplemental_application folder should give a lot of context, but i'm wondering if we should call it LeaseUpApplicationDetailsPage to distinguish it from the pre-lottery version

app/javascript/utils/applicationDetailsUtils.js
```diff
@@ -6,3 +6,8 @@ export const totalSetAsidesForPref = (listing, prefName) =>
   listing.listing_lottery_preferences.find((pref) =>
     pref.lottery_preference.name.includes(prefName)
   )?.available_units ?? 0
+
```
akegan: should this go in the supp app directory? Might be clearer if renamed to leaseUpApplicationDetailsUntils?

app/javascript/context/subreducers/ApplicationDetailsSubreducer.js
```diff
@@ -0,0 +1,148 @@
+import LEASE_STATES from 'context/actionCreators/application_details/leaseSectionStates'
```
akegan: Another structure-related question: I'm wondering about the trade-off of having all of the reducers in one spot vs having them close to the rest of their relevant stuff. 

I think this is another point for ApplicationDetails being a bit vague too.

akegan (commented):
Still looking at the code, but here are some initial comments. Mainly around the name 'ApplicationDetails'.

QA also looks good. Will continue reviewing code tomorrow morning

james-wills-sf:
@akegan just renamed and re-organized files

what we used to call `actions` before this global state stuff I renamed to be "*RequestUtils`, all action creator related stuff goes in the feature folders under `/actions/`, and I renamed the new ApplicationDetails* files to SupplementalApplication*.

akegan:
@james-wills-sf bummer but found e2e tests found another bug so I'm not going to merge this today (and likely not tomorrow). [This onSave function](https://github.com/SFDigitalServices/sf-dahlia-lap/blob/main/app/javascript/components/supplemental_application/sections/PreferencesTable.js#L73-L74) is returning undefined instead of "true" and the preference panels aren't collapsing on successful on save. Makes me a bit worried about error handling there too. I think we'll have to do some more QA around saving edge cases like w/open preference panels, rental assistances etc to make other things are still closing as expected too. 

To fix the preferences it seems like we could:
1. Add a preference save success action
2. Close preference panels with supp app success (not sure what the timing would be if we get a supp app success then a preferences failed. 

Still seems like we might have to pull out the open/closed state from the preference table which adds some more complication.


I'm also getting intermittent `ArgumentError - A copy of Api::V1 has been removed from the module tree but is still active!:` errors which are leading to the supp app not loading. I'm not really clear on why it's happening (might be because of the route I removed? 😕 )

james-wills-sf:
@akegan I think I was able to fix the bug 🤞 waiting on e2e tests now.

About your second issue: looking through the rails logs I did see a few of those in the past, but I've had the same terminal open for weeks so I'm not sure when they were from. I wasn't able to reproduce any new errors, but let me know if you have ideas on where this is coming from.

akegan:
@james-wills-sf I just kicked off e2e tests again to see if they'll pass. Doing another review + QA now.

akegan:
@james-wills-sf I edited a preference + saved with the general save and ended up with this save/cancel buttons still showing 
image


Then tried again and the panel didn't collapse at all 😕  
image

james-wills-sf:
@akegan okay I fixed the bug you were seeing. Sorry for another big change, but ExpandableTable has always been hard for me to work with because it's event based instead of state based. That makes it hard to integrate in a react-y way. I went ahead and created a `StatefulExpandableTable` that just takes a set of rows to expand

I think this new version won't have as many bugs, and if there are bugs they should be easier to fix

app/javascript/components/supplemental_application/sections/PreferencesTable.js
```diff
@@ -1,10 +1,16 @@
 import React from 'react'
 
-import { map, reject, overSome, findIndex, orderBy, kebabCase } from 'lodash'
+import { reject, overSome, findIndex, orderBy, kebabCase } from 'lodash'
 
 import { memberNameFromPref } from 'components/applications/application_form/preferences/utils'
 import TableWrapper from 'components/atoms/TableWrapper'
 import ExpandableTable from 'components/molecules/ExpandableTable'
```
akegan: looks like this is just being imported for the ExpanderButton, which you also have in StateFullExpandableTable

akegan (dismissed):
QA looks good, but I might need to do another pass after you add review instructions to the ticket to make sure that I got everything. 

StatefulExpandableTable feels a bit copypasta-y. Are you up for following this PR w/ a PR to convert rental assistance over too? It looks like that's the only use of ExpandableTable.

james-wills-sf:
@akegan yep I didn't realize rental assistance is the onlyother usecase, I'll switch that for StatefulExpandableTable (and then rename StatefulExpandableTable back to ExpandableTable.

I can add review instructions now, I'll ping you when they're done

james-wills-sf:
@akegan review instructs updated

akegan:
Sad news, just found another bug:

1. Go to sample supp app, e.g.  https://dahlia-lap-full-pr-477.herokuapp.com/lease-ups/applications/a0o0P00000IvWt8QAF
2. Go to short form tab
3. Click on flagged record sets
4. Click on the back button

you'd expect to go back to the application, and the url changes but the page stays on flagged record sets until you get back to https://dahlia-lap-full-pr-477.herokuapp.com/lease-ups/listings

Same problem if you click away from supp app to go to pending flagged apps + hit the back button

akegan (approved):
re-approving since @james-wills-sf is going to fix the routing issue in a follow-up PR

PR #478, opened 2021-01-04
Add Application signature to leaseups [DAH-657]

software-project:
[DAH-657](https://sfgovdt.jira.com/browse/DAH-657)

Adds application signature to lease ups.


app/javascript/utils/utils.js
```diff
@@ -32,7 +32,8 @@ export const filterChanged = (prev, current) => {
   const changedFields = {}
   forEach(current, (value, key) => {
     if (!isEqual(prev[key], value)) {
-      if (isPlainObject(value) && prev[key] !== null) {
+      // date fields need to be treated as whole
+      if (isPlainObject(value) && prev[key] !== null && !('year' in value)) {
```
james-wills-sf: can you move ('year' in value) to a function like `isDateObject()` that checks if day/month/year are present and are the only fields present? I could see some other objects containing a `year` field without being a date object

app/models/force/application.rb
```diff
@@ -78,6 +78,7 @@ class Application < Force::ObjectBase
       { custom_api: 'processingStatus', domain: 'processing_status', salesforce: 'Processing_Status' },
       { custom_api: 'subStatus', domain: 'sub_status', salesforce: 'Sub_Status' },
       { custom_api: '', domain: 'status_last_updated', salesforce: 'Processing_Date_Updated' },
+      { custom_api: '', domain: 'supp_app_signed_date', salesforce: 'Supp_App_Signed_Date', type: 'date' },
```
james-wills-sf: since you're converting this to customApi in line 189, this should have custom_api: 'suppAppSignedDate', right?

spec/javascript/components/supplemental_application/sections/ConfirmedHousholdIncome.test.js
```diff
@@ -58,11 +58,11 @@ describe('ConfirmedHouseholdIncome', () => {
     const mockAmiCharts = [
       {
         ami_chart_type: 'chart1',
-        ami_chart_year: 2020
+        ami_chart_year: 2021
```
james-wills-sf, line 61: I thinnk after your rebase these changes will break this test since I added `currentYearOverride={2020}` to the getWrapper function in this file.

james-wills-sf (commented):
@software-project added a few comments, still need to QA but the code looks good

software-project:
@james-wills-sf I've addressed your issues. Regarding the one with tests, it seems to be working like that here.

james-wills-sf (approved):
@software-project code and QA lgtm!

james-wills-sf (changes requested):
@software-project 
Found a bug in QA:

Add a date, save, then delete the date, save again, and go to salesforce

The date should have been cleared out, but it's not. When you refresh the page the old date appears again

software-project:
@james-wills-sf there were actually 4 issues and 1 still remains. When we are clearing the date fields whole key is removed from application object. So I've changed changeFilter to set those values to null. Then strong parameters didn't allow nil for date hash, so thats changed. Showing validation error wasn't working (fixed). Last one is that SF endpoint is not accepting null for date. I let Larry know and he is working on that.

akegan:
Mild QA comment that the validation error should go under the fields, not sure why it's next to the fields. 
image

spec/controllers/api/v1/short_form_controller_spec.rb
```diff
@@ -171,7 +171,7 @@
           end
           expect(response).to have_http_status(:success)
           json = JSON.parse(response.body)
-          expect(json['application']['listing_id']).to eq('a0W0P00000GbyuQUAR')
+          expect(json['application']['listing_id']).to eq('a0W0P00000HbtLhUAJ')
```
akegan: wondering why this listing id would change?

spec/vcr/api/supplementals_controller/show/supplemental_application.yml
```diff
@@ -528,3164 +528,6 @@ http_interactions:
     body:
       encoding: ASCII-8BIT
       string: '{"totalSize":0,"done":true,"records":[]}'
-    http_version: null
-  recorded_at: Wed, 09 Dec 2020 19:45:56 GMT
```
akegan, line 532: I'm confused by why so many lines were deleted in this file, doesn't seem like there were so many changes to this request?

akegan (commented):
Couple of questions here for @software-project. Going to look into fixing the placement of the error message now

akegan:
I just pushed a couple of changes that fixed the issue where the validation message was showing up next to the fields. 
The reason why it came up was because there was enough space next to the fields, basically. 

To fix it I added a div around the inputs in MultiDateField and made it display:flex, then changed the error message to be display: flex + justify contents at the end. If anyone has a better idea I am up for reverting.

While trying to fix I noticed that the application signature date wasn't wrapped in formgrid row + item, which is likely why the bottom margin seemed off. I fixed that and reverted changes to ContentSection.Sub to allow for a customizable marginBottom because ideally we don't want to be customizing margins on things whenever possible, but instead using consistent patterns that fit together as expected.

Next, I'm going to look into the VCR casette + rspec changes that I had questions about to see if I can tell what's going on

Edit: Looks like there's a real e2e error too with date handling + the paper application page.
```
Expected value: "Please enter a Date of Birth"
Received array: ["Please enter a First Name", "Please enter a Last Name"]
```

spec/controllers/api/v1/short_form_controller_spec.rb
software-project: I was surprised as well, but I thought it might have been due to refresh? But on the other had we didnt had one in a long time.

akegan: ok looking into it now

akegan: I have a vague memory of finding an application that didn't belong on yellow acres + moving it back. probably was this application a0o0P00000JQawYQAT. I'll go ahead and change the application

spec/vcr/api/supplementals_controller/show/supplemental_application.yml
akegan: looking at it again, I think that it changed when we moved to making separate requests for different supp app pieces instead of including it all in the same request. And we just didn't refresh the cassette then + it happened to not break the tests 🤪 . Either in https://github.com/SFDigitalServices/sf-dahlia-lap/pull/469 or https://github.com/SFDigitalServices/sf-dahlia-lap/pull/471

akegan:
I've updated the rspec test to use an application from yellow acres + fixed the e2e test that was failing because we weren't validating dates in form arrays as expected.

@james-wills-sf this should be ready for re-review. I added some additional review instructions in the ticket to cover the wider impacts of the PR

app/javascript/utils/form/final_form/MultiDateField.js
```diff
@@ -40,53 +40,65 @@ export const MultiDateField = ({
...
   } else {
     errorField = form.getState().errors[fieldName.split('.')[0]]
   }
   const baseFieldName = last(fieldName.split('.'))
   const error =
-    form && touched && errorField && errorField[baseFieldName] && errorField[baseFieldName].all
+    form &&
+    touched &&
+    errorField &&
+    ((errorField[baseFieldName] && errorField[baseFieldName].all) || errorField.all)
```
james-wills-sf: nit: can be `errorField[baseFieldName]?.all ||`
```suggestion
    (errorField[baseFieldName]?.all || errorField.all)
```

app/javascript/utils/form/final_form/MultiDateField.js
```diff
@@ -40,53 +40,65 @@ export const MultiDateField = ({
   })
   let errorField
   if (formName) {
-    errorField = form.getState().errors[formName] ? form.getState().errors[formName][index] : null
+    if (form.getState().errors[formName]) {
```
james-wills-sf, line 43: This error stuff is super complicated, I wonder if there are react final form error handling best practices we can start to apply across the app (not saying anything needs to change now)

james-wills-sf (approved):
@akegan changes lgtm with one nit

Edit: Followed review instructions and QA looks good too

app/javascript/utils/form/final_form/MultiDateField.js
akegan: agreed!

PR #481, opened 2021-01-06
Fix react-routing/turbolinks conflict [DAH-685]

james-wills-sf:
ticket: DAH-685

Turbolinks and react-router are both accessting history state and it's causing DAH-685, where clicking the back button changes the url but doesn't actually load the page.

Removing Turbolinks fixes the issue.

## What was Turbolinks doing?
Two things, I think:
1. Making the site behave more like a single page app. Possibly making it faster by only replacing the site body and keeping the header the same.
  a. Only noticeable thing I noticed is that it makes the console log persist when clicking around which is nice but not critical
2. Triggering google analytics tracking on page load

(1) is not critical, the speed improvement was not noticeable and I think eventually the whole partners site should be migrated into one SPA (or at least multiple SPA's that are all routed via react).

(2) is more important, but there are ways around it without turbolinks. One interesting finding is that the lease-up SPA actually broke analytics tracking

### For non-react routed pages

Now that non-react routed pages are loaded like normal rails pages, we can use a regular window load listener in `application.html.slim`:
```
window.addEventListener('load', (event) =>
  ga('send', 'pageview', location.pathname + location.search);
})
```

### For react-routed pages
Note: we haven't been getting any google analytics tracking for these pages since we implemented the SPA

the new `<GoogleAnalyticsTracker />` component is a simple route that matches all URLs that will fire pageview events on url changes.

I tested to make sure these events fire properly, and that the first page load doesn't fire (to avoid double logs since the slim file will log pageview on initial page)


app/views/layouts/application.html.slim
```diff
@@ -41,3 +41,4 @@ html
 
     javascript:
       var SALESFORCE_BASE_URL = "#{{current_user&.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL']}}"
+      var SALESFORCE_BASE_URL = "#{{current_user&.admin ? ENV['SALESFORCE_INSTANCE_URL'] : ENV['COMMUNITY_LOGIN_URL']}}"
```
akegan: is this an accidental duplication of this line? or is there a difference

akegan (approved):
Looks good, just one minor question about what looks like typo.

I didn't do any QA of Google analytics because I don't think we have it set up for review apps. Let me know if you want me to double check something though

app/views/layouts/application.html.slim
james-wills-sf: This was an accident, removing it now

PR #482, opened 2021-01-07
Add Author to status history [DAH-636]

software-project:
[DAH-636](https://sfgovdt.jira.com/browse/DAH-636)


akegan (approved):
Looks good to me!

Confirmed that Sonia's name will end up on 2 lines if the screen is small enough but it still looks OK 
image

PR #483, opened 2021-01-07
Replace ExpandableTable with Stateful version [DAH-531]

james-wills-sf:
[DAH-531]

- Remove last usages of old ExpandableTable and replace them with StatefulExpandableTable
  - RentalAssistances
  - Pattern library
- Delete ExpandableTable and rename StatefulExpandableTable to ExpandableTable
- Fix a bug where clicking "Add comment" automatically started saving the form (because `button` defaults to `type='submit'`)
  - switched it to `type='button'` 

[DAH-531]: https://sfgovdt.jira.com/browse/DAH-531


app/javascript/context/subreducers/SupplementalApplicationSubreducer.js
```diff
@@ -2,6 +2,10 @@ import LEASE_STATES from 'components/supplemental_application/utils/leaseSection
 
 import ACTIONS from '../actions'
 
+// This id is never sent to the backend, only used to identify frontend
+// "new" assistance objects that haven't been persisted yet.
+export const NEW_ASSISTANCE_PSEUDO_ID = -101
```
akegan: why -101?

akegan:
We kinda need to work on our error handling. I just tested by turning off wifi + if you save just a preference or rental assistance you don't get an error message or anything saying it didn't go through.

It's an improvement over current main though since there the rows just collapse anyways w/o new info saved in them

akegan (approved):
other than what's probably a need for better error handling that's way out of scope for this PR I think things look good! 

I tested:
1. Adding new + editing rental assistance and saving individually
2. Adding new + editing rental assistance and saving with save lease
3. ""  "" and saving with save application button
4. What happens w/ all 3 if there's an error (no internet) -> see other comment on PR
5. Deleting a rental assistance
6. Opening a rental assistance + hitting cancel.

Let me know if there's anything else you think I may have missed. 

I also kicked off tests again, looks like they timed out

app/javascript/context/subreducers/SupplementalApplicationSubreducer.js
james-wills-sf: No reason. Actually this should probably be a string since the other IDs are strings, so this can be more descriptive like "new_assistance_pseudo_id"

I'll go ahead and make that change

james-wills-sf:
@akegan I think those steps make sense. Added similar ones to the review steps in JIRA

PR #489, opened 2021-01-20
Fix bulk status issues when changing pages [DAH-695][DAH-697]

james-wills-sf:
[DAH-695]
[DAH-697]

Fix two bugs:
- When you select checkboxes on page 1, then go to page 2, checkboxes from the previous page will still be "checked", You can tell by looking at the bulk checkbox indicator at the top, it will be in the "some checked" state
  - This was because we weren't completely clearing the old state on page change, we were just overriding the state with the new application values.
- When you make status changes to a row on page 1 that shares an application with a row on page 2, switching to page 2 doesn't show the updated changes until after a refresh
  - This is because we only updated the application state for the current page, when we fetch page2 records we're still working with old state
  - The fix was to force a refresh on the next page if any statuses were updated on the current page.

[DAH-695]: https://sfgovdt.jira.com/browse/DAH-695

[DAH-697]: https://sfgovdt.jira.com/browse/DAH-697

### DIFF
--- a/app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
+++ b/app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
@@ -86,6 +86,7 @@ const LeaseUpApplicationsPage = () => {
     applications: [],
     pages: 0,
     atMaxPages: false,
+    forceRefreshNextPageUpdate: false,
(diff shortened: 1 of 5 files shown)


software-project (changes requested):
Works as expected. There are just two debugging console logs that I suppose should be removed?

james-wills-sf:
@software-project forgot to push my last commit! updated now

software-project (approved):
👍

PR #491, opened 2021-01-21
[hotfix] Fix lease-up table links [DAH-698]

james-wills-sf:
[DAH-698]

I was going to add right click functionality to the whole row, but that is actually really, really difficult. It also wasn't in the older releases, so I think it's okay.

[DAH-698]: https://sfgovdt.jira.com/browse/DAH-698


akegan (approved):
looks good to me but might be a bit before I can test. Maybe Emily can test the review app instead?

PR #513, opened 2021-07-26
Fix missing AMI chart type in Salesforce DAH-824

akegan:
DAH-824

AMI Chart type was missing in Salesforce because we were choosing not to allow for a placeholder "Select One..." in the dropdown. This led to the value not saving to Salesforce. I removed the option because it seemed like it's better to have the leasing agent actively choose the value so that it's not filled out when the leasing agent is working on earlier pieces of the application. 

Review instrux in the ticket.


software-project (approved):
LGTM 👍

PR #550, opened 2022-11-02
DAH-1321 unit dropdown lease ended

christine-sfg:
-Add an additional field to pull from salesforce `Lease_Status__c`
-Update the 'Assigned Unit Number' dropdown to include units that have an application id, so long as the lease associated to it has ended

image


joni-f-g (approved):
🎉

PR #551, opened 2022-12-02
DAH-1321 unit dropdown lease ended

christine-sfg:
DAH-1321

-Rather than flattening the first lease object into the unit object, return an array of leases related to the unit. 
-On the supplemental application page, filter the available units to not include any units that have a lease in draft or signed status

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,10 +33,14 @@ yarn-debug.log*
 .env
 /public/packs
 /public/packs-test
-/node_modules
 npm-debug.log
 
 # Using yarn rather than npm
 package-lock.json
 coverage
 TAGS
+
+# IDE
+.vscode/*
+!.vscode-default/*
+.idea/*
--- a/.gitignore
+++ b/.gitignore
@@ -44,3 +44,4 @@ TAGS
 .vscode/*
 !.vscode-default/*
 .idea/*
+*.iml
--- a/app/javascript/components/supplemental_application/sections/Lease.js
+++ b/app/javascript/components/supplemental_application/sections/Lease.js
@@ -147,11 +147,18 @@ const Lease = ({ form, values }) => {
     ...map(confirmedPreferences, pluck('id', 'preference_name'))
   ])
 
+  /**
+   * available units fit the following criteria
+   *  - if it doesn't have any leases
+   *  - if there are leases, they cannot be in draft or signed status
+   *  - if the unit has an application, it must match the current one
+   */
   const availableUnits = state.units.filter(
     (unit) =>
-      !unit.application_id ||
-      unit.application_id === state.application.id ||
-      (unit.application_id && unit.lease_status === 'Ended')
+      !Object.prototype.hasOwnProperty.call(unit, 'leases') ||
+      (unit.leases &&
+        !unit.leases.some((lease) => ['draft', 'signed'].includes(lease.lease_status))) ||
+      (unit.leases && unit.leases.some((lease) => lease.application_id === state.application.id))
   )
 
   const availableUnitsOptions = formUtils.toOptions(
@@ -169,10 +176,18 @@ const Lease = ({ form, values }) => {
     return prefUsed?.preference_name === pref
   }
 
+  /**
+   * Checks if a given unit is using the same preference as the provided preference
+   * @param unit
+   * @param pref
+   * @return {false|*} returns true if the unit is used with the same preference on a different application
+   */
   const unitIsUsedWithPrefElsewhere = (unit, pref) =>
-    unit.application_id &&
-    unit.application_id !== state.application.id &&
-    unit.preference_used_name === pref
+    unit.leases &&
+    unit.leases.some(
+      (lease) =>
+        lease.application_id !== state.application.id && pref === lease.preference_used_name
+    )
 
   const numUnitsUsingPref = (pref) =>
     state.units.filter(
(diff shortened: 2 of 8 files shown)


.gitignore
```diff
@@ -33,10 +33,15 @@ yarn-debug.log*
 .env
 /public/packs
 /public/packs-test
-/node_modules
```
christine-sfg, line 36: this is a dup. already included on line 26

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -147,8 +147,17 @@ const Lease = ({ form, values }) => {
...
+  /**
+   * available units fit the following criteria
+   *  - if it doesn't have any leases
+   *  - if there are leases, they cannot be in draft or signed status
+   *  - if the unit has an application, it must match the current one
+   */
   const availableUnits = state.units.filter(
-    (unit) => !unit.application_id || unit.application_id === state.application.id
+    (unit) =>
+      !Object.prototype.hasOwnProperty.call(unit, 'leases') ||
```
shawnbot: Not a blocker, but this might be better for understanding the logic here:

```suggestion
      !Array.isArray(unit.leases) ||
```

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -147,8 +147,17 @@ const Lease = ({ form, values }) => {
...
+      !unit?.leases?.some((lease) => ['Draft', 'Signed'].includes(lease.lease_status)) ||
+      unit?.leases?.some((lease) => lease.application_id === state.application.id)
```
shawnbot: I'm pretty sure that since you're already calling `hasOwnProperty(unit, 'leases')` you haven't come across any instances of `unit` being nullish, so the `unit?.` null coalescing here is unnecessary. If you assert that `unit.leases` is an array in the above suggestion, then you can remove the `?` before `.leases` too:

```suggestion
      !unit.leases.some((lease) => ['Draft', 'Signed'].includes(lease.lease_status)) ||
      unit.leases.some((lease) => lease.application_id === state.application.id)
```

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -166,10 +175,18 @@ const Lease = ({ form, values }) => {
...
+   * Checks if a given unit is using the same preference as the provided preference
+   * @param unit
+   * @param pref
+   * @return {false|*} returns true if the unit is used with the same preference on a different application
+   */
   const unitIsUsedWithPrefElsewhere = (unit, pref) =>
-    unit.application_id &&
-    unit.application_id !== state.application.id &&
-    unit.preference_used_name === pref
+    unit.leases &&
```
shawnbot: This would be more consistent with the above suggestions, and provides lets TypeScript-enabled IDEs infer that `unit.leases` is an array so that you get code hinting :smile:

```suggestion
    Array.isArray(unit.leases) &&
```

app/models/force/unit.rb
```diff
@@ -18,6 +18,7 @@ class Unit < Force::ObjectBase
       { domain: 'max_ami_for_qualifying_unit', salesforce: 'Max_AMI_for_Qualifying_Unit' },
       { domain: 'reserved_type', salesforce: 'Reserved_Type' },
       { domain: 'priority_type', salesforce: 'Priority_Type' },
+      { domain: 'leases', salesforce: 'Leases'}
```
shawnbot: Missing a space here :neckbeard: 

```suggestion
      { domain: 'leases', salesforce: 'Leases' }
```

app/services/force/units_service.rb
```diff
@@ -23,12 +22,16 @@ def units_and_leases_for_listing(listing_id)
...
-        # Flatten lease fields and merge with unit fields
-        lease = unit['Leases'] ? unit['Leases'][0] : nil
-        domain_lease_fields = Force::Lease.from_salesforce(lease).to_domain if lease
         domain_unit = Force::Unit.from_salesforce(unit).to_domain
-        domain_unit.merge(domain_lease_fields || {})
+        if domain_unit['leases']
+          domain_leases = domain_unit['leases'].map do |lease|
+            domain_lease = Force::Lease.from_salesforce(lease).to_domain if lease
+          end
+          domain_unit['leases'] = domain_leases
```
shawnbot: I'm confused by this. I _think_ what's happening (because it appears to work) is that the `.map` lambda is returning the assignment of a local variable `domain_lease`. If that's the case, the `domain_lease =` bit can go away, and I think it'd be clearer to have an explicit `return`.

I'm also curious if there actually are any falsy `lease` values that come through that function. If there are, then you're possibly setting `domain_unit['leases']` to an array containing `nil` values, which the front end would need to deal with. One way to avoid a sparse array would be to filter `domain_unit['leases']` before mapping. I _think_ you can rewrite the whole thing (without assigning to `domain_leases` and reassigning) to:

```suggestion
          domain_unit['leases'] = domain_unit['leases']
            .filter do |lease| true if lease
            .map do |lease| Force::Lease.from_salesforce(lease).to_domain
```

LMK if you want to pair on it to make sure it's doing the right thing 😁

christine-sfg: Thanks for all the great feedback! This makes sense.

christine-sfg:
Just realized I was working on an old branch. Creating a new one.

PR #552, opened 2022-12-03
DAH-1321 cleanup filtering for units and leases

christine-sfg:
[DAH-1321](https://sfgovdt.jira.com/browse/DAH-1321)

-Rather than flattening the first lease object into the unit object, return an array of leases related to the unit.
-On the supplemental application page, filter the available units to not include any units that have a lease in draft or signed status

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.7.1'
+ruby '2.7.0'
 
 git_source(:github) do |repo_name|
   repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.7.0'
+ruby '2.7.1'
 
 git_source(:github) do |repo_name|
   repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.7.7'
+ruby '2.7.0'
 
 git_source(:github) do |repo_name|
   repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
--- a/Gemfile
+++ b/Gemfile
@@ -1,5 +1,5 @@
 source 'https://rubygems.org'
-ruby '2.7.0'
+ruby '2.7.7'
 
 git_source(:github) do |repo_name|
   repo_name = "#{repo_name}/#{repo_name}" unless repo_name.include?("/")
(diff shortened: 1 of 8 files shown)


app/services/force/units_service.rb
```diff
@@ -23,12 +22,15 @@ def units_and_leases_for_listing(listing_id)
...
       domain_units = result.map do |unit|
-        # Flatten lease fields and merge with unit fields
-        lease = unit['Leases'] ? unit['Leases'][0] : nil
-        domain_lease_fields = Force::Lease.from_salesforce(lease).to_domain if lease
         domain_unit = Force::Unit.from_salesforce(unit).to_domain
-        domain_unit.merge(domain_lease_fields || {})
+        if domain_unit['leases']
+          domain_unit['leases'] = domain_unit['leases']
+                                    .filter do |lease| true if lease end
+                                    .map do |lease| Force::Lease.from_salesforce(lease).to_domain end
```
shawnbot: Ack sorry, I forgot the `end` in my suggestion! 🤦 

I think you can also use block syntax here, which is more similar to JS arrow function:

```suggestion
                                    .filter { |lease| true if lease }
                                    .map { |lease| Force::Lease.from_salesforce(lease).to_domain }
```

app/javascript/components/supplemental_application/sections/Lease.js
```diff
@@ -147,11 +147,17 @@ const Lease = ({ form, values }) => {
...
+   */
   const availableUnits = state.units.filter(
     (unit) =>
-      !unit.application_id ||
-      unit.application_id === state.application.id ||
-      (unit.application_id && unit.lease_status === 'Ended')
+      !Array.isArray(unit.leases) ||
+      !unit.leases.some((lease) => ['Draft', 'Signed'].includes(lease.lease_status)) ||
+      unit.leases.some((lease) => lease.application_id === state.application.id)
   )
```
shawnbot: I'm pretty sure that most JS interpreters will optimize this properly, but if you're filtering hundreds of units if it might be good to separate out the function that checks the lease status, which is currently creating a new array every time it executes:

```js
const draftOrSigned = ['Draft', 'Signed']

// ...
!unit.leases.some(lease => draftOrSigned.includes(lease.lease_status)) ||
```

shawnbot (approved):
This looks great! Just a couple of non-blocking nits, which you can file away for later (FIXME comments referencing these suggestions?) if you need to get this out.

Gemfile
```diff
@@ -33,6 +33,7 @@ gem 'jbuilder', '~> 2.5'
 # gem 'capistrano-rails', group: :development
 
 gem 'newrelic_rpm'
+gem 'mini_portile2', '~> 2.5', '>= 2.5.1'
```
christine-sfg, line 36: heroku build was suddenly failing, erroring out from missing this gem. as a workaround we are explicitly adding it

PR #598, opened 2024-02-01
feat: DAH-2008 getting and updating veteran fields from salesforce

alulabeshue-sfgov:
[DAH-2008](https://sfgovdt.jira.com/browse/DAH-2008)

[DAH-2008]: https://sfgovdt.jira.com/browse/DAH-2008?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/services/force/soql/preference_service.rb
```diff
@@ -7,11 +7,12 @@ class PreferenceService < Force::Base
       FIELD_NAME = :preferences
       FIELDS = load_fields(FIELD_NAME).freeze
 
-      def app_preferences_for_application(application_id)
+      def app_preferences_for_application(application_id, should_order: false)
         result = parsed_index_query(%(
           SELECT #{query_fields(:app_preferences_for_application)}
           FROM Application_Preference__c
           WHERE Application__c = '#{application_id}'
+          #{should_order ? 'Order by Preference_Order__c asc' : ''}
```
chadbrokaw: Should `Order by` be capitalized to `ORDER BY`? Not sure what the convention is exactly

chadbrokaw (approved):

app/services/force/soql/preference_service.rb
alulabeshue-sfgov: SOQL states that their clauses are not case sensitive. But the rest of the code is using capital clauses so I will follow the same pattern.

PR #603, opened 2024-02-09
chore: DAH-1945 Upgrade to Rails 7 part 1

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-1945


cliu02 (approved):
I did not encounter any issues when clicking through the Partners Sign In side.

PR #611, opened 2024-02-27
DAH-2135 remove archive call

tallulahkay:
DAH-2135

You'll see that I updated the vcr test manually - this is because I know a full update is coming soon so I'm not worried about it: https://sfgovdt.jira.com/browse/DAH-2121


alulabeshue-sfgov (approved):
LG!

PR #616, opened 2024-03-05
chore: DAH-2153 adding logs when application fetch fails in lease up page

alulabeshue-sfgov:
[DAH-2153](https://sfgovdt.jira.com/browse/DAH-2153)

[DAH-2153]: https://sfgovdt.jira.com/browse/DAH-2153?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


tallulahkay:
Should we attach a ticket to this?

app/controllers/api/v1/short_form_controller.rb
```diff
@@ -126,10 +126,6 @@ def application_db_params
           )
   end
 
-  def custom_api_application_service
```
tallulahkay, line 129: oh lol how did this get duplicated

tallulahkay:
@alulabeshue-sfgov Could you demo or explain how I mark as the preference as received? It says that field is auto populated, so I cannot edit it

alulabeshue-sfgov:
@tallulahkay 

> Should we attach a ticket to this?

I was thinking to create a small ticket but I then thought to not because it was quick to just make the change lol. But might be better for tracking with our releases. I can create a small ticket with a description and acceptance criteria

> oh lol how did this get duplicated

Im not sure why lol. But it looks like the same function was added in this [PR](https://github.com/SFDigitalServices/sf-dahlia-lap/pull/418) when that function already existed.

> @alulabeshue-sfgov Could you demo or explain how I mark as the preference as received? It says that field is auto populated, so I cannot edit it

Yea I've been using the `MOHCD Preference Assertion`, when this goes to `Valid for lottery` it also makes the preference `Receives Preference`

tallulahkay (approved):
@alulabeshue-sfgov This works, I think it will be helpful!!

PR #24, opened 2018-05-16
Chores/refactor controller #157578818

fddayan:

PR #617, opened 2024-03-07
feat: DAH-2133 Create Rails API endpoints for flagged applications

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2133


cliu02:
From the review instructions, it looks like someone with MOHCD admin privileges is needed to review (the automatic assigned reviewers don't have this login).

app/controllers/applications/flagged_controller.rb
```diff
@@ -1,46 +1,11 @@
 # frozen_string_literal: true
```
tallulahkay, line 1: I wonder if we should just delete this file?

tallulahkay (approved):

tallulahkay:
It looks good to me and functionally works well!

app/controllers/applications/flagged_controller.rb
jimlin-sfgov: We still need this file for routing purposes unfortunately. To delete this file, we would have to update the `routes.rb` file, some react routing code, and tests.

app/controllers/api/v1/flagged_applications_controller.rb
```diff
@@ -21,7 +47,27 @@ def flagged_application_params
...
+  def pending_data
+    flagged_record_set_get_service.pending_review_record_sets
+    # fields = flagged_record_set_get_service.pending_review_index_fields
+    # [pending_review_record_sets, fields]
+  end
+
+  def duplicated_data
+    flagged_record_set_get_service.marked_duplicate_record_sets
+    # fields = flagged_record_set_get_service.marked_duplicate_index_fields
+    # [marked_duplicate_record_sets, fields]
```
alulabeshue-sfgov: **question:** Do we need these commented out lines? You are already returning the record set and it doesnt look like fields were ever used in FlaggedApplicationsIndexPage or FlaggedApplicationsShowPage

config/routes.rb
```diff
@@ -49,8 +49,9 @@
         resources :field_update_comments, only: %w[index create]
       end
 
-      scope '/flagged-applications' do
-        put 'update' => 'flagged_applications#update'
+      resources :flagged_applications, path: 'flagged-applications', only: %w[index update]
+      scope 'flagged-applications' do
+        get 'record-set/:id' => 'flagged_applications#record_set'
```
alulabeshue-sfgov, line 54: **note:** we already have talks to move and redesign our APIs. For the url `/api/v1/flagged-applications?type=pending` I would think it returns a list of applications that have been flagged but instead it is returning a list of listings that have flagged applications. Same happens with the web page url `/applications/flagged?type=pending`. Maybe something like `/listings/applications?type=flagged` might

cypress/e2e/FlaggedApplicationsShowPage.e2e.js
```diff
@@ -5,7 +5,17 @@ const commentInputSelector = 'input[type="text"]'
...
+// describe('FlaggedApplicationsIndexPage', () => {
+//   // once we have fixtures, we can intercept these API endpoints:
+//   // cy.intercept("api/v1/flagged-applications?type=pending", { fixture: "foobar" }).as("pendingFlaggedRecordsWithTitle")
+//   // cy.intercept("api/v1/flagged-applications?type=duplicated)", { fixture: "foobar" }).as("duplicatedFlaggedRecordsWithTitle")
+// })
+
 describe('FlaggedApplicationsShowPage', () => {
+  // once we have fixtures, we can intercept this API endpoint:
+  // cy.intercept("api/v1/flagged-applications/record-set/*", { fixture: "foobar" }).as("flaggedRecordSet")
+
```
alulabeshue-sfgov: **question:** are we leaving these in here for the next ticket? do we want to commit these or pull them into the next commit if so?

app/controllers/api/v1/flagged_applications_controller.rb
jimlin-sfgov: These commented out lines were in use (but not used by React) before I moved data fetching to the Rails API. These comments can be removed, since we can reference the blame if we ever finding ourselves needing `fields` again.

cypress/e2e/FlaggedApplicationsShowPage.e2e.js
jimlin-sfgov: Yes, this is for the next ticket. I wrote them as a convenience for who ever picks up that ticket, but I can remove them if it is trivial for the next ticket's assignee to recreate it.

config/routes.rb
jimlin-sfgov: Yes, I agree that the naming conventions in the code is inconsistent with the data being processed, and it was a big source of confusion for me as a relative newcomer to the LAP repository. [This](https://github.com/SFDigitalServices/sf-dahlia-lap/pull/24#issuecomment-389608615) is when the naming got muddled for the sake of "clean code."

Under the hood, we are using SOQL queries to load data from the `Flagged_Record_Set__c` Salesforce table, so it would make more sense to organize the endpoints like:
```ruby
# used to update status and comment of a single flagged application
resources :flagged_applications, path: 'flagged-applications', only: %w[update]
# used to get pending flagged record sets, marked-duplicate flagged record sets, and individual flagged record sets
resources :flagged_application_record_sets, path: 'flagged-application-record-sets', only: %w[index show]
```

This isn't in the scope of the ticket, but happy to create a new ticket to capture this.

alulabeshue-sfgov: I think its fine for now since we will be creating ticket in the future for new APIs, just wanted to note. Thanks!

cypress/e2e/FlaggedApplicationsShowPage.e2e.js
alulabeshue-sfgov: I think this is fine as well, more of a nitpick. But I'l leave it up to you.

alulabeshue-sfgov (approved):
LGTM! Thanks Jim!

PR #620, opened 2024-03-12
fix: DAH-2159: Apply clear filter 

chadbrokaw:
[DAH-2159](https://sfgovdt.jira.com/browse/DAH-2159)

[DAH-2159]: https://sfgovdt.jira.com/browse/DAH-2159?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/javascript/components/lease_ups/LeaseUpApplicationsFilterContainer.js
+++ b/app/javascript/components/lease_ups/LeaseUpApplicationsFilterContainer.js
@@ -132,7 +132,10 @@ const LeaseUpApplicationsFilterContainer = ({
               <div className='filter-group'>
                 <div className='filter-group_item__large'>
                   <SearchField
-                    onClearClick={() => form.change('search', '')}
+                    onClearClick={() => {
+                      form.change('search', '')
+                      form.submit()
+                    }}
                     fieldName='search'
                     id='test-search'
                     placeholder='Application, First Name, Last Name...'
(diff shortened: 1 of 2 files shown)


alulabeshue-sfgov:
@chadbrokaw For the filters this change looks good, I now see the `Clear Filters` button working. In the ticket it mentions search and having to hit the `Search` button again. Did I read that correctly, will you be covering that scenario?

chadbrokaw:
> @chadbrokaw For the filters this change looks good, I now see the `Clear Filters` button working. In the ticket it mentions search and having to hit the `Search` button again. Did I read that correctly, will you be covering that scenario?

Thank you for flagging! I updated the PR to cover that scenario

alulabeshue-sfgov (approved):
In the ticket there is the below text in the description which makes me think if the expectation is that clearing the field alone will also clear the search, which it does not with this change (you need to click the `x` button or clear the field and hit enter again). Might be a good idea to follow up with John, who reported the ticket, if this is the expected behavior.

> To get back to seeing all applicants, you have to clear the search and then click Search again, which is not very intuitive.  Seems like clearing the search field should be enough to return you to the full list.

Other than that, LGTM!

PR #622, opened 2024-03-20
feat: DAH-2116 veterans nice to haves

alulabeshue-sfgov:
[DAH-2116](https://sfgovdt.jira.com/browse/DAH-2116)

[DAH-2116]: https://sfgovdt.jira.com/browse/DAH-2116?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


app/javascript/components/supplemental_application/actions/preferenceActionCreators.js
```diff
@@ -5,10 +5,15 @@ import ACTIONS from 'context/actions'
  * Returns a promise that resolves to true if the save was successful, false otherwise.
  */
 
-export const updateSavedPreference = async (dispatch, preferenceIndex, formApplicationValues) => {
+export const updateSavedPreference = async (
+  dispatch,
+  preferenceIndexToUpdate,
+  preferenceIndexToClose,
+  formApplicationValues
+) => {
```
jimlin-sfgov, line 13: question: I'm curious about what this change addresses, maybe a bug? I'm still not familiar with how state management works in LAP.

jimlin-sfgov (approved):
amazing work!

app/javascript/components/supplemental_application/actions/preferenceActionCreators.js
alulabeshue-sfgov: It stems from the requirement in this ticket where they no longer want veteran preference to be editable but the non veteran preference should remain editable from the veteran panel. The preference that i need to save is not always the current index, but the preference i need to close will be the current index.

https://github.com/SFDigitalServices/sf-dahlia-lap/blob/f300f8e74a116ef8474a62150498046983c9e9b0/app/javascript/components/supplemental_application/sections/preferences/Panel.js#L74-L82

Screenshot 2024-03-26 at 12 06 31 PM

PR #623, opened 2024-03-21
chore: DAH-1904 clean up preference type query

tallulahkay:
DAH-1904


app/javascript/utils/layeredPreferenceUtil.js
```diff
@@ -20,7 +20,6 @@ export const addLayeredValidation = (preferences) => {
     const filteredPrefs = preferences.filter(
       (data) =>
         data.application_id === preference.application_id &&
-        data.preference_record_type === preference.preference_record_type &&
```
tallulahkay, line 23: @alulabeshue-sfgov do we want to replace this with some sort of check? It's been undefined === undefined recently, but maybe we should check the equality of custom_preference_type after removing the string "V-" from the field?

alulabeshue-sfgov: @tallulahkay good catch! Locally I am seeing that the preference_record_type is still available and being used in this logic, are you seeing that they are undefined? Going forward you are right, we will still need a way to compare the non veterans to veterans to match counterparts. Also we are starting to look into having educator preferences have a similar behavior which will further complicate this logic. Im thinking we should huddle with salesforce team on what is the best way to calculate the final confirmation and where should it live.

PR #628, opened 2024-04-03
fix: DAH-2116 using Custom_Preference_Type instead of Preference_Name

alulabeshue-sfgov:
# Issue
Check this [comment](https://sfgovdt.jira.com/browse/DAH-2116?focusedCommentId=78700) on DAH-2116

# Changes
1. calls to `/api/v1/supplementals` returns `Custom_Preference_Type`
2. `custom_preference_type` is used instead of `preference_name` when checking if a row should display the edit button

# Review Instructions
1. Go to a lease up application like `/lease-ups/applications/a0o8H000002DaaMQAS`
2. In the Preferences Table COP and DTHP rows (including veterans) should not display an edit button
3. Preferences like LW and NRHP should display an edit button


app/javascript/components/supplemental_application/sections/PreferencesTable.js
```diff
@@ -121,7 +125,14 @@ const PreferencesTable = ({
...
 
-  const rows = buildRows(application, applicationMembers, fileBaseUrl)
+  const sortedAndFilteredPreferences = sortAndFilter(application.preferences)
+
+  const rows = buildRows(
+    sortedAndFilteredPreferences,
+    application.proof_files,
+    applicationMembers,
+    fileBaseUrl
+  )
```
jimlin-sfgov, line 135: question: For all of the changes made above this line, was the purpose just to be able to pass the `originals={sortedAndFilteredPreferences}` prop to the `sortedAndFilteredPreferences` component?

jimlin-sfgov (approved):
looks good!

app/javascript/components/supplemental_application/sections/PreferencesTable.js
alulabeshue-sfgov: Yes. By the time we call `ExpandableTable` the rows passed in only contain a subset of the original data, not including the preference type. I noticed that `ExpandableTable` accepts an `originals` prop which can later be called.

PR #631, opened 2024-04-30
DAH 2293 support first come first serve listings

tallulahkay:
DAH-2293


app/services/force/soql/lease_up_application_service.rb
```diff
@@ -0,0 +1,88 @@
...
+          search_fields.each do |search_field|
+            search_term_clause.push("#{search_field} like '%#{sanitized_term}%'")
+          end
+          search.push("(#{search_term_clause.join(' OR ')})")
+        end
+        # If there are multiple terms, join the clauses with "AND" because we expect each term to find a match.
+        search.join(' AND ')
+      end
+
+      def build_applications_filters(opts)
```
jimlin-sfgov, line 41: suggestion (non-blocking): If you want to fix the complexity warning you can extract the logic into three methods, one for each option; the method would then look something like:

```ruby
      def build_applications_filters(opts)
        "#{accessibility_filter(opts)}#{status_filter(opts)}#{household_filter(opts)}"
      end
```

app/services/force/soql/lease_up_application_service.rb
```diff
@@ -0,0 +1,88 @@
...
+        # This builds the syntax need for soql to do an
+        # OR statement when passing the option of `Vision/Hearing`
+        # https://developer.salesforce.com/docs/atlas.en-us.soql_sosl.meta/soql_sosl/sforce_api_calls_soql_querying_multiselect_picklists.htm
+        filters = ''
+        if opts[:accessibility].present?
+          accessibility_string = opts[:accessibility].map do |accessibility|
+            accessibility.split(', ').map { |item| "'#{item}'" }
+          end.join(', ')
+          filters += " and Has_ADA_Priorities_Selected__c INCLUDES (#{accessibility_string}) "
```
jimlin-sfgov, line 50: question: how did you figure out that you didn't need to prepend `Application__r.` to this and other similar queries?

app/javascript/components/lease_ups/utils/leaseUpRequestUtils.js
```diff
@@ -25,11 +27,16 @@ export const getApplications = async (listingId, page, filters) => {
...
-      const preferences = map(records, buildLeaseUpAppPrefModel)
-      const layeredPreferences = addLayeredValidation(preferences)
+    .then(({ records, pages, listing_type }) => {
+      let apps
+      if (listing_type === LISTING_TYPE_FIRST_COME_FIRST_SERVED) {
+        apps = map(records, buildLeaseUpAppFirstComeFirstServedModel)
+      } else {
+        const preferences = map(records, buildLeaseUpAppPrefModel)
+        apps = addLayeredValidation(preferences)
+      }
```
jimlin-sfgov, line 37: suggestion: this change in logic warrants an update to the tests in `spec/javascript/components/lease_ups/utils/leaseUpRequestUtils.test.js`

tallulahkay: Good q! In preference_service, we are querying preferences, so when a user searches, their searches and filters are supposed to apply to the application, but in the backend, we still do the query on preferences, so we need index into the associated application and then filter/search on those

In lease_up_application_service, we are now querying applications directly (because the fcfs listings don't have preferences), so we don't have to index into the application object from the preference anymore - we are already starting with application

Old query:
Grab all preference objects, index into associated application, apply searches and filters

New query:
Grab all applications, apply searches and filters

app/services/force/soql/lease_up_application_service.rb
tallulahkay: Good q! In preference_service, we are querying preferences, so when a user searches, their searches and filters are supposed to apply to the application, but in the backend, we still do the query on preferences, so we need index into the associated application and then filter/search on those

In lease_up_application_service, we are now querying applications directly (because the fcfs listings don't have preferences), so we don't have to index into the application object from the preference anymore - we are already starting with application

Old query:
Grab all preference objects, index into associated application, apply searches and filters

New query:
Grab all applications, apply searches and filters

app/services/force/soql/lease_up_application_service.rb
tallulahkay: That's amazing and sooo satisfying thank you

app/javascript/components/lease_ups/utils/leaseUpRequestUtils.js
tallulahkay: omg great call because in writing this test I realized some data missing from the new `leaseUpAppFirstComeFirstServedModel.js` - yay for test driven development!

app/services/force/soql/lease_up_application_service.rb
jimlin-sfgov: Thank you! I must have glossed over the main table used in `builder.from(...)`, which would have clued me in. I appreciate your work with our homegrown soql orm 😅

spec/javascript/components/lease_ups/utils/leaseUpRequestUtils-fcfs.test.js
```diff
@@ -0,0 +1,64 @@
...
+        status_last_updated: '2020-05-28T21:23:06.000+0000',
+        total_household_size: 1,
+        sub_status: 'Approval letter sent'
+      }
+
+      const expectedResults = { records: [expectedRowData], pages: 10 }
+      expect(await getApplications(fakeListingId, 0)).toEqual(expectedResults)
+    })
+  })
+})
```
jimlin-sfgov: question: There's already a test file for `leaseUpRequestUtils`, why create a new one? I could be wrong, but our code coverage tool matches files to test files based on filename.

tallulahkay: I found it was the easiest way that I could return a different response from the mock api - let me try to see if I can do it in the same file

jimlin-sfgov: Thanks, my main concern is that reported code coverage maybe lower than it actually is, if our coverage tooling does not detect this test file.

tallulahkay: @jimlin-sfgov updated the test so that it's all in one file

jimlin-sfgov (approved):
great job!

PR #633, opened 2024-05-06
DAH 2142 salesforce cleanup job

tallulahkay:
DAH-2142

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -204,7 +204,6 @@ workflows:
           requires:
             - build
       - waiter: 
-          <<: *non_production_jobs
           requires:
             - build
       - cleanup-test-data:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -183,3 +203,10 @@ workflows:
           name: e2e-production
           requires:
             - build
+      - waiter:
```
tallulahkay: We need to make sure it also runs on prod

alulabeshue-sfgov (commented):
question: In this partners PR you are running the cleanup job without having to run the rails server, just calling a rake job. In [webapp](https://github.com/SFDigitalServices/sf-dahlia-web/blob/main/.circleci/config.yml#L249) we are running the rails server before calling the rake job. Do we need the additional steps in webapp to start the rails server? The start server step in circleci adds some time to the build so was wondering if we need it.

tallulahkay:
@alulabeshue-sfgov you're so right - don't think we need the start-server part in webapp - I'll make a ticket to confirm and remove

alulabeshue-sfgov (approved):
LGTM, thanks!

tallulahkay:
I converted to draft bc reggie is changing the endpoint to a DELETE

PR #639, opened 2024-05-30
fix/DAH-2351 update application new page test

tallulahkay:
DAH-2351


chadbrokaw (approved):
Test are passing in CircleCI

PR #647, opened 2024-07-18
DAH-2334 make subtitles editable

tallulahkay:
DAH-2334

There was a request to make subtitles editable.


jimlin-sfgov (approved):
User interface could be better, but designers aren't involved in this feature so that's ok.
Screenshot 2024-07-24 at 2 56 14 PM

PR #648, opened 2024-07-22
Fix/DAH-2475 fix layered pref

tallulahkay:
DAH-2475


app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -3,24 +3,57 @@
...
+        end
+      end
+
+      applications_response = soql_preference_service.app_preferences_for_listing(params, general)
+      applications = [*prev_applications, *applications_response[:records]]
+      acc_size += applications_response.total_size
+
+      return { applications: applications, acc_size: acc_size } unless applications_response.total_size > OFFSET_LIMIT
+
+      get_application_preferences(general, applications, acc_size)
```
jimlin-sfgov: suggestion (non-blocking): Try to avoid recursive method calls, ruby has other ways to loop through logic like blocks and lambdas.

app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -3,24 +3,57 @@
...
+        if general
+          params[:general_lottery_rank] = prev_applications[-1]['Application']['General_Lottery_Rank']
+        else
+          params[:preference_order] = prev_applications[-1]['Preference_Order']
+          params[:preference_lottery_rank] = prev_applications[-1]['Preference_Lottery_Rank']
+        end
+      end
+
+      applications_response = soql_preference_service.app_preferences_for_listing(params, general)
+      applications = [*prev_applications, *applications_response[:records]]
```
jimlin-sfgov: note: `array1 + array2` achieves the same thing in ruby.

app/services/force/soql/preference_service.rb
```diff
@@ -55,59 +55,78 @@ def buildAppPreferencesFilters(opts)
...
-      ]
-      search = []
-      for search_term in search_terms
-        # For each term we join the fields with OR to return a record if it matches
-        # across any of these fields.
-        search_term_clause = []
-        sanitized_term = ActiveRecord::Base.sanitize_sql_like search_term
-        for search_field in search_fields
-          search_term_clause.push("#{search_field} like '%#{sanitized_term}%'")
+      def buildAppPreferencesSearch(search_terms_string)
```
jimlin-sfgov: question: are you not getting Rubocop linting errors?
Screenshot 2024-07-22 at 5 31 32 PM

app/services/force/soql/preference_service.rb
```diff
@@ -55,59 +55,78 @@ def buildAppPreferencesFilters(opts)
...
+        # For each term, we search across name and application number for a match
+        # For multiple terms, we expect the record to match each of the terms individually.
+        search_terms = search_terms_string.split(',')
+        search_fields = [
+          'Application__r.Name',
+          'Application__r.Applicant__r.First_Name__c',
+          'Application__r.Applicant__r.Last_Name__c',
+        ]
+        search = []
+        for search_term in search_terms
```
jimlin-sfgov: issue: follow Rubocop guidelines and use block-style looping, e.g. `search_terms.each do ... end`.

jimlin-sfgov (commented):
Didn't test yet but I noticed a lot of linting issues. We should add Rubocop to our codeclimate config so we can catch these before the review stage of PRs.

app/services/force/soql/preference_service.rb
tallulahkay: I am, but am not addressing the errors that I didn't create

app/services/force/soql/preference_service.rb
tallulahkay: I'm only addressing rubocop things that I wrote - would rather not update old code if I don't have to

app/controllers/api/v1/lease_up_applications_controller.rb
tallulahkay: I think this is a perfect use case for recursion because it's necessary here that the next call depends on the response of the previous call - let's discuss?

tallulahkay:
@jimlin-sfgov I'm re-requesting your review because I'm not going to edit code that I didn't touch

it looks like my ide auto-formatted some indenting issues the buildAppPreferencesFilters chunk of code had, but I didn't actually change anything. I've reverted the indenting changes so that it doesn't show up in the code diff

Sorry that it was confusing to see what changes I actually made - happy to discuss tomorrow after you review again!

jimlin-sfgov:
> it looks like my ide auto-formatted some indenting issues

@tallulahkay Thanks for reverting to keep the diff clean. I'm confused why overcommit didn't catch that though, maybe it doesn't consider whitespace when looking for diffs?

app/controllers/api/v1/lease_up_applications_controller.rb
jimlin-sfgov: Happy to talk about it, here's an alternative implementation without recursion:
```ruby
def foo(general: false)
  params = lease_up_apps_params
  total_size = 0
  applications = []
  failsafe_count = 1_000

  failsafe_count.times do
    if applications.present? && general
      params[:general_lottery_rank] = applications[-1]['Application']['General_Lottery_Rank']
    elsif applications.present?
      params[:preference_order] = applications[-1]['Preference_Order']
      params[:preference_lottery_rank] = applications[-1]['Preference_Lottery_Rank']
    end
    applications_response = soql_preference_service.app_preferences_for_listing(params, general)
    applications.concat(applications_response[:records])
    return [applications, total_size] unless applications_response.total_size > OFFSET_LIMIT
  end

  raise 'escaped infinite loop'
end
```

app/controllers/api/v1/lease_up_applications_controller.rb
```diff
@@ -3,24 +3,57 @@
...
+
+        applications[:listing_type] = listing_type
+        render json: applications
       else
         # All other listings need to be queried by preferences first
-        applications = soql_preference_service.app_preferences_for_listing(lease_up_apps_params)
-        applications[:records] = Force::Preference.convert_list(applications[:records], :from_salesforce, :to_domain)
+        application_preferences = get_application_preferences
+        general_applications = get_application_preferences(true)
+        combined = [*application_preferences[:applications], *general_applications[:applications]]
```
jimlin-sfgov: question: As someone not familiar with this controller, these variable names look like we're mixing data types. Are we combining applications and preferences into a single array?

jimlin-sfgov:
note: For listings with a lot applications the initial load time is approaching the [Heroku timeout threshold](https://devcenter.heroku.com/articles/request-timeout) of 30 seconds. E.g. for Islais place, the load time for this PR is around 24 seconds, compared to 7 seconds in Prod.

app/controllers/api/v1/lease_up_applications_controller.rb
tallulahkay: ah wonder if it's because of the naming - I see how it's inconsistent. We are really dealing with application preferences exclusively and appending application data to each application preference, but the list is based on application preferences. In the general case, it's basically a fake application preference for each general application. Does that make sense? Is it worth me going in and updating the variable naming?

tallulahkay:
> note: For listings with a lot applications the initial load time is approaching the [Heroku timeout threshold](https://devcenter.heroku.com/articles/request-timeout) of 30 seconds. E.g. for Islais place, the load time for this PR is around 24 seconds, compared to 7 seconds in Prod.

Yeah it's totally unideal - I hope having that many preferences is anomalous, and that we can hear from leasing agents if anything in partners prod crashes. Once Reggie's api is released to prod I can fix the timing!! Which should be soon

app/controllers/api/v1/lease_up_applications_controller.rb
tallulahkay: this led me down a cool rabbit hole!! Found this: https://nithinbekal.com/posts/ruby-tco/

Didn't know about tail recursion, that they are the same as loops, impact on stack size, etc

tallulahkay:
@jimlin-sfgov ugh although maybe I jumped the gun and should implement the api rn

tallulahkay:
@jimlin-sfgov I went back to the drawing board and moved the loop to the frontend (I also used your while loop suggestion!) so that we are making multiple api calls to our rails backend rather than just multiple calls to salesforce wrapped by one api call - doesn't improve the user experience but at least avoids the risk of a heroku-enforced timeout

app/controllers/api/v1/lease_up_applications_controller.rb
jimlin-sfgov: Without explicit typing, I would say good variable names are very important in ruby code. Something to keep in mind going forward.

jimlin-sfgov (approved):
lgtm!

PR #651, opened 2024-07-31
DAH 2498 fcfs use sf ranks

tallulahkay:
Put as a branch so it can be reviewed, but will release as a hotfix


alulabeshue-sfgov (approved):
LG!

PR #653, opened 2024-08-02
DAH-2461 regard layered validations

tallulahkay:
DAH-2461


cliu02 (approved):
Looks good, validated preferences match the lease details dropdown!

PR #654, opened 2024-08-02
DAH-2475 fix fcfs call

tallulahkay:
DAH-2475

I noticed a bug in full after the original PR for this ticket got merged. FCFS listings don't have general applications, so it was just returning duplicate applications.

See https://dahlia-lap-full.herokuapp.com/lease-ups/listings/a0W6s000008C9TlEAK - it returns all the applicants twice
Compare to the pr app on the same lease up listing and it should only return all the applicants once.


cliu02 (approved):
Looks good locally, duplicates have been removed so applicants only show once!

PR #656, opened 2024-08-08
DAH 2488 Decouple Validations from List Retrieval

tallulahkay:
DAH-2488

This change calls lease up listing applications separately for processing layered validations so that they exist on filters. I implemented it a little differently than described in the ticket. Open to changing the implementation upon suggestions! I tried to limit when a second call is made to keep response times relatively less slow. It tows the line between trying to minimize changes and solve the problem...

### DIFF
--- a/app/javascript/apiService.js
+++ b/app/javascript/apiService.js
@@ -77,7 +77,7 @@ const fetchApplications = async ({ page, filters }) =>
     true
   )
 
-const fetchLeaseUpApplications = async (listingId, page, { filters }) => {
+const fetchLeaseUpApplications = async (listingId, page, { filters }, withGeneral = true) => {
   const generalApps = {
     records: [],
     pages: 0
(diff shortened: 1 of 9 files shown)


app/javascript/apiService.js
```diff
@@ -77,7 +77,7 @@ const fetchApplications = async ({ page, filters }) =>
     true
   )
 
-const fetchLeaseUpApplications = async (listingId, page, { filters }) => {
+const fetchLeaseUpApplications = async (listingId, page, { filters }, withGeneral = true) => {
```
fwextensions: Not that I need to understand it, but `withGeneral` isn't immediately clear.  Maybe `includeGeneralApps`?  But I'm not sure what a "general application" is.

tallulahkay: Good call! It's the applications without preferences - they don't need to be called for processing layered validations. I'll update the parameter name and add a code comment

jimlin-sfgov (approved):
LGTM

PR #659, opened 2024-08-22
feat: DAH-2549 add hcbs units to partners

tallulahkay:
DAH-2549

### DIFF
--- a/app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
+++ b/app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
@@ -29,7 +29,7 @@ const getPrefKey = (application) => {
     : application.preference_record_type
 }
 
-const getAccessibilityKeys = (application) => {
+export const getAccessibilityKeys = (application) => {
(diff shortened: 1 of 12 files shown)


app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
```diff
@@ -29,15 +29,27 @@ const getPrefKey = (application) => {
     : application.preference_record_type
 }
 
+export const getAccessibilityKeys = (application) => {
+  const accessibilityKeys = compact(Object.keys(application.has_ada_priorities_selected))
```
jimlin-sfgov, line 33: Question: previously there was `... || []` to handle an empty value, is that no longer necessary?

tallulahkay: No longer necessary because I'm checking for existence of has_ada_priorities_selected before it enters this function now

jimlin-sfgov (approved):
LGTM!

PR #664, opened 2024-09-13
fix: DAH-2698 only call extra information from preferences when necessary

tallulahkay:
DAH-2698


app/javascript/components/lease_ups/LeaseUpApplicationsTableContainer.js
```diff
@@ -79,14 +80,25 @@ const LeaseUpTableContainer = ({
...
 
   useEffect(() => {
     // don't need layered validation for fcfs
-    if (listingType !== LISTING_TYPE_FIRST_COME_FIRST_SERVED) {
+    // don't need layered validation for non-veteran listings
+    // don't need layered validation for initial call
+    if (!hasFilters || (preferences && preferences.every((pref) => !pref.includes('Veteran')))) {
+      const prefMap = {}
+      addLayeredValidation(applications).forEach((preference) => {
+        console.log(preference)
```
jimlin-sfgov: issue: remove debugging line.

jimlin-sfgov (approved):
needs removal of debugging line, otherwise LG

tested on prod data, and it looks like initial load time is halved for the "Trinity Phase IV" listing

PR #667, opened 2024-09-19
fix: DAH-2718 Update Leaseup Application ADA Priorities Correctly

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2718

### DIFF
--- a/app/javascript/utils/utils.js
+++ b/app/javascript/utils/utils.js
@@ -41,8 +41,12 @@ export const filterChanged = (prev, current) => {
   const changedFields = {}
   forEach(current, (value, key) => {
     if (!isEqual(prev[key], value)) {
-      // date fields need to be treated as whole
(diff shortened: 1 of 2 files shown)


app/javascript/utils/utils.js
```diff
@@ -37,21 +37,15 @@ export const isChanged = (prev, current) => {
...
 export const filterChanged = (prev, current) => {
-  if (!prev) {
-    return current
-  }
+  if (!prev) return current
   const changedFields = {}
   forEach(current, (value, key) => {
     if (!isEqual(prev[key], value)) {
       // date fields need to be treated as whole
       if (isPlainObject(value) && prev[key] !== null && !isDateObject(value)) {
```
tallulahkay: non-blocking: another option with the same functionality that would align with previous logic: add `key !== has_ada_priorities_selected` to this if statement

tallulahkay (approved):

PR #674, opened 2024-09-26
DAH-2383 fix e2e tests

tallulahkay:
DAH-2383


jimlin-sfgov (approved):
LGTM

@tallulahkay Do you have an idea of why we needed to update our tests? e.g. did something in the Salesforce API change, or was it React code?

tallulahkay:
> LGTM
> 
> @tallulahkay Do you have an idea of why we needed to update our tests? e.g. did something in the Salesforce API change, or was it React code?

For half of it, it was a change to salesforce data (The HUD stuff) and the other half was the changes to how listings are loaded in the lease up applications page

PR #677, opened 2024-10-02
DAH-2782 switch to whitelist for url overrides in production

tallulahkay:
DAH-2782

### DIFF
--- a/app/javascript/utils/hooks/useFeatureFlag.js
+++ b/app/javascript/utils/hooks/useFeatureFlag.js
@@ -2,7 +2,7 @@ import { useFlag as useFlagUnleash, useFlagsStatus } from '@unleash/proxy-client
 
 // If you want to prevent a user from changing this feature flag via the URL, you can add it to the urlBlockList set.
 // When set, the feature flag will only be determined by the Unleash API.
-const urlBlockList = new Set([''])
+const urlWhiteList = new Set([''])
 
 export const useFeatureFlag = (flagName, defaultValue) => {
   const urlParams = new URLSearchParams(window.location.search)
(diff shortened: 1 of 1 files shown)


app/javascript/utils/hooks/useFeatureFlag.js
```diff
@@ -2,7 +2,7 @@ import { useFlag as useFlagUnleash, useFlagsStatus } from '@unleash/proxy-client
 
 // If you want to prevent a user from changing this feature flag via the URL, you can add it to the urlBlockList set.
 // When set, the feature flag will only be determined by the Unleash API.
-const urlBlockList = new Set([''])
+const urlWhiteList = new Set(['test-partners', 'testFlag'])
```
jimlin-sfgov, line 5: question: probably not in scope, but have we discussed why it would be preferred to put the white list in source code, as opposed to env vars for example?

app/javascript/utils/hooks/useFeatureFlag.js
```diff
@@ -15,7 +15,8 @@ export const useFeatureFlag = (flagName, defaultValue) => {
 
   const unleashFlag = useFlagUnleash(flagName)
 
-  if (doesURLHaveFlag && !urlBlockList.has(flagName) && process.env.UNLEASH_ENV === 'development') {
+  const useUrlOverride = urlWhiteList.has(flagName) || process.env.UNLEASH_ENV === 'development'
```
jimlin-sfgov: nit (non-blocking): rename to `allowUrlOverride`

jimlin-sfgov (commented):
I got this error when trying to log in in the review app. `error=redirect_uri_mismatch&error_description=redirect_uri%20must%20match%20configuration`

Looks like this [has happened before](https://sfdigitalservices.slack.com/archives/C030EGQAKK2/p1716215937006579), but I don't know how to fix it.

app/javascript/utils/hooks/useFeatureFlag.js
tallulahkay: We haven't - I think that's a good topic for whatever discussion we have about managing feature flags

tallulahkay:
> I got this error when trying to log in in the review app. `error=redirect_uri_mismatch&error_description=redirect_uri%20must%20match%20configuration`
> 
> Looks like this [has happened before](https://sfdigitalservices.slack.com/archives/C030EGQAKK2/p1716215937006579), but I don't know how to fix it.

This is my bad - to fix it, you add the callback url (<Application URL>/users/auth/salesforcesandbox/callback) to the DAHLIA Partners connected app in salesforce full. It has to be done for every new PR review app, and I forgot to do it! I will think about where to put this information to live.
 
Screenshot 2024-10-07 at 2 32 19 PM

jimlin-sfgov (approved):
LGTM

PR #688, opened 2024-10-31
fix: DAH-2905 Treat blank post_lottery_validation values as 'Unconfirmed'

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-2905


tallulahkay (approved):
This looks good!

PR #690, opened 2024-11-12
fix: DAH-2931 restore pagination

tallulahkay:
DAH-2931

Restores pagination for partners app.


tallulahkay:
@alulabeshue-sfgov @chadbrokaw I am still working on code coverage, but you can start reviewing!

tallulahkay:
@alulabeshue-sfgov @chadbrokaw ok, test cov is there! ready for review!

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -94,6 +98,13 @@ const LeaseUpApplicationsPage = () => {
     eagerPagination: new EagerPagination(ROWS_PER_PAGE, SERVER_PAGE_SIZE)
   })
 
+  useEffect(() => {
+    if (flagsReady && !partnersPaginationEnabled) {
+      state.eagerPagination = new EagerPagination(ROWS_PER_PAGE, 50000)
```
jimlin-sfgov: issue: use `setState` instead of assignment when changing React states.

app/javascript/components/lease_ups/utils/leaseUpRequestUtils.js
```diff
@@ -64,6 +106,9 @@ export const getApplications = async (
 
 export const getListing = async (listingId) => apiService.getLeaseUpListing(listingId)
 
+/**
+ * @deprecated No longer used anywhere.
+ */
```
jimlin-sfgov, line 116: question: are there followup plans for the deprecation comments in this PR? Also, why the `@` sign?

app/services/force/soql/preference_pagination_service.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+
+      def listing_subquery(listing_id)
+        builder.from(:Listing_Lottery_Preference__c)
+               .select(:Id)
+               .where_eq('Listing__c', "'#{listing_id}'")
+      end
+    end
+  end
+end
```
jimlin-sfgov, line 122: question: Since you've created a new service file, are there plans to eventually remove the old `preference_service.rb` file?

jimlin-sfgov (changes requested):

app/javascript/components/lease_ups/utils/leaseUpRequestUtils.js
tallulahkay: The @ sign because it is a magic thing that makes the function crossed out wherever it is used - I wonder if that only works on my linter

Followup plans is a good question. I imagine it should be the last step after the feature flag has been on for a while. @alulabeshue-sfgov maybe we should add a ticket to remove the deprecated code to the current epic?

app/services/force/soql/preference_pagination_service.rb
tallulahkay: We haven't yet talked about these plans, but I think we should. If we weren't doing a feature flag I would have just completely changed this file, but doing it this way allows us to use a feature flag (and make it easier to review, I think? you be the judge of that)

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
tallulahkay: Thank you!

jimlin-sfgov: Following up on a pairing session we had:
We should use `setState({ eagerPagination: new EagerPagination(ROWS_PER_PAGE, 50000) })` here.
However, the `useAsync` function further down needs to add `state.eagerPagination` to it's dependency array so it will fetch things properly whenever the pagination configuration changes due to the flag.

jimlin-sfgov (approved):
LGTM

not sure why e2e tests are failing though

tallulahkay:
> LGTM
> 
> not sure why e2e tests are failing though

@jimlin-sfgov The e2e tests are failing bc of a change in salesforce. I made a ticket to fix: https://sfgovdt.jira.com/browse/DAH-2974

PR #697, opened 2024-12-27
feat: DAH-3042 conditionally hide HCBS units checkbox

tallulahkay:
DAH-3042

Only show HCBS Units checkbox in the supplemental application page if the name of the listing includes "Kelsey".


app/javascript/components/supplemental_application/SupplementalApplicationContainer.js
```diff
@@ -108,7 +108,10 @@ const ConfirmedPreferencesSection = ({
       </FormGrid.Row>
     </ContentSection.Sub>
     <ContentSection.Sub title='Household Reserved and Priority Units'>
-      <ConfirmedUnits form={form} />
+      <ConfirmedUnits
+        form={form}
+        showHCBSUnitsCheckbox={application.listing.name.includes('Kelsey')}
```
jimlin-sfgov: question: why do a string match on the listing name instead adding `Listing__r.Custom_Listing_Type__c` around [here](https://github.com/SFDigitalServices/sf-dahlia-lap/blob/8717802afed87fe6e254e0f25743390fe1fa58ad/config/salesforce/fields/applications.yml#L95) and checking that?

tallulahkay: Just because I forgot that we ended up doing a Custom Listing Type for the Kelsey so I'll change it to that

jimlin-sfgov (approved):
LGTM

PR #699, opened 2025-01-03
fix: hardcoding year to prevent updating the test every year

alulabeshue-sfgov:
## Description

Hardcode the year for supplemental application test to avoid having to change the test every year

## Jira ticket

n/a

## Review instructions

`frontend` tests that are failing on the [main](https://github.com/SFDigitalServices/sf-dahlia-lap/commits/main/) branch should pass on this PR

### DIFF
--- a/spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
+++ b/spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
@@ -187,11 +187,17 @@ describe('SupplementalApplicationPage', () => {
   })
 
   test('it should render as expected', async () => {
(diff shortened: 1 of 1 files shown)


chadbrokaw (approved):
This look good, not a blocker for merging but I wonder if at some point we should make this a global change so that all tests run assuming the same time?

alulabeshue-sfgov:
> This look good, not a blocker for merging but I wonder if at some point we should make this a global change so that all tests run assuming the same time?

Makes sense. I tried it a bit but it broke a lot of tests that are expecting a specific date so I thought it was a bigger effort than the purpose of this ticket which is to fix the broken main branch.

PR #711, opened 2025-03-18
feat: DAH-3321 Restrict access to lottery pdf generator

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-3321


josh-chou (approved):
looks good!  👍

PR #719, opened 2025-04-24
feat: DAH-3359 GraphQL Integration for Lease Up Applications pages

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-3359


app/services/force/graphql/lease_up_application_preferences.rb
```diff
@@ -0,0 +1,221 @@
...
+    # Query for Application Preferences that are in lease up, with filters
+    class LeaseUpApplicationPreferences < Force::GraphqlQuery
+      def initialize(params)
+        super(params)
+        @salesforce_object_name = 'Application_Preference__c'
+      end
+
+      def query
+        call(build_lease_up_applications_query(@params))
+        @params[:page].to_i.times do
```
jimlin-sfgov, line 15: note: due to limitations with Salesforce's graphql implementation, we cannot skip pages. E.g. to get the results for page 3, we need to query for pages 1 and 2.

https://developer.salesforce.com/docs/platform/graphql/guide/paginate-use-upperbound.html

app/services/force/graphql_query.rb
```diff
@@ -0,0 +1,125 @@
...
+      @params = params
+      @record_batch_size = params[:record_batch_size].try(:to_i) || 2_000 # supports 200 to 2_000
+      @records = []
+      @response = nil
+      @paging_cursor = nil
+      @total_count = nil
+    end
+
+    def call(query_string = nil)
+      query_str = query_string || @query_string
```
jimlin-sfgov, line 22: Note: Provides an option to make a GraphQL call with a custom query string, to make it easier to debug and develop future GraphQL-based endpoints.

```ruby
user = User.last
query = Force::GraphqlQuery.new(user, {})
query.call(<<~GQL
  ...
GQL)
query.instance_variable_get(:@response)
```

app/services/force/flagged_record_set_service.rb
```diff
@@ -65,7 +65,7 @@ def flagged_record_set(application_id)
     def update_flagged_application(data)
       data = Hashie::Mash.new(data)
       return nil unless data[:Id]
-      puts "updating flagged application with: #{data.as_json}"
+      # puts "updating flagged application with: #{data.as_json}"
```
alulabeshue-sfgov, line 68: **question:** why comment out this line?

config/boot.rb
```diff
@@ -3,3 +3,4 @@
 ENV['BUNDLE_GEMFILE'] ||= File.expand_path('../Gemfile', __dir__)
 
 require 'bundler/setup' # Set up gems listed in the Gemfile.
+require 'logger' # Fix concurrent-ruby removing logger dependency which Rails itself does not have
```
alulabeshue-sfgov: **question:** why is this now required?

app/services/force/responses.rb
```diff
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+require 'facets/hash/rekey'
```
alulabeshue-sfgov, line 3: **question:** why is this now required?

app/services/force/graphql_query.rb
```diff
@@ -0,0 +1,115 @@
...
+  # GraphQL Requests to Salesforce
+  class GraphqlQuery
+    GRAPHQL_ENDPOINT = '/services/data/v63.0/graphql'
+
+    attr_reader :total_count
+
+    def initialize(user, params)
+      @client = ClientFactory.new(user).build
+      @params = params
+      @record_batch_size = params[:record_batch_size].try(:to_i) || 2_000 # supports 200 to 2_000
```
alulabeshue-sfgov: **suggestion** move magic number `2_000` to a constant

app/javascript/components/lease_ups/LeaseUpApplicationsPage.js
```diff
@@ -32,6 +32,9 @@ import TableLayout from '../layouts/TableLayout'
 import { createFieldUpdateComment } from '../supplemental_application/utils/supplementalRequestUtils'
 
 const ROWS_PER_PAGE = 20
+// for lease up applications endpoint, Rails has different paging limits
+// see app/services/force/graphql_query.rb for details
+const GRAPHQL_SERVER_PAGE_SIZE = 2000
```
alulabeshue-sfgov: **suggestion:** move this const to EagerPagination.js so that it can be used by other pages in the future when migrating to graphql

alulabeshue-sfgov (commented):
Testing looks good, nice job with all the code changes and code clean up! A few questions and comments.

app/services/force/flagged_record_set_service.rb
jimlin-sfgov: It was cluttering up my CLI when I ran tests. I figured it was used for debugging so I commented it out. If you think we should have observability on this, I can convert it to a log.

config/boot.rb
jimlin-sfgov: I ran into issues when running the app locally, details here: https://stackoverflow.com/a/79379493

It seems like the Rails upgrade fixed the issue, so I can remove it now.

app/services/force/responses.rb
jimlin-sfgov: It's due to how `require` and Ruby inheritance works. Without this, the `massage` method (which uses `rekey`) in `graphql_query.rb
` won't work: https://github.com/SFDigitalServices/sf-dahlia-lap/pull/719/files#diff-2a0d7f9626778288810acce61c8b624106a72233071a93d49aee6c80da2f41ccR63

You can try it yourself locally by removing this line and then viewing a lease-up listing.

alulabeshue-sfgov (approved):
LGTM!

PR #724, opened 2025-05-01
chore: DAH-2969 clean up feature flag and code for pagination in partners

josh-chou:
https://sfgovdt.jira.com/browse/DAH-2969


alulabeshue-sfgov (commented):
1. In the code there are comments of `todo remove in DAH-2969` which are deprecated functions that should be cleaned up as part of this ticket
-- including cleaning up unused code that use these deprecated functions like `LeaseUpApplicationsTableContainer.buildApplicationsWithLayeredValidations`
2. `LeaseUpApplicationsTable` has the param `prefMap` that is no longer used and can be cleaned up
-- including the use of `const [prefMap, setPrefMap] = useState({})` in `LeaseUpApplicationsTableContainer`

alulabeshue-sfgov (approved):
LGTM! Nice job with all the code clean up!

PR #728, opened 2025-05-19
feat: DAH-3411 Decrease GraphQL page size to 200

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-3411


josh-chou (approved):
loaded for me in under 3 secs.

PR #731, opened 2025-05-30
feat: DAH-3433 add lease only for available unit status

alulabeshue-sfgov:
## Ticket

[DAH-3433](https://sfgovdt.jira.com/browse/DAH-3433)

## Description

After discussing with MOHCD we landed on only units that have a status of `Available` should available in partners to create a lease for. If the unit has an undefined status or a status of `Occupied`, users should not be able to create a lease for it.

## Review Instructions

Follow review instructions on jira ticket

[DAH-3433]: https://sfgovdt.jira.com/browse/DAH-3433?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


jimlin-sfgov (approved):
LGTM

PR #733, opened 2025-06-02
DAH-3439 Use lottery results api

jimlin-sfgov:
https://sfgovdt.jira.com/browse/DAH-3439


app/javascript/components/lease_ups/lottery_results_page/LotteryResultsPdfGenerator.js
```diff
@@ -28,9 +28,16 @@ const LotteryResultsPdfGenerator = (props) => {
...
 
   useEffect(() => {
-    apiService.fetchApplicationsForLotteryResults(props.listing_id).then((res) => {
-      setApplications(res.records)
-    })
+    const urlParams = new URLSearchParams(window.location.search)
+    if (urlParams.has('useLotteryResultApi')) {
+      apiService.fetchLotteryResults(props.listing_id).then((res) => {
+        setApplications(res.lotteryBuckets)
+      })
```
alulabeshue-sfgov, line 37: **issue:** this [request](https://dahlia-partners-pr-733.herokuapp.com/api/v1/lottery-results?listing_id=a0W0P00000HbwpUUAR&use_lottery_result_api=true) is returning a field `URL` which has a link to a teams meeting invite

app/javascript/components/lease_ups/lottery_results_page/utils/processLotteryBuckets.js
```diff
@@ -153,3 +153,40 @@ export const processLotteryBuckets = (applicationPreferences) => {
...
+
+  const unfilteredResults = []
+  buckets.forEach((bucket) =>
+    bucket.preferenceResults.forEach((result) =>
+      unfilteredResults.push({
+        lottery_number: result.lotteryNumber,
+        unsorted_lottery_rank: result.lotteryRank
+      })
+    )
+  )
```
alulabeshue-sfgov: **suggestion:** we can reduce the number of nested loops by combining the building of `massagedBuckets` and `unfilteredResults` into a single nested loop

app/javascript/components/lease_ups/lottery_results_page/LotteryManager.js
```diff
@@ -16,7 +16,12 @@ const LotteryManager = ({ applicationPrefs, listing }) => {
   })
 
   // process applications into buckets
-  const processedBuckets = applicationPrefs ? processLotteryBuckets(applicationPrefs) : null
+  let processedBuckets = null
+  if (applications && applications[0].totalSubmittedApps) {
```
alulabeshue-sfgov: **suggestion** pull this logic into a named function (i.e. `wasApiUsed()`) or add a comment. Makes it more understandable for the next eng rather than having to understand `totalSubmittedApps` is returned by the API

alulabeshue-sfgov (commented):
Left some comments, also this PR fails for [listings](https://dahlia-partners-pr-733.herokuapp.com/listings/a0W7y000003wjxVEAQ/lottery-results) that have not ran a lottery

app/javascript/components/lease_ups/lottery_results_page/LotteryResultsPdfGenerator.js
jimlin-sfgov: The Salesforce api  `LotteryResult` endpoint that's used here is the same one used to show lottery results on the webapp. There will be some fields that are used in the webapp but not in partners, like the `URL` field.

This PR's goal is to check the viability of using a `LotteryResult` endpoint for a feature that is not available to the public or even Leasing Agents, so I think it's okay, but please let me know otherwise.

The fact that the `URL` field has a teams link is a data issue(?): https://sfhousing--full.sandbox.lightning.force.com/lightning/r/Listing__c/a0W0P00000HbwpUUAR/view

app/javascript/components/lease_ups/lottery_results_page/utils/processLotteryBuckets.js
jimlin-sfgov: I attempted to combine, but my resulting code looks less legible and has the same LOC. I think it's not worth the performance boost.

Edit: is it less legible? I'm on the fence now

```js
// ...
  const massagedBuckets = []
  const unfilteredResults = []
  buckets.forEach((bucket) => {
    const massagedBucket = {
      shortCode: bucket.preferenceShortCode || 'generalLottery',
      preferenceResults: []
    }
    bucket.preferenceResults.forEach((result) => {
      unfilteredResults.push({
        lottery_number: result.lotteryNumber,
        unsorted_lottery_rank: result.lotteryRank
      })
      massagedBucket.preferenceResults.push({ lottery_number: result.lotteryNumber })
    })
    massagedBuckets.push(massagedBucket)
  })
// ...
```

app/javascript/components/lease_ups/lottery_results_page/LotteryResultsPdfGenerator.js
alulabeshue-sfgov: Got it, I didnt realize it was a user input from a salesforce object.

app/javascript/components/lease_ups/lottery_results_page/utils/processLotteryBuckets.js
alulabeshue-sfgov: I agree with the original code is simpler to understand. Im ok with leaving the code as is. Thanks for taking a look!

alulabeshue-sfgov (approved):
LG!

PR #774, opened 2025-12-11
fix: DAH-3829 MALFORMED_QUERY error when a listing has no applications

josh-chou:
## Description

Fix an error when a listing has no applications

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-3829

### Version Control

### Code quality

### Review instructions

- [ ] instructions work for PA testers

### Request eng review

- [x] PA tested in the review environment (use `needs product acceptance` label)

### DIFF
--- a/app/services/force/field_update_comment_service.rb
+++ b/app/services/force/field_update_comment_service.rb
@@ -33,6 +33,8 @@ module Force
     end
 
     def bulk_status_history_by_applications(application_ids)
+      return [] if application_ids.empty?
+
       quoted_ids = application_ids.map { |item| "'#{item}'" }
       result = parsed_index_query(
         %(


fwextensions:
One line fix!  Ship it!

app/services/force/field_update_comment_service.rb
```diff
@@ -33,6 +33,8 @@ def status_history_by_application(application_id)
     end
 
     def bulk_status_history_by_applications(application_ids)
+      return [] if application_ids.empty?
```
jimlin-sfgov, line 36: I don't think this actually makes the error message go away, `[].join(',')` still returns an empty string, which means we'll still have `WHERE Application__c in ()` in the query.

josh-chou: The function will automatically return [] without running the query.  So it never gets to line 43 where the join is done.  Unless you are referring to another join in following query some place else?

jimlin-sfgov: I've misread the code 🤦  thanks for explaining

jimlin-sfgov (approved):
LGTM

PR #791, opened 2026-01-29
chore: DAH-0000 resolve flaky SupplementalApplicationPage tests

fwextensions:
## Description

Add proper async handling to tests that were failing intermittently due to race conditions. Tests were querying for DOM elements before async data loading completed. This was causing `SupplementalApplicationPage.test.js` to sometimes fail in slower environments on Windows, like the VS Code terminal, but then it would work in the Windows terminal. Ensuring that async calls are always awaited and bumping the timeouts seems to have fixed it.

Changes:
- Add waitFor to getWrapper() to ensure form is loaded before tests run
- Use findByRole instead of getByRole for elements that appear after async ops
- Add await to selectEvent.openMenu() calls which are async
- Wrap remaining state-changing interactions in await act()

### Version Control

- [ ] branch name contains the Jira ticket number

### Code quality

### Review instructions

- [ ] instructions specify which environment(s) it applies to
- [ ] instructions work for PA testers

### Request eng review

- [ ] PR has `needs review` label

- [ ] PA tested in the review environment (use `needs product acceptance` label)

### DIFF
--- a/spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
+++ b/spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
@@ -1,9 +1,13 @@
-import { render, screen, fireEvent, within, act } from '@testing-library/react'
+import { render, screen, fireEvent, within, act, waitFor } from '@testing-library/react'
 import { useFlag as useFlagUnleash, useFlagsStatus, useVariant } from '@unleash/proxy-client-react'
 import { cloneDeep } from 'lodash'
 import selectEvent from 'react-select-event'
 
 import supplementalApplication from '../../fixtures/supplemental_application'
+
+// Increase default test timeout for this file - tests involve multiple async operations
+// and can be slow in resource-constrained environments (e.g., VS Code terminal)
+jest.setTimeout(15000)
 import { leaseUpAppWithUrl, renderAppWithUrl } from '../../testUtils/wrapperUtil'
 
 const mockInitialLoad = jest.fn()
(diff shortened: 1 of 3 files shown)


spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
```diff
@@ -1,11 +1,15 @@
...
 import { useFlag as useFlagUnleash, useFlagsStatus, useVariant } from '@unleash/proxy-client-react'
 import { cloneDeep } from 'lodash'
 import selectEvent from 'react-select-event'
 
 import supplementalApplication from '../../fixtures/supplemental_application'
 import { leaseUpAppWithUrl, renderAppWithUrl } from '../../testUtils/wrapperUtil'
 
+// Increase default test timeout for this file - tests involve multiple async operations
+// and can be slow in resource-constrained environments (e.g., VS Code terminal)
+jest.setTimeout(15000)
```
jimlin-sfgov, line 11: Thought: this is a smell for our long-standing issue of slow loading times, which we may work on one day...

jimlin-sfgov (approved):
LGTM

On the other hand, we can choose not to fix this, and use the flakiness as a reason to get you a beefed up laptop @fwextensions

spec/javascript/components/supplemental_application/SupplementalApplicationPage.test.js
fwextensions: That's what Opus thinks.

>   Slow patterns in current tests:
> 
>   1. Expensive cloning - Every API mock does _cloneDeep(mockedApplication) on a 406-line object
>   2. Redundant async - Mocks are async but return immediately (adds microtask overhead)
>   3. Full page render per test - Nested beforeEach calls getWrapper() repeatedly:
>     - "when no unit is selected": 2 tests = 2 renders
>     - "when unit without priority is selected": 4 tests = 4 renders
>     - "when unit with priority is selected": 2 tests = 2 renders
> 
>   That's 8 full renders just in Lease Section nested describes

PR #799, opened 2026-02-13
fix: DAH-3944 Check for I2A flag 

cliu02:
## Description

Add check for main I2A flag.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-3944

### Version Control

### Code quality

### Review instructions

- [x] instructions work for PA testers

### Request eng review

- [ ] PA tested in the review environment (use `needs product acceptance` label)

### DIFF
--- a/app/javascript/utils/inviteApplyEmail.js
+++ b/app/javascript/utils/inviteApplyEmail.js
@@ -41,11 +41,14 @@ export const IsStatusesEnabled = () => {
 export const getLeaseUpSubstatusOptions = (listingId) => {
   const isInviteApplyEnabled = IsInviteToApplyEnabledForListing(listingId)
   const isStatusesEnabled = IsStatusesEnabled()
+  // I2A enabled and new statuses enabled
   if (isInviteApplyEnabled && isStatusesEnabled) {
     return LEASE_UP_SUBSTATUSES
-  } else if (!isStatusesEnabled) {
+    // I2A enabled and new statuses not enabled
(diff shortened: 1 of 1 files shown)


app/javascript/utils/inviteApplyEmail.js
```diff
@@ -41,11 +41,14 @@ export const IsStatusesEnabled = () => {
...
   const isStatusesEnabled = IsStatusesEnabled()
+  // I2A enabled and new statuses enabled
   if (isInviteApplyEnabled && isStatusesEnabled) {
     return LEASE_UP_SUBSTATUSES
-  } else if (!isStatusesEnabled) {
+    // I2A enabled and new statuses not enabled
+  } else if (isInviteApplyEnabled && !isStatusesEnabled) {
     return LEASE_UP_SUBSTATUS_OPTIONS
   } else {
+    // I2A not enabled and new statuses not enabled
```
jimlin-sfgov, line 51: question: This is the final `else` condition, so it also handles the scenario. where "I2A **not** enabled and new statuses **enabled**". Just want to confirm that this is intended?

cliu02: Good point, I was assuming that we would never have this case so there are only three options in reality. I'm not sure what we would want to show.

jimlin-sfgov (approved):
LGTM

As a reminder, Partners review apps need to be configured in Salesforce "Manage Connected Apps" settings to properly log in.

PR #806, opened 2026-02-25
fix: DAH-3962 i2a ga save autofilled data to salesforce

josh-chou:
## Description

Fix bug where data is not saved to salesforce because modals were skipped.
The fix is to move saving of all data to the final review modal instead of after each modal.

## Jira ticket

https://sfgovdt.jira.com/browse/DAH-3962

### Version Control

### Code quality

- [ ] [the set of changes is small](https://google.github.io/eng-practices/review/developer/small-cls.html#what-is-small)
- [x] if the PR is a bugfix, there are tests and logs around the bug

### Review instructions

- [x] instructions work for PA testers
- [ ] instructions have already been performed at least once

### Request eng review

- [ ] Use `Housing Eng` group to automatically assign reviewers, and/or assign specific engineers

- [x] PA tested in the review environment (use `needs product acceptance` label)


app/javascript/components/lease_ups/InviteToApplyModals.js
```diff
@@ -273,40 +245,80 @@ export const InviteToApplyModals = forwardRef((props, ref) => {
...
+                      email: exampleEmail
+                    })
+                  } else {
+                    props.setPageState({
+                      loading: false,
+                      showPageInfo: true
+                    })
+                  }
+                })
+            }
```
alulabeshue-sfgov, line 311: **nit:** break this logic down into function(s)

alulabeshue-sfgov (approved):

app/javascript/components/lease_ups/InviteToApplyModals.js
josh-chou: @alulabeshue-sfgov What do you have in mind? It would break down into single line functions that don't get called anywhere else.  Perhaps more commenting for clarity would be better?

alulabeshue-sfgov: @josh-chou I was thinking maybe just pull it into a single function, that the name of the function can five insights on what is happening. But comments sounds good as well.

